PR #1, opened 2020-08-05
UCC headers reflecting the interfaces and semantics as discussed in WG

manjugv:
The header files provide useful documentation. Complete documentation is at:
* [UCC Spec](https://github.com/manjugv/ucc_wg_public/blob/master/docs/UCC_V1.0_API_Draft_gorentla.pdf)
* [WG discussion notes](https://github.com/openucx/ucc/wiki/UCF-Collectives-Working-Group)


src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ *                              UCC_.
+ *   @param [in]  filename      If not NULL, read configuration values from the
+ *                              file defined by @e filename. If the file does not exist, it will be ignored
+ *                              and no error will be reported to the user.
+ *
+ * @return Error code as defined by @ref ucc_status_t
+ */
+
+ucc_status_t ucc_lib_config_read(const char *env_prefix, const char *filename,
+                                 ucc_lib_config_t **config_p);
```
vspetrov, line 118: ucc_lib_config_t **config_p should be imho ucc_lib_config_h *configh_p. Since ucc_lib_config internals are never exposed to the user we should declare it as "handle", ie typedef struct ucc_lib_config* ucc_lib_config_h . ?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ * @brief Release configuration descriptor
+ *
+ * The routine releases the configuration descriptor that was allocated through
+ * @ref ucc_lib_config_read "ucc_lib_config_read()" routine.
+ *
+ * @param [in] config        Pointer to the configuration descriptor to be released. 
+ *                           Configuration descriptor as defined by
+ */
+
+void ucc_lib_config_release(ucc_lib_config_t *config);
```
vspetrov, line 131: ucc_lib_config_h config

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ * @ref ucc_lib_config_h "configuration" descriptor.
+ *
+ * @param [in]  config        @ref ucc_lib_config_h "Configuration descriptor"
+ *                            to print.
+ * @param [in]  stream        Output stream to print the configuration to.
+ * @param [in]  title         Configuration title to print.
+ * @param [in]  print_flags   Flags that control various printing options.
+ */
+
+void ucc_lib_config_print(const ucc_lib_config_t *config, FILE *stream,
```
vspetrov, line 147: ucc_confif_print_flags_t - is -it defined?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+
+typedef enum {
+   UCC_CONTEXT_EXCLUSIVE = 0 ,
+   UCC_CONTEXT_SHARED
+} ucc_context_type_t;
+
+enum ucc_context_attribs_field {
+    UCC_CONTEXT_ATTRIBS_TYPE                   = UCC_BIT(0),
+    UCC_CONTEXT_ATTRIBS_COLL_SYNC_TYPE         = UCC_BIT(1),
+    UCC_CONTEXT_ATTRIBS_COLL_OOB               = UCC_BIT(2)
```
vspetrov, line 234: COLL_OOB is probably should not be part of attribs? It is always passed by user.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_CONTEXT_PARAMS_TYPE                   = UCC_BIT(0),
+    UCC_CONTEXT_PARAMS_COLL_SYNC_TYPE         = UCC_BIT(1),
+    UCC_CONTEXT_PARAMS_COLL_OOB               = UCC_BIT(2)
+};
+
+typedef struct ucc_context_params {
+    uint64_t                mask;
+    ucc_context_type_t      ctx_type;
+    ucc_coll_sync_type_t    sync_type;
+    ucc_context_oob_coll_t  oob_func;
```
vspetrov, line 247: i don't like oob_func name - it is not a func. maybe just "oob"?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+
+
+typedef struct ucc_context_attribs {
+    uint64_t                mask;
+    ucc_context_type_t      ctx_type;
+    ucc_coll_sync_type_t    sync_type;
+} ucc_context_attribs_t;
+
+
+ucc_status_t ucc_context_config_read(ucc_lib_h lib, const char *env_prefix,
```
vspetrov, line 258: i dont think we need "env_prefix" there. It is only for "lib" object. Since context is created from lib it inherits the "env_prefix"

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    ucc_mem_hints_t         hints;
+    ucc_mem_constraints_t   constraints;
+} ucc_mem_map_params_t;
+
+typedef struct  ucc_team_oob_coll {
+    int             (*allgather)(void *src_buf, void *recv_buf, size_t size,
+                                 void *allgather_info,  void **request);
+    ucc_status_t    (*req_test)(void *request);
+    ucc_status_t    (*req_free)(void *request);
+    uint32_t 	     participants;
```
vspetrov, line 385: alignment

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ *  @brief @ref ucc_team_create_test routines tests the status of team handle.
+ *  If required it can progress the communication but cannot block on the
+ *  communications.
+ *
+ *  @param  [in] ucc_team  Team handle to test
+ *
+ * @return Error code as defined by @ref ucc_status_t
+ */
+
+ucc_status_t ucc_team_create_test();
```
vspetrov, line 460: ucc_team_h as argument

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_DT_FLOAT32,
+    UCC_DT_FLOAT64,
+    UCC_DT_USERDEFINED,
+    UCC_DT_OPAQUE
+} ucc_datatype_t;
+
+typedef struct ucc_coll_buffer_info {
+    uint64_t        mask;
+    void            *src_buffer;
+    ucc_count_t     *src_counts;
```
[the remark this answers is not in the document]
manjugv: To support both MPI_Collective interfaces and MPI_Collective_l interfaces, which will be included in the upcoming version of MPI. In the WG, we discussed multiple options and settled on this approach. See the discussion [here](https://github.com/manjugv/ucc_wg_public/blob/master/slides/gorentla_ucc_schedule_tasks_datatypes_July8th.pdf).

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+typedef struct ucc_coll_ext_op_args {
+    ucc_coll_buffer_info_t      buffer_info;
+    ucc_reduction_info_t        reduction_info;
+    ucc_error_type_t            error_type;
+    ucc_coll_id_t               tag;
+    uint64_t                    root;
+} ucc_coll_ext_op_args_t;
+
+typedef struct ucc_coll_op_args {
+    uint64_t                    mask;
```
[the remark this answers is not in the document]
manjugv: The idea was to optimize for collectives that do not require any buffer arguments (barrier). Only keep minimal in op_args and all extended in ext_args. It was an ask from WG for optimization.

src/api/ucc.h
manjugv: Agree.

src/api/ucc.h
manjugv: I’m fine with that.

src/api/ucc.h
manjugv: I debated about it. Given that our semantics require that only one outstanding team create operation at any given time, do we want it? I don’t have a strong opinion on it.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    uint64_t        mask;
+    void            *src_buffer;
+    ucc_count_t     *src_counts;
+    ucc_aint_t      *src_displacements;
+    void            *dst_buffer;
+    ucc_count_t     *dst_counts;
+    ucc_aint_t      *dst_displacements;
+    ucc_datatype_t  src_datatype;
+    ucc_datatype_t  dst_datatype;
+    uint64_t        flags;
```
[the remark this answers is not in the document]
manjugv: We discussed multiple uses for the flags, but did not flesh out. We want to express 1) MPI_IN_PLACE 2) To differentiate between 32 bit and 64 counts/offsets and 3) Whether the buffers are persistent or not ? Good catch.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ * @ref ucc_team_get_my_ep routine queries and returns the endpoint of the
+ * participant invoking the interface.
+ *
+ * @param [out]  ep          Endpoint of the participant calling the routine
+ * @param [in]   ucc_team    Team handle
+ *
+ * @return Error code as defined by @ref ucc_status_t
+ */
+
+ucc_status_t ucc_team_get_my_ep(ucc_team_h ucc_team, uint64_t *ep);
```
Sergei-Lebedev, line 544: Why dedicated function is needed to query team ep? Is it the same to query ep using ucc_team_get_attribs?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ * @ref ucc_team_get_size routine queries the size of the team. It reflects the
+ * number of unique endpoints in the team.
+ *
+ * @param [in]   ucc_team  Team handle
+ * @param [out]  size      The size of team as number of endpoints
+ *
+ * @return Error code as defined by @ref ucc_status_t
+ */
+
+ucc_status_t ucc_team_get_size(ucc_team_h ucc_team, uint32_t *size);
```
Sergei-Lebedev, line 529: Should we make team size part of team attributes?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ * the team handle. It takes in parameters ucc_context_h, num_handles,
+ * ucc_team_params_t and returns a ucc_team_handle_h. The ucc_team_params_t
+ * provides user configuration to customize the team. The routine returns
+ * immediately after posting the operation with the new team handle. However,
+ * the team handle is not ready for posting the collective operation.
+ * ucc_team_create_test operation is used to learn the status of the new team
+ * handle. On error, the team handle will not be created and corresponding error
+ * code as defined by ucc_status_t is returned.
+ *
+ * @param  [in] contexts     Communication context abstracting the resources
```
Sergei-Lebedev, line 434: params description of ucc_team_create_post is not consistent with actual function definition

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_GATHER             = UCC_BIT(6),
+    UCC_SCATTER            = UCC_BIT(7),
+    UCC_MCAST              = UCC_BIT(8),
+    UCC_FANIN              = UCC_BIT(9),
+    UCC_FANOUT             = UCC_BIT(10),
+    UCC_COLL_LAST          = UCC_BIT(11)
+} ucc_coll_type_t;
+
+typedef enum {
+    UCC_THREAD_SINGLE       = UCC_BIT(0),
```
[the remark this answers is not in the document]
manjugv: Agree.

src/api/ucc.h
manjugv: It is more of a convenient interface for frequently used functionality.

src/api/ucc.h
manjugv: Same here. It is more of a convenient interface for frequently used functionality.

src/api/ucc.h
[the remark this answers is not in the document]
vspetrov: @manjugv i don't see the optimization here - the size of coll_op_args is still the same. You still need to set "mask" values. So - to me, "ext" struct inside another struct does not make it better.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_OP_XOR              = UCC_BIT(7),
+    UCC_OP_LAND             = UCC_BIT(8),
+    UCC_OP_LOR              = UCC_BIT(9),
+    UCC_OP_LXOR             = UCC_BIT(10),
+    UCC_OP_BAND             = UCC_BIT(11),
+    UCC_OP_BOR              = UCC_BIT(12),
+    UCC_OP_BXOR             = UCC_BIT(13),
+    UCC_OP_MAXLOC           = UCC_BIT(14),
+    UCC_OP_MINLOC           = UCC_BIT(15)
+} ucc_reduction_op_t;
```
alex--m, line 48: IIRC we wanted this to be a union of enum and pointer, to support passing a user-defined operation.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+} ucc_team_params_t;
+
+typedef struct ucc_team_attribs {
+    uint64_t               mask;
+    ucc_post_ordering_t    ordering;
+    uint64_t               outstanding_colls;
+    uint64_t               ep;
+    ucc_ep_range_type_t    ep_range;
+    ucc_coll_sync_type_t   sync_type;
+    ucc_mem_map_params_t   mem_params;
```
alex--m, line 417: possibly missing: how do I specify my index ("rank") in the team, e.g., for Gather() operations (where the offset depends on my rank)?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_DT_UINT8,
+    UCC_DT_UINT16,
+    UCC_DT_UINT32,
+    UCC_DT_UINT64,
+    UCC_DT_UINT128,
+    UCC_DT_FLOAT16,
+    UCC_DT_FLOAT32,
+    UCC_DT_FLOAT64,
+    UCC_DT_USERDEFINED,
+    UCC_DT_OPAQUE
```
alex--m, line 583: if OPAQUE (or user-defined) is chosen how do we pass the pointer? I seem to recall we considered making this a union of enum+pointer, but I'm not sure what was the conclusion. I'm asking because I don't see any other relevant field in `ucc_coll_buffer_info_t`.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+typedef enum {
+    UCC_COLLECTIVE_EP_RANGE_CONTIG = 0, 
+    UCC_COLLECTIVE_EP_RANGE_NONCONTIG = 1
+} ucc_ep_range_type_t;
+
+typedef struct ucc_team_params {
+    uint64_t                mask;
+    ucc_post_ordering_t     ordering;
+    uint64_t                outstanding_colls;
+    uint64_t                ep;
```
alex--m, line 403: I think we should clarify what "ep" means. Is it an index (if yes, in which array/table?), a pointer (to what object?), etc.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    ucc_ep_range_type_t     ep_range;
+    ucc_coll_sync_type_t    sync_type;
+    ucc_team_oob_coll_t     oob_collective;
```
alex--m, line 406: Does it make sense to pass it once, for the entire UCC library, instead of here? most of it looks like it's common for all teams.

src/api/ucc.h
alex--m: Furthermore, since UCG works on top of UCT - in order to be able to port it into UCC I need some what to translate from "rank" to UCT address. "rank" is also used when the user asks for a specific "root" for the operation (e.g. reduce). 

Could you please clarify how can I translate from "Communicator ranks" to UCC "ep" and/or some sort of addresses?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    ucc_mem_map_params_t    mem_params;
```
alex--m, line 407: I'm missing topology information. More specifically, I need information about which members of my team are on the same host and which are on the network. Today UCG gets this information from MPI (as a "distance vector"), and I'm not sure how I achieve this using the given API.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_ERR_TYPE_GLOBAL=1
+} ucc_error_type_t;
+
+typedef uint16_t ucc_coll_id_t ;
+
+typedef struct ucc_coll_ext_op_args {
+    ucc_coll_buffer_info_t      buffer_info;
+    ucc_reduction_info_t        reduction_info;
+    ucc_error_type_t            error_type;
+    ucc_coll_id_t               tag;
```
alex--m, line 616: MPI collectives API doesn't have tags - do you mean this will be an optional feature, which will behave like P2P tags? If so, is it "always on"? (if always on, sounds like we should have a way to turn it off, to save cycles)

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+} ucc_coll_buffer_info_t;
+
+typedef struct ucc_reduction_info {
+    ucc_datatype_t      dt;
+    ucc_reduction_op_t  op;
+    size_t              count;
+} ucc_reduction_info_t;
+
+typedef enum {
+    UCC_ERR_TYPE_LOCAL=0,
```
alex--m, line 606: spaces around '='

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_ERR_TYPE_GLOBAL=1
+} ucc_error_type_t;
```
alex--m, line 608: I'm assuming this field is not future-extendable anyway (no mask) - why not replace it with a boolean?

src/api/ucc_status.h
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @brief Status codes for the UCC operations
+ */
+
+typedef enum {
+    /* Operation completed successfully */
+    UCC_OK                              =   0,
+
+    /* Operation is posted and is in progress */
+    UCC_INPROGRESS                      =   1,
```
alex--m, line 22: Do we intend to use this? according to UCX semantics, collective_post() returns UCC_OK if the operation is completed (no need to call test()) and UCC_INPROGRESS if it started successfully...

src/api/ucc.h
alex--m: Other stuff possibly missing: limitations/capacity. How do we know if a certain team can only support up to "count == 1000" or reduce only integers?

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+/**
+ * @brief UCC library initializatoin parameters
+ */
+
+typedef struct ucc_lib_params {
+    uint64_t                mask;
+    ucc_thread_mode_t       thread_mode;
+    ucc_coll_type_t         coll_types;
+    ucc_reduction_op_t      reduction_types;
+    ucc_coll_sync_type_t    sync_type;
```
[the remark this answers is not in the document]
alex--m: Also, we should make room here for "modifiers" which apply to multiple collective operation types. These come to mind:

- Variadic (GatherV)
- Variable-datatype (GatherW)
- Neighborhood collectives
- Stable reduction (for Reduce/Allreduce)

src/api/ucc_status.h
```diff
@@ -0,0 +1,35 @@
...
+    UCC_OPERATION_INITIALIZED           =   2,
+    UCC_ERR_OP_NOT_SUPPORTED            =   3,
+    UCC_ERR_NOT_IMPLEMENTED             =   4,
+    UCC_ERR_INVALID_PARAM               =   5,
+    UCC_ERR_NO_MEMORY                   =   6,
+    UCC_ERR_NO_RESOURCE                 =   7,
+
+    UCC_ERR_LAST                        = -100,
+} ucc_status_t;
+
```
alex--m, line 34: UCX also has a "status#->string" function here, I think we should, too.

src/api/ucc.h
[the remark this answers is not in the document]
vspetrov: @manjugv So, why not option 3 from the discussion? I mean, why not uint64_t? Why do we want to bother with aint definition?

src/api/ucc.h
vspetrov: it is definitely optional and off by default. Supposed to be used for "unordered" collectives (commonly used in ML frameworks).

alex--m: Sounds good, but In that case - it should probably be indicated by a flag or something...

src/api/ucc.h
vspetrov: oob_collective struct has a "coll_info" argument which is different for every team/communicator. This is a context used to run oob collective.

src/api/ucc.h
alex--m: I agree with Valentin, I think it's much more "readable" with the argument, and it will also let us make chage this semantic in the future if we want.

src/api/ucc.h
alex--m: Valentin explained it to me during the weekly call, but I'm still not sure about the address resolution flow. Suppose ep is contiguous/sequential - how to translate from "ep=2" to address (e.g. QPN)? If it's just a random 8-byte ID - does the address resolution flow have to support the values of "ep" from the other peers?

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: We could. In the WG, there was more interest in this, as 32 bit was the most common case.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    ucc_count_t     *dst_counts;
+    ucc_aint_t      *dst_displacements;
+    ucc_datatype_t  src_datatype;
+    ucc_datatype_t  dst_datatype;
+    uint64_t        flags;
+} ucc_coll_buffer_info_t;
+
+typedef struct ucc_reduction_info {
+    ucc_datatype_t      dt;
+    ucc_reduction_op_t  op;
```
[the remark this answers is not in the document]
manjugv: It should be related. Agree.

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: @alex--m ? IIRC, this was your preference.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+} ucc_coll_sync_type_t;
+
+/**
+ * @brief UCC library initializatoin parameters
+ */
+
+typedef struct ucc_lib_params {
+    uint64_t                mask;
+    ucc_thread_mode_t       thread_mode;
+    ucc_coll_type_t         coll_types;
```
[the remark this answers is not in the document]
manjugv: Agree.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+typedef enum {
+    UCC_BARRIER            = UCC_BIT(0),
+    UCC_BCAST              = UCC_BIT(1),
+    UCC_ALLREDUCE          = UCC_BIT(2),
+    UCC_REDUCE             = UCC_BIT(3),
+    UCC_ALLTOALL           = UCC_BIT(4),
+    UCC_ALLGATHER          = UCC_BIT(5),
+    UCC_GATHER             = UCC_BIT(6),
+    UCC_SCATTER            = UCC_BIT(7),
+    UCC_MCAST              = UCC_BIT(8),
```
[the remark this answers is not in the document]
manjugv: We don't need this now. Will remove it.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+    UCC_FANIN              = UCC_BIT(9),
+    UCC_FANOUT             = UCC_BIT(10),
+    UCC_COLL_LAST          = UCC_BIT(11)
```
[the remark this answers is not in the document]
manjugv: Yes, no need for it.

src/api/ucc.h
manjugv: Sure. I'm fine with it.

src/api/ucc.h
manjugv: Sure. I'm fine with it.

src/api/ucc.h
manjugv: As I mentioned, I don't have a stronger opinion. I agree, it is cleaner also.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+ *                   Library initialization and finalize 
+ * *************************************************************
+ */
+
+typedef enum {
+    UCC_THREAD_MODE       = UCC_BIT(0),
+    UCC_COLL_TYPES        = UCC_BIT(1),
+    UCC_REDUCTION_TYPES   = UCC_BIT(2),
+    UCC_SYNC_TYPE         = UCC_BIT(3) 
+} ucc_lib_params_mask_t;
```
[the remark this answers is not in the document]
manjugv: Sure. Works for me.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+} ucc_context_type_t;
+
+enum ucc_context_attribs_field {
+    UCC_CONTEXT_ATTRIBS_TYPE                   = UCC_BIT(0),
+    UCC_CONTEXT_ATTRIBS_COLL_SYNC_TYPE         = UCC_BIT(1),
+    UCC_CONTEXT_ATTRIBS_COLL_OOB               = UCC_BIT(2)
+};
+
+enum ucc_context_params_field {
+    UCC_CONTEXT_PARAMS_TYPE                   = UCC_BIT(0),
```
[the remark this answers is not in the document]
manjugv: Sure. Works for me.

src/api/ucc.h
```diff
@@ -0,0 +1,709 @@
...
+
+
+/**
+ * *************************************************************
+ *                   Teams Section 
+ * *************************************************************
+ */
+
+enum ucc_team_params_field {
+    UCC_PARAMS_POST_ORDERING          = UCC_BIT(0),
```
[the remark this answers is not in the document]
manjugv: Sure. Works for me.

src/api/ucc_version.h
```diff
@@ -0,0 +1,18 @@
...
+#define UCC_VERSION(_major, _minor) \
+	(((_major) << UCC_VERSION_MAJOR_SHIFT) | \
+	 ((_minor) << UCC_VERSION_MINOR_SHIFT))
+#define UCC_VERSION_MAJOR_SHIFT    24
+#define UCC_VERSION_MINOR_SHIFT    16
+
+
+#define UCC_API_MAJOR    1
+#define UCC_API_MINOR    0
+#define UCC_API_VERSION  UCC_VERSION(1, 0)
```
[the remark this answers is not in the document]
manjugv: Will fix it.

src/api/ucc.h
manjugv: Will fix it.

src/api/ucc.h
manjugv: Yes, the topology is something that we have not handled. In the WG, we decided to push this next version of API. So, it can be handled more comprehensively. Let's touch upon this in the next WG call.

src/api/ucc.h
manjugv: Good point. Though in the first version I don't see something beyond these two values, we can think of other variations. I think for future-proof, we should add a mask.

src/api/ucc_status.h
manjugv: Given that collectives are relatively high-latency operations, we should expect to use it.

src/api/ucc_status.h
manjugv: We could add it. I see no harm.

src/api/ucc.h
manjugv: It is defined in UCS. However, we will adapt the definition to suit our needs. I will add it to ucc_def.

src/api/ucc.h
manjugv: > Other stuff possibly missing: limitations/capacity. How do we know if a certain team can only support up to "count == 1000" or reduce only integers?

Are you referring to number of outstanding collectives ? If so, I have captured that with` uint64_t                outstanding_colls `.

In case of "reductions on integers" constraint, at the library level, we can capture with datatypes + collectives. Are you suggesting, we should provide such filters at the team level?

src/api/ucc.h
manjugv: Fixed.

src/api/ucc.h
manjugv: I'm not sure, I understand. The team is created with option of ordered or unordered collectives. Then, the tag is a required field for all unordered collectives, and it is not required for all collectives on "ordered" team. 

Let's touch upon this in the next call.

src/api/ucc.h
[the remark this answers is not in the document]
alex--m: Originally, my concern was that this structure is very big, and I was wondering if we can reduce its size. Mask is one way, but IIRC valentin suggested the type of collective dictates the specific fields to be used, so we don't need the mask if we clearly state which fields are used for each type of call.

src/api/ucc.h
alex--m: No, outstanding collectives is only one such example. My example was about an upper limit on the size of the reduced vector. For example, some transports may only support 100 operands, or only a specific type of operand. For example, some hardware supports BFLOAT16 - maybe we want to expose this? just a suggestion, I'm not sure.

src/api/ucc_status.h
alex--m: This opens up several questions about completion. UCX has send_nb(), send_nbr() and send_nbx(). The differences are related to the way the request is allocated and filled as successful (there's some kind of race condition between creation and completion, but I don't have all the details). We should learn from their experience...

manjugv: I misread your question. I thought you are asking just about UCC_INPROGRESS as it was pointing to line 22. 

The semantics that you describe for UCC_OK is not supported (or doesn't have parallel) in UCC .i.e., there is no implicit creation and destruction of the request object. The request object is always created. Remember, in UCC, the collective operation is initialized with the ucc_collective_init interface. So before calling ucc_collective_post the request object is already created. The post is really just invoking the operation. So, post is returning UCC_OK or UCC_INPROGRESS based on the operation’s status. The request is finalized only after calling ucc_collective_finalize interface. 

I will put this on tomorrow's agenda. So, if it is not clear, we can discuss it.

manjugv: Aug 19th WG Discussion: 
* Will not support implicit creation and destruction of the request object. 
* Discuss with MPICH team, if they need request object to be created by the MPICH rather than UCC

src/api/ucc.h
manjugv: Aug 19th WG Discussion: 
* Briefly discussed this and will continue the discussion in the next meeting.

src/api/ucc.h
manjugv: Aug 19th WG Discussion: 
Instead of enum+pointer, we will have a “user-defined” reduction filed to coll_op_args.

src/api/ucc.h
manjugv: Aug 19th WG Discussion: 
* Given that the structure can be expanded (mask), we can add additional filters as we find compelling use cases.

src/api/ucc.h
manjugv: Aug 19th WG Discussion: 
* A separate field for “user-defined” reductions will be added. 
* User need to pass UCC_OP_USERDEFINED option and reduction function to UCC

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: Aug 19th WG Discussion: 
* Will add a reduction wrapper/callback to address the needs of user-defined reductions and user-defined datatypes
* We can add more library tailoring operations as needed.

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: Aug 19th WG Discussion: 
* Merge the reduction_info, op_args, and ext_op_args structures. Remove mask from the buffer info. 
* Introduce buffer info structure to reflect the requirements for collectives that have single count buffers.

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: Aug 19th WG Discussion: 
No changes. The ucc_aint_t and ucc_count to stay, so both MPI 3 and MPI 4 requirements can be supported.

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: Aug 19th WG Discussion:

Merge the reduction_info, op_args, and ext_op_args structures. Remove mask from the buffer info.
Introduce buffer info structure to reflect the requirements for collectives that have single count buffers.

src/api/ucc_status.h
manjugv: Sep 9th WG discussion:

UCC currently requires the programming models to create and destroy the request. There is no implicit request object creation and destroying. During the WG discussion, Ken from the MPICH team was convinced that having explicit request API is helpful and enough for their purposes .i.e., they will hold the pointer of the request object in their (MPICH) request object. So, no changes are required here.

src/api/ucc_status.h
manjugv: Added an API

src/api/ucc.h
manjugv: Added it.

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: Added more information about flags

vspetrov (approved):

alex--m (approved):

src/api/ucc_def.h
```diff
@@ -0,0 +1,157 @@
...
+/* Reflects the definition in UCS - The i-th bit */
+#define UCC_BIT(i)               (1ul << (i))
+
+/* Reflects the definition in UCS */
+/**
+ * @ingroup UCC_UTILS
+ * @brief Print configurations
+ */
+typedef enum {
+    UCS_CONFIG_PRINT_CONFIG        = UCC_BIT(0),
```
vspetrov, line 107: UCS->UCC

src/api/ucc.h
```diff
@@ -0,0 +1,1497 @@
...
+ *  A local operation to release the resources and
+ *  cleanup. All participants that invoked @ref ucc_init should call this
+ *  routine.
+ *
+ *  @endparblock
+ *
+ *  @return Error code as defined by ucc_status_t
+ */
+
+ucc_status_t ucc_finalize(ucc_lib_h lib_p);
```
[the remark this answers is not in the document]
Sergei-Lebedev: I don't see any reason for destroy/finalize functions to return error code except for debugging. So alternatively to status return these functions might print kind of warning in case something went wrong.

src/api/ucc.h
```diff
@@ -0,0 +1,1497 @@
...
+ */
+
+#ifndef UCC_H_
+#define UCC_H_
+
+#include <api/ucc_def.h>
+#include <api/ucc_version.h>
+#include <api/ucc_status.h>
+#include <stdio.h>
+#include <stdbool.h>
```
Sergei-Lebedev, line 16: Does this mean that we stick to C99 and is it safe to use C99 bool with C++?

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: What is advantage of returning void ? If you don't need it, you can always ignore it. :) I'm missing something.

src/api/ucc.h
```diff
@@ -0,0 +1,1497 @@
...
+ *  the collective operation. ucc_team_create_test operation is used to learn
+ *  the status of the new team handle. On error, the team handle will not
+ *  be created and corresponding error code as defined by ucc_status_t is
+ *  returned.
+ *
+ *  @endparblock
+ *
+ *  @return Error code as defined by ucc_status_t
+ */
+ucc_status_t ucc_team_create_post(ucc_context_h context,
```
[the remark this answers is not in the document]
manjugv: Yes. I thought, we kind of pushed it down the lane. May be, I'm remembering wrong. But, I think we might be better of adding it now. I will fix it.

src/api/ucc_status.h
```diff
@@ -0,0 +1,42 @@
...
+    UCC_INPROGRESS                      =   1,
+
+    /* Operation initialized but not posted */
+    UCC_OPERATION_INITIALIZED           =   2,
+    UCC_ERR_OP_NOT_SUPPORTED            =   3,
+    UCC_ERR_NOT_IMPLEMENTED             =   4,
+    UCC_ERR_INVALID_PARAM               =   5,
+    UCC_ERR_NO_MEMORY                   =   6,
+    UCC_ERR_NO_RESOURCE                 =   7,
+    UCC_ERR_LAST                        = -100,
```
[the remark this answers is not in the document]
manjugv: Sure. I don't have any issues with it. @vspetrov @alex--m ?

src/api/ucc.h
```diff
@@ -0,0 +1,1497 @@
...
+    UCC_BARRIER            = UCC_BIT(0),
+    UCC_BCAST              = UCC_BIT(1),
+    UCC_ALLREDUCE          = UCC_BIT(2),
+    UCC_REDUCE             = UCC_BIT(3),
+    UCC_ALLTOALL           = UCC_BIT(4),
+    UCC_ALLGATHER          = UCC_BIT(5),
+    UCC_GATHER             = UCC_BIT(6),
+    UCC_SCATTER            = UCC_BIT(7),
+    UCC_FANIN              = UCC_BIT(8),
+    UCC_FANOUT             = UCC_BIT(9)
```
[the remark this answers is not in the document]
manjugv: Sure. Will add it.

src/api/ucc.h
manjugv: @Sergei-Lebedev I don't understand your question, can you elaborate. Remember, we are using this only at the interface not for data transfer.  Just FYI, https://en.cppreference.com/w/cpp/language/types

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: On #2, the previous version had it, but WG decided to remove UCC_COLL_TYPE_LAST. Also, UCC_COLL_TYPE_ALL, can't we define it as macro ?

src/api/ucc.h
Sergei-Lebedev: 1. Because stdbool is only available in C99 and we have stdbool.h in main header file it means that every applications that uses ucc will have to be compiled with c99 or c11 support.
2. _Bool a.k.a bool in c99 and c++ bool are in fact different types and my question is it safe to rely on it, for instance when ucc is precompiled and different c++ compiler is used to build user application

src/api/ucc.h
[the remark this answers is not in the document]
Sergei-Lebedev: Sure it can be a macro, just if we have UCC_COLL_TYPE_LAST then

UCC_COLL_TYPE_ALL = UCC_BIT(UCC_COLL_TYPE_LAST)-1

src/api/ucc.h
manjugv: Are you saying this is going to be an issue or just asking? I don’t know if this is going to be an issue. It is only one place we need this. So, it not a huge blocker. We have three options

1. We can be conservative and change bool -> _Bool
2. Other option is to define our own bool UCC_DT_TYPE_BOOL
3. Leave as it is and deal with it later

src/api/ucc_status.h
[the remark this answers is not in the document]
vspetrov: sounds good to me

src/api/ucc.h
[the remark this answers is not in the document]
vspetrov: COLL_TYPE_LAST can only defined in the enum of integers (not BITs), right? does this mean that we want to add another enum? Or, do you mean hardcode like "#define COLL_TYPE_LAST 10" ?

src/api/ucc_status.h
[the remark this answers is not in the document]
manjugv: Added the changes.

src/api/ucc.h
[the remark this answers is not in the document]
manjugv: Sep 30th: As discussed in the WG, no action for now.

src/api/ucc.h
manjugv: Sep 30th WG Meeting: Change bool to uint32_t

PR #5, opened 2020-10-01
Documentation infrastructure 

manjugv:
### What ?

- GNU build infrastructure limited to make docs
- Doxygen infrastructure to build the documentation from the header files
  - Supports both pdf (via latex) and HTML
- Documentation 
   - Overview for each of the section - design goals, library, teams, context, and collectives  
   - Component architecture and image files

### How to build ?  

`./autogen.sh && ./configure --prefix=<path-to-install>  && make docs`


docs/doxygen/ucc_doxygen.conf
```diff
@@ -0,0 +1,2541 @@
...
+#
+# Note: Only use a user-defined header if you know what you are doing! The
+# following commands have a special meaning inside the header: $title,
+# $datetime, $date, $doxygenversion, $projectname, $projectnumber,
+# $projectbrief, $projectlogo. Doxygen will replace $title with the empty
+# string, for the replacement values of the other commands the user is referred
+# to HTML_HEADER.
+# This tag requires that the tag GENERATE_LATEX is set to YES.
+
+LATEX_HEADER           =   docs/doxygen/header.tex
```
vspetrov, line 1787: add $(SRCDIR) -> $(SRCDIR)/docs/doxygen/header.tex. Otherwise build from the outside of the src tree fails.

docs/doxygen/header.tex
```diff
@@ -0,0 +1,142 @@
...
+             bookmarksnumbered=true,
+             pdfencoding=unicode
+            }
+\pagenumbering{roman}
+\begin{titlepage}
+\vspace*{7cm}
+\begin{center}%
+    {\Large Unified Collective Communications (UCC) Specification}\\
+\vspace*{0.5cm}
+{\small Version @MAJOR\_VERSION@.@MINOR\_VERSION@}\\
```
vspetrov, line 130: These @MAJOR\_VERSION@.@MINOR\_VERSION@ objects are not expanded. The output in the result pdf is same:
"Version @MAJOR_VERSION@.@MINOR_VERSION@"

Need some shell/perl majic here.

vspetrov:
@manju, docs are not installed into <install_path> - is that expected behavior?

docs/doxygen/ucc_doxygen.conf
manjugv: Good catch.

manjugv:
> @manju, docs are not installed into <install_path> - is that expected behavior?

Yes.

docs/doxygen/header.tex
manjugv: Apparently, there is an easier way. I have fixed it. Thanks!

vspetrov (approved):

alex--m (approved):

PR #6, opened 2020-10-06
ucc/core: constructor & ucc_lib_init/destroy

vspetrov:
Minimal code to build libucc.so. constructor fn loads ucc_ccm_<name>.so CollectiveCoMponents plugins from the <install>lib/ucc location. No components so far in this PR.


configure.ac
```diff
@@ -117,4 +114,32 @@ AS_IF([test "x$with_docs_only" = xyes],
...
 
 
+AC_ARG_ENABLE([debug],
+              AS_HELP_STRING([--enable-debug], [Enable extra debugging code (default is NO).]),
+              [], [enable_debug=no])
+
+if test $enable_debug = yes; then
+	AC_DEFINE([ENABLE_DEBUG], [1], [Enable debugging code])
+	CFLAGS="$CFLAGS -O0 -g3"
+	AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_TRACE_POLL], [Highest log level])
```
alex--m, line 124: Why compile-time if we can do run-time, as part of params?

configure.ac
```diff
@@ -117,4 +114,32 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+else
+	AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_INFO], [Highest log level])
+	CFLAGS="$CFLAGS -O3 -DNDEBUG"
+fi
+
+AM_CONDITIONAL([HAVE_UCX], [false])
```
alex--m, line 130: shouldn't this be inside ucx.m4?

configure.ac
```diff
@@ -117,4 +114,32 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+m4_include([m4/ucx.m4])
+CHECK_UCX
```
alex--m, line 132: Food for thought: UCG indirectly depends on UCX as well... what if it's on a different UCX? not urgent, but I need to cover that sooner or later

src/api/ucc_status.h
```diff
@@ -31,6 +31,7 @@ typedef enum {
     UCC_ERR_INVALID_PARAM               =   -3,
     UCC_ERR_NO_MEMORY                   =   -4,
     UCC_ERR_NO_RESOURCE                 =   -5,
+    UCC_ERR_NO_MESSAGE                  =   -6,
```
alex--m, line 34: not clear what this means... could you please add a comment?

src/ccm/ucc_ccm.h
```diff
@@ -0,0 +1,103 @@
...
+    ucc_ccm_iface_t *iface;
+    ucc_ccm_lib_t   *ccm_lib;
+} ucc_ccm_context_config_t;
+extern ucs_config_field_t ucc_ccm_context_config_table[];
+
+typedef struct ucc_ccm_iface {
+    ucc_component_iface_t          super;
+    char*                          name;
+    int                            priority;
+    ucc_lib_params_t               params;
```
alex--m, line 42: probably needs to be a pointer... if for no other reason - at least for ABI compatibility (if params grow)

src/ccm/ucc_ccm.h
```diff
@@ -0,0 +1,103 @@
...
+typedef struct ucc_ccm_team {
+    ucc_ccm_iface_t *iface;
+    ucc_ccm_lib_t   *ccm_lib;
+} ucc_ccm_team_t;
+
+#define UCC_CCM_LOG(_log_component, _level, _fmt, ...) do {             \
+        ucs_log_component(_level, &_log_component, _fmt, ## __VA_ARGS__); \
+    } while (0)
+
+#define ucc_ccm_error(_lib, _fmt, ...)        \
```
alex--m, line 82: I think we should seperate this into a separate logging header, common to all components (not just CCM/CL).

src/core/Makefile.am
```diff
@@ -0,0 +1,15 @@
...
+
+libucc_core_la_CPPFLAGS = $(AM_CPPFLAGS)
+libucc_core_la_LDFLAGS  = -ldl -version-info $(SOVERSION) --no-undefined
+
+libucc_core_la_SOURCES = \
+	ucc_constructor.c \
+	ucc_lib.c         \
+	ucc_lib.h         \
+	ucc_global_opts.c \
+	ucc_global_opts.h
```
alex--m, line 15: Don't headers go outside SOURCES? obviously it works, but IIRC UCX puts most in noinst_HEADERS or similar.

src/core/ucc_constructor.c
```diff
@@ -0,0 +1,73 @@
...
+    if (strlen(ucc_lib_global_config.ccm_path) == 0) {
+        get_default_lib_path();
+    }
+    ucc_info("UCC components path: %s", ucc_lib_global_config.ccm_path);
+
+    if (!ucc_lib_global_config.ccm_path) {
+        ucc_error("Failed to get ucc library path. set UCC_CCM_PATH.\n");
+        return;
+    }
+    ucc_components_load("ccm",
```
alex--m, line 64: Should probably also be common to all component frameworks we will have. Also, though we didn't have a conclusive decision on this - what do you think about having a "component" and a subclass "dl_component" with "component" as super?

src/core/ucc_global_opts.c
```diff
@@ -0,0 +1,24 @@
...
+
+ucs_config_field_t ucc_lib_global_config_table[] = {
+    {"LOG_LEVEL", "warn",
+     "UCC logging level. Messages with a level higher or equal to the selected "
+     "will be printed.\n"
+     "Possible values are: fatal, error, warn, info, debug, trace, data, func, poll.",
+     ucs_offsetof(ucc_config_t, log_component),
+     UCS_CONFIG_TYPE_LOG_COMP},
+
+    {"COMPONENT_PATH", "",
```
alex--m, line 16: does this work like LD_LIBRARY_PATH, or differently?

src/core/ucc_lib.c
```diff
@@ -0,0 +1,220 @@
...
+                          ucc_lib_h *ucc_lib)
+{
+    ucs_status_t status;
+    ucc_lib_info_t *lib;
+
+    if (ucc_lib_data.n_ccms_loaded == 0) {
+        return UCC_ERR_NO_MESSAGE;
+    }
+
+    lib = malloc(sizeof(*lib));
```
alex--m, line 126: we'd probably want to do something like "memtrack.h" for UCX... for now maybe doing `#define ucs_malloc malloc` is enough, but if we're exploring UCS - memory tracking is high on my list. :)

src/core/ucc_lib.c
```diff
@@ -0,0 +1,220 @@
...
+    //ucs_config_parser_warn_unused_env_vars_once("UCC_");
+    return UCC_OK;
+
+error:
+    *ucc_lib = NULL;
+    if (lib) free(lib);
+    return status;
+}
+
+ucc_status_t ucc_lib_config_read(const char *env_prefix,
```
alex--m, line 160: Why not re-use UCX code for this?

src/core/ucc_lib.c
```diff
@@ -0,0 +1,220 @@
...
+        CHECK_LIB_CONFIG_CAP(coll_types,   COLL_TYPES);
+        ccm_config  = malloc(ccm_iface->ccm_lib_config.size);
+        ucs_config_parser_fill_opts(ccm_config, ccm_iface->ccm_lib_config.table,
+                                    config->full_prefix, ccm_iface->ccm_lib_config.prefix, 0);
+        status = ccm_iface->init(params, config, ccm_config, &ccm_lib);
+
+        if (UCS_OK != status) {
+            ucs_config_parser_release_opts(ccm_config, ccm_iface->ccm_lib_config.table);
+            ucc_error("lib_init failed for component: %s\n", ccm_iface->name);
+            goto error;
```
alex--m, line 96: I think if one component fails - we should skip it and continue.

src/ccm/ucc_ccm.h
alex--m: I just ran into ucc_log.h ... why not put it all there?

src/ccm/ucc_ccm.h
```diff
@@ -0,0 +1,103 @@
...
+    ucs_log_component_config_t log_component;
+    int                        priority;
+} ucc_ccm_lib_t;
+
+typedef struct ucc_ccm_context {
+    ucc_ccm_lib_t *ccm_lib;
+} ucc_ccm_context_t;
+
+typedef struct ucc_ccm_team {
+    ucc_ccm_iface_t *iface;
```
alex--m, line 74: If "iface" is what I think it is - shouldn't it be part of the API for our "context storage"? I assume this iface is something multiple CCMs/CLs can share?

configure.ac
vspetrov: yes, it is already inside.

configure.ac
vspetrov: @alex--m UCS logger requires UCS_MAX_LOG_LEVEL macro defined during build time. So, we have to define it to some value. We 100% followed UCX logic here. Enabling high max log level in debug build (TRACE) and lower level (INFO) for release build.

configure.ac
vspetrov: Do you mean another UCX version ? Lets discuss it in WG, i think this is tricky.

src/api/ucc_status.h
vspetrov: This is general purpose return code without explicit error specification. Same exists in ucx. I'll add comment. (this is will be another PR)

src/core/ucc_lib.c
vspetrov: We actually exactly re-using UCX code. This "ucc_lib_config_read" is our ucc.h API and its implementation is essentially ucs_config_parser_fill_opts but with minor change: we add "UCC_" prefix.

vspetrov:
Closing this. The code has been merged into master via set of smaller PRs.

PR #8, opened 2020-10-07
Use clang-format for code style check

manjugv:
### What ?

Use clang-format for code style check

### How ?

`clang-format -style=file *.c`

**Exhaustive list of options and description:** 

https://clang.llvm.org/docs/ClangFormatStyleOptions.html

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,126 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortLambdasOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: MultiLine
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:   
+  AfterCaseLabel:  false
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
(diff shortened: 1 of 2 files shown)


alex--m (approved):
I didn't really try it, but I think this direction is great! :)

.clang-format
```diff
@@ -0,0 +1,126 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: false
```
vspetrov, line 7: AlignConsecutiveDeclarations: ->true

.clang-format
```diff
@@ -0,0 +1,126 @@
...
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyBreakTemplateDeclaration: 10
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Right
+ReflowComments:  true
+SortIncludes:    true
```
vspetrov, line 101: SortIncludes: ->false

manjugv:
I have experimented more with this. It will not give us a complete solution. There are edge cases that are not desirable. So, we should take the output from clang-format and look at examples (docs/codestyle.md) as a guide to further fix it. Over time, we might have a better way to automate. @vspetrov @Sergei-Lebedev @alex--m, Thoughts ?

vspetrov:
i think it is fine for now. The main problem i see already is that it formats the whole file. If there is some "undesirable" corner case it will appear again and again each time the file is formatted. Right now i have a script that checks which files are changed in the PR that i'm preparing and it runs clang-format on those files. But i can't make it work only on a subset of a file. this i inconvenient - need to go through the whole file after formatting to revert "false-positive" clang changes. It would be great to have  a way to apply the formatting only on the "diff" of the current change. But i don't know how to achieve this.

vspetrov (approved):

manjugv:
> i think it is fine for now. The main problem i see already is that it formats the whole file. If there is some "undesirable" corner case it will appear again and again each time the file is formatted. Right now i have a script that checks which files are changed in the PR that i'm preparing and it runs clang-format on those files. But i can't make it work only on a subset of a file. this i inconvenient - need to go through the whole file after formatting to revert "false-positive" clang changes. It would be great to have a way to apply the formatting only on the "diff" of the current change. But i don't know how to achieve this.

I'm not sure I understand whether you need (1) a part of the file to be formatted or (2) subset of total files. For (1) you have an option of` "--lines=<string> "` that you can work.

vspetrov:
> > i think it is fine for now. The main problem i see already is that it formats the whole file. If there is some "undesirable" corner case it will appear again and again each time the file is formatted. Right now i have a script that checks which files are changed in the PR that i'm preparing and it runs clang-format on those files. But i can't make it work only on a subset of a file. this i inconvenient - need to go through the whole file after formatting to revert "false-positive" clang changes. It would be great to have a way to apply the formatting only on the "diff" of the current change. But i don't know how to achieve this.
> 
> I'm not sure I understand whether you need (1) a part of the file to be formatted or (2) subset of total files. For (1) you have an option of` "--lines=<string> "` that you can work.

yes i want (1). i'll try --lines option.

PR #11, opened 2020-10-08
ucx.m4 

vspetrov:
**What?**
- ucx.m4 script
- "enable-debug" option in configure.ac
- added "with-ucx" to CI

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -2,6 +2,9 @@ name: CI
 
 on: [push, pull_request]
 
+env:
+  OPEN_UCX_LINK: https://github.com/openucx/ucx
+  OPEN_UCX_BRANCH: v1.9.x
 jobs:
   tests:
     runs-on: ubuntu-latest
@@ -10,9 +13,16 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends doxygen doxygen-latex
+    - name: Get UCX
+      run: git clone ${OPEN_UCX_LINK} -b ${OPEN_UCX_BRANCH} /tmp/ucx
+    - name: Build UCX
+      run: |
+        cd /tmp/ucx && ./autogen.sh
+        ./contrib/configure-release --without-java --disable-numa --prefix $PWD/install
+        make -j install
     - uses: actions/checkout@v1
     - name: Build UCC
       run: |
         ./autogen.sh
-        ./configure --prefix=$PWD/install
-        make install
+        ./configure --prefix=$PWD/install --with-ucx=/tmp/ucx/install
+        make -j`nproc` install
--- a/configure.ac
+++ b/configure.ac
@@ -21,6 +21,10 @@ define([ucc_git_sha], esyscmd([sh -c "git rev-parse --short HEAD"]))
 define([libucc_so_version], 1:0:0)
 
 AC_INIT([ucc], [ucc_ver_major.ucc_ver_minor])
+: ${CPPFLAGS=""}
+: ${CFLAGS=""}
+: ${CXXFLAGS=""}
+
 AC_USE_SYSTEM_EXTENSIONS
 AC_GNU_SOURCE
 AC_CONFIG_HEADERS([config.h])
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -10,9 +14,14 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends doxygen doxygen-latex
+    - name: Get HPCX
+      run: wget "${HPCX_PKG_LINK}/${HPCX_PKG_NAME}.tbz" -P /tmp
```
Sergei-Lebedev, line 18: Do we really need HPCX now? Maybe git clone and build only ucx instead?

configure.ac
```diff
@@ -117,4 +114,29 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+if test $enable_debug = yes; then
+	AC_DEFINE([ENABLE_DEBUG], [1], [Enable debugging code])
+	CFLAGS="$CFLAGS -O0 -g3"
+	AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_TRACE_POLL], [Highest log level])
+else
+	AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_INFO], [Highest log level])
+	CFLAGS="$CFLAGS -O3 -DNDEBUG"
+fi
+
+m4_include([m4/ucx.m4])
```
Sergei-Lebedev, line 130: there is already config/m4 directory, isn't it better to move ucx.m4 to it?

configure.ac
```diff
@@ -117,4 +114,29 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+CHECK_UCX
+AC_MSG_RESULT([UCX support: $ucx_happy])
```
Sergei-Lebedev, line 132: since ucx/ucs is a requirement then configure script should fail if ucx wasn't found

vspetrov: agreed. fixed. Added minimal required version check: UCX 1.9 - before that UCS does not have required config_read functionality.

configure.ac
vspetrov: yes

.github/workflows/main.yaml
vspetrov: done. Building against v1.9. I don't think we want to use master (smth might change easily in ucx/master and it may break our build here)

Sergei-Lebedev (approved):

alex--m (commented):
the rest looks good to me

configure.ac
```diff
@@ -117,4 +115,31 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+	CFLAGS="$CFLAGS -O3 -DNDEBUG"
+fi
+
+CHECK_UCX
+AC_MSG_RESULT([UCX support: $ucx_happy])
+if test $ucx_happy != "yes"; then
+   AC_MSG_ERROR([UCX is not available])
+fi
+
+includes="-I${UCC_TOP_SRCDIR}/src -I${UCC_TOP_SRCDIR}/src/core"
```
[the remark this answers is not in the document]
vspetrov: @alex--m lets only keep {UCC_TOP_SRCDIR}/src and use relative include paths from there. i'll fix.

.github/workflows/main.yaml
```diff
@@ -10,9 +13,16 @@ jobs:
...
+      run: git clone ${OPEN_UCX_LINK} -b ${OPEN_UCX_BRANCH} /tmp/ucx
+    - name: Build UCX
+      run: |
+        cd /tmp/ucx && ./autogen.sh
+        ./contrib/configure-release --without-java --disable-numa --prefix $PWD/install
+        make -j install
     - uses: actions/checkout@v1
     - name: Build UCC
       run: |
         ./autogen.sh
```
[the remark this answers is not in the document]
vspetrov: @alex--m i'm not sure i understand what you mean? You want to have "with-ucg" as autogen option? Why is it simpler? Do you want to add it  right now in the CI?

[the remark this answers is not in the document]
alex--m: For the time being, UCG is an optional part of UCX (master). To build UCX with UCG - one needs to pass `--with-ucg` to autogen and `--enable-ucg` to configure. Long-term this will change, but short-term I don't see any other way to integrate UCG code into UCC. Specifically, UCG uses a lot of internal UCX APIs (like stats.h and profiling.h) which UCX does not expose publically (and might never will). If you don't mind, I recommend adding both to the UCC CI, otherwise the CI will not be able to test the "reactive" component any time soon...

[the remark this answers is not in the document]
vspetrov: @alex--m lets add ucg related flags when we have the corresponding component, as separate PR - right now there is no ucg detection in configure.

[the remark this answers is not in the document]
alex--m: I suppose that's ok.

alex--m (approved):

PR #14, opened 2020-10-09
UCC constructor/destructor

vspetrov:
Depends on #8 and #11 

**What?**

- ucc_constructor/ucc_destructor. Constructor just reads global options now.
- Global options definition: log_level, component path

### DIFF
--- a/src/api/ucc_status.h
+++ b/src/api/ucc_status.h
@@ -17,13 +17,11 @@
 
 typedef enum {
     /* Operation completed successfully */
-    UCC_OK                              =   0,
+    UCC_OK                              =    0,
 
-    /* Operation is posted and is in progress */
-    UCC_INPROGRESS                      =   1,
+    UCC_INPROGRESS                      =    1, /*!< Operation is posted and is in progress */
 
-    /* Operation initialized but not posted */
-    UCC_OPERATION_INITIALIZED           =   2,
+    UCC_OPERATION_INITIALIZED           =    2, /*!< Operation initialized but not posted */
 
     /* Error status codes */
     UCC_ERR_OP_NOT_SUPPORTED            =   -1,
@@ -31,6 +29,7 @@ typedef enum {
     UCC_ERR_INVALID_PARAM               =   -3,
     UCC_ERR_NO_MEMORY                   =   -4,
     UCC_ERR_NO_RESOURCE                 =   -5,
+    UCC_ERR_NO_MESSAGE                  =   -6, /*!< General purpose return code without specific error */
     UCC_ERR_LAST                        = -100,
 } ucc_status_t;
 
--- a/src/core/ucc_constructor.c
+++ b/src/core/ucc_constructor.c
@@ -1,0 +1,64 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "config.h"
+#include "ucc_global_opts.h"
+#include "utils/ucc_malloc.h"
+#include <link.h>
+#include <dlfcn.h>
+#include <string.h>
+
+static int callback(struct dl_phdr_info *info, size_t size, void *data)
+{
+    char *str;
+    char *component_path;
+    if (NULL != (str = strstr(info->dlpi_name, "libucc.so"))) {
+        int pos        = (int)(str - info->dlpi_name);
+        component_path = (char *)ucc_malloc(pos + 8);
+        if (!component_path) {
+            //TODO add ucc_error(msg) when logging as added
+            return -1;
+        }
+        ucc_strncpy_safe(component_path, info->dlpi_name, pos);
+        component_path[pos] = '\0';
+        strcat(component_path, "ucc");
+        ucc_global_config.component_path =
+            ucc_global_config.component_path_default = component_path;
+    }
+    return 0;
+}
+
+static void get_default_lib_path()
+{
+    dl_iterate_phdr(callback, NULL);
+}
+
+ucc_status_t ucc_constructor(void)
+{
+    ucc_status_t status;
+    if (!ucc_global_config.initialized) {
+        ucc_global_config.initialized            = 1;
+        ucc_global_config.component_path_default = NULL;
+
+        status = ucc_config_parser_fill_opts(
+            &ucc_global_config, ucc_global_config_table, "UCC_", NULL, 1);
+        if (UCC_OK != status) {
+            //TODO add ucc_error(msg) when logging is added
+            return status;
+        }
+        if (strlen(ucc_global_config.component_path) == 0) {
+            get_default_lib_path();
+        }
+    }
+    return UCC_OK;
+}
+
+__attribute__((destructor)) static void ucc_destructor(void)
+{
+    if (ucc_global_config.component_path_default) {
+        ucc_free(ucc_global_config.component_path_default);
+    }
+}
--- a/src/core/ucc_global_opts.c
+++ b/src/core/ucc_global_opts.c
@@ -1,0 +1,30 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "ucc_global_opts.h"
+#include "utils/ucc_compiler_def.h"
+
+ucc_global_config_t ucc_global_config = {
+    .log_component  = {UCC_LOG_LEVEL_WARN, "UCC"},
+    .component_path = "",
+    .initialized    = 0,
+};
+
+ucc_config_field_t ucc_global_config_table[] = {
+    {"LOG_LEVEL", "warn",
+     "UCC logging level. Messages with a level higher or equal to the selected "
+     "will be printed.\n"
+     "Possible values are: fatal, error, warn, info, debug, trace, data, func, "
+     "poll.",
+     ucc_offsetof(ucc_global_config_t, log_component),
+     UCC_CONFIG_TYPE_LOG_COMP},
+
+    {"COMPONENT_PATH", "", "Specifies dynamic components location",
+     ucc_offsetof(ucc_global_config_t, component_path), UCC_CONFIG_TYPE_STRING},
+
+    NULL};
+UCC_CONFIG_REGISTER_TABLE(ucc_global_config_table, "UCC global", NULL,
+                          ucc_global_config)
--- a/src/core/ucc_global_opts.h
+++ b/src/core/ucc_global_opts.h
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifndef UCC_GLOBAL_OPTS_H_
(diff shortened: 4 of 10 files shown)


src/utils/ucc_malloc.h
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (C) Mellanox Technologies Ltd. 2001-2020.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifndef UCC_MALLOC_H_
+#define UCC_MALLOC_H_
+
+#include "config.h"
+
+#define ucc_malloc malloc
```
manjugv, line 11: @vspetrov Do we expect to use ucs for the definitions of these functions?

src/core/ucc_constructor.c
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2001-2020.  ALL RIGHTS RESERVED.
```
manjugv, line 2: @vspetrov Is it really 2001-2020?

src/utils/ucc_malloc.h
vspetrov: If/when ucs exposes ucs/debug/memtrack.h as public header, we can switch to ucs_malloc/calloc/realloc. This is why i suggest we use UCC level define.

configure.ac
```diff
@@ -117,4 +115,31 @@ AS_IF([test "x$with_docs_only" = xyes],
...
 
 
+AC_ARG_ENABLE([debug],
+              AS_HELP_STRING([--enable-debug], [Enable extra debugging code (default is NO).]),
+              [], [enable_debug=no])
+
+if test $enable_debug = yes; then
+	AC_DEFINE([ENABLE_DEBUG], [1], [Enable debugging code])
+	CFLAGS="$CFLAGS -O0 -g3"
+	AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_TRACE_POLL], [Highest log level])
```
manjugv, line 125: Should we move this to a separate file (config/ucx.m4 or config/ucs.m4 ? We should avoid mixing UCC and UCS?

src/utils/ucc_malloc.h
manjugv: Agree.

src/core/ucc_global_opts.h
```diff
@@ -0,0 +1,30 @@
...
+    /* Log level above which log messages will be printed*/
+    ucs_log_component_config_t log_component;
+
+    /* Coll component libraries path */
+    char *component_path;
+    char *component_path_default;
+} ucc_global_config_t;
+
+extern ucc_global_config_t ucc_global_config;
+extern ucs_config_field_t  ucc_global_config_table[];
```
manjugv, line 28: Using UCS at some places directly and hiding/not used in some places (like malloc) make me uncomfortable. More and more, I think about this, we need a consistent approach. Should we create a UCS component under utilities and then use all UCS services underneath it? I know extra function indirection seems laborious and not ideal, but from a maintenance / future extensions perspective, we might be better off.

vspetrov: So, seems we are back to where we started long ago. I remember we made a decision that we rely on UCS for all utilities. This is why we use it directly. For example, you are saying we want to abstract ucs_config_* set of APIs and then implement them using other components. There is no other existing implementation right now for this. Do you mean re-writing all this code ourselves inside ucc? Then why use services library in the first place? Malloc is another example - we have std "malloc" implementation already, but we don't have ucs_malloc exposed yet. I would use ucs_malloc directly if i had it already exposed in public ucs headers.

Lets discuss this in WG. But to me, abstracting UCS has little sense. Potential benefit is likely to never be realised. I mean, i don't see we will implement another services library internally.

vspetrov: Update the PR as we discussed. Put all the ucs_ utilities under src/utils and re-defined them with ucc_ namespace.

alex--m (approved):

src/api/ucc_status.h
```diff
@@ -31,6 +31,8 @@ typedef enum {
     UCC_ERR_INVALID_PARAM               =   -3,
     UCC_ERR_NO_MEMORY                   =   -4,
     UCC_ERR_NO_RESOURCE                 =   -5,
+    /* General purpose return code without specific error */
+    UCC_ERR_NO_MESSAGE                  =   -6,
```
manjugv, line 35: Minor : Can you make this comment doxygen friendly? Example: https://github.com/openucx/ucc/blob/b790f356fef88981f196e828bd1ca1de5025947d/src/api/ucc.h#L982

src/core/ucc_constructor.c
```diff
@@ -0,0 +1,60 @@
...
+
+ucc_status_t ucc_constructor(void)
+{
+    ucc_status_t status;
+    if (!ucc_global_config.initialized) {
+        ucc_global_config.initialized            = 1;
+        ucc_global_config.component_path_default = NULL;
+
+        status = ucc_config_parser_fill_opts(
+            &ucc_global_config, ucc_global_config_table, "UCC_", NULL, 1);
```
manjugv, line 43: Minor: ident (I'm not sure, if this another clang-format artifact)

manjugv:
Besides minor comments, it looks good to me.

src/core/ucc_global_opts.h
```diff
@@ -0,0 +1,29 @@
...
+#include "utils/ucc_log.h"
+
+typedef struct ucc_global_config {
+    /* Log level above which log messages will be printed*/
+    ucc_log_component_config_t log_component;
+
+    /* Coll component libraries path */
+    char *component_path;
+    char *component_path_default;
+    int   initialized;
```
Sergei-Lebedev, line 21: According to ucx codestyle should be aligned by *

src/core/ucc_global_opts.c
```diff
@@ -0,0 +1,30 @@
...
+    {"LOG_LEVEL", "warn",
+     "UCC logging level. Messages with a level higher or equal to the selected "
+     "will be printed.\n"
+     "Possible values are: fatal, error, warn, info, debug, trace, data, func, "
+     "poll.",
+     ucc_offsetof(ucc_global_config_t, log_component),
+     UCC_CONFIG_TYPE_LOG_COMP},
+
+    {"COMPONENT_PATH", "", "Specifies dynamic components location",
+     ucc_offsetof(ucc_global_config_t, component_path), UCS_CONFIG_TYPE_STRING},
```
Sergei-Lebedev, line 26: UCS_CONFIG_TYPE_STRING -> UCC_CONFIG_TYPE_STRING

src/core/ucc_constructor.c
```diff
@@ -0,0 +1,60 @@
...
+#include <dlfcn.h>
+#include <string.h>
+
+static int callback(struct dl_phdr_info *info, size_t size, void *data)
+{
+    char *str;
+    char *component_path;
+    if (NULL != (str = strstr(info->dlpi_name, "libucc.so"))) {
+        int pos        = (int)(str - info->dlpi_name);
+        component_path = (char *)ucc_malloc(pos + 8);
```
Sergei-Lebedev, line 20: check return for NULL

src/core/ucc_constructor.c
```diff
@@ -0,0 +1,60 @@
...
+        strncpy(component_path, info->dlpi_name, pos);
```
Sergei-Lebedev, line 21: consider using ucs_strncpy_safe

Sergei-Lebedev (changes requested):

src/core/ucc_constructor.c
vspetrov: @manjugv what about indent here? you mean that fn arguments are on the next line after open brace? - this is how our clang format is configured.

src/utils/ucc_parser.h
```diff
@@ -0,0 +1,31 @@
...
+#define UCC_CONFIG_REGISTER_TABLE UCS_CONFIG_REGISTER_TABLE
+
+static inline ucc_status_t
+ucc_config_parser_fill_opts(void *opts, ucc_config_field_t *fields,
+                            const char *env_prefix, const char *table_prefix,
+                            int ignore_errors)
+{
+    ucs_status_t status = ucs_config_parser_fill_opts(
+        opts, fields, env_prefix, table_prefix, ignore_errors);
+    return status == UCS_OK ? UCC_OK : UCC_ERR_NO_MESSAGE;
```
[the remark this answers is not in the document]
vspetrov: Fixed. However, i couldn't add the mapping array since ucs_statuses are negative values, additionally there is no LAST value to allocate static array and i didn't want to hardcode that size. So, i've added a static inline function with switch that maps only those statuses that we have in UCC all others are mapped to UCC_ERR_NO_MESSAGE

src/core/ucc_lib.c
```diff
@@ -4,10 +4,9 @@
  * See file LICENSE for terms.
  */
 
-#ifdef HAVE_CONFIG_H
-#  include "config.h"
-#endif
+#include "config.h"
 
+#include "ucc_global_opts.h"
```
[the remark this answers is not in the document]
vspetrov: @Sergei-Lebedev  what do you mean? I was thinking that we always use quotes for internal headers and <> for external ones. What is our style ?

src/core/ucc_global_opts.h
vspetrov: @Sergei-Lebedev but this is how our clang-format does this. I think we agreed that we are not "ucx codestyle". We have our own style: clang-format + some tweaks. Do we want to fix that manually as well? I'm fine with "*" not used for alignment. @manjugv @alex--m ?

src/core/ucc_constructor.c
vspetrov: done

src/core/ucc_constructor.c
vspetrov: done

src/core/ucc_global_opts.c
vspetrov: done

src/core/ucc_lib.c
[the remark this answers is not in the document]
Sergei-Lebedev: I'm asking because we have both quotes and <> here
```C
#include "ucc_global_opts.h"
#include <api/ucc.h>
```
or we suppose api/ucc.h is external?

src/api/ucc_status.h
vspetrov: Done

src/core/ucc_global_opts.h
Sergei-Lebedev: Fine by me as well, just wanted to make sure that we don't follow ucx codestyle in this case

src/core/ucc_lib.c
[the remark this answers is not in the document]
vspetrov: fixed

Sergei-Lebedev (approved):

src/core/ucc_global_opts.h
manjugv: Fine by me.

src/core/ucc_constructor.c
manjugv: I thought so. What do you think is better here? follow clang-format or align it with the braces. I'm always used to aligning with braces. For me, I want us to be consistent. I don't really one way or another.

PR #15, opened 2020-10-12
UCC/API: version check

Sergei-Lebedev:
add function to check UCC API version compatibility
make ucc_version.h consistent with configure


src/api/ucc.h
```diff
@@ -473,6 +473,20 @@ ucc_status_t ucc_lib_config_modify(ucc_lib_config_h *config, const char *name,
...
 
 
+/**
+ *  @ingroup UCC_LIB
+ *
+ *  @brief The @ref ucc_init_version is an internal routine that checks
+ *  compatibility with particular UCC API version.
+ *  @ref ucc_init should be used to create UCC library handle.
+ */
+ucc_status_t ucc_init_version(unsigned api_major_version,
```
vspetrov, line 483: @Sergei-Lebedev does it compile fine with ucc_init_version undefined? Maybe add src/core/ucc_lib.c and put ucc_init_version with empty implementation there?

vspetrov (approved):

src/api/ucc.h
manjugv: Is there a need to expose in ucc.h ? If it is purely internal, I prefer not to expose in api/ucc.h.

Sergei-Lebedev: @vspetrov 
Fixed
@manjugv 
Yes, that is the whole idea, because when user will compile the application this application will automatically contain ucc api version. If then the application is ran with wrong ucc lib it will report error.
Same is done in UCP https://github.com/openucx/ucx/blob/5c6bd6ed6d788df6ddcbab9ecd2c662bf3fae85b/src/ucp/api/ucp.h#L1586

manjugv: Thanks for the context. 

Can you define a different Doxygen group for this, so, we will not generate it as a part of the specification? May be `@ingroup UCC_LIB_INTERNAL` or something ? 

"particular UCC API version" > particular **a** UCC API version
to create UCC library handle > to create **the** UCC library handle

Sergei-Lebedev: @manjugv fixed

manjugv (approved):

alex--m (approved):

PR #20, opened 2020-10-15
UCC API fixes

alex--m:
Minor fixes found during UCG component experiments...

### DIFF
--- a/src/api/ucc.h
+++ b/src/api/ucc.h
@@ -2,6 +2,7 @@
  * @file ucc.h
  * @date 2020
  * @copyright Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ * @copyright Copyright (C) Huawei Technologies Co., Ltd. 2020.  ALL RIGHTS RESERVED.
  *
  * See file LICENSE for terms.
  */
@@ -268,6 +269,7 @@ typedef enum {
  *  @param [in]  inoutvec       The input elements to be reduced and output of the reduction
  *  @param [in]  count          The number of elements of type "dtype" to be reduced
  *  @param [in]  dtype          Datatype passed to the reduction operation
+ *  @param [in]  op             Operation type passed to the reduction operation
  *
  *
  *  @parblock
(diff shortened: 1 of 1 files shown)


src/api/ucc.h
```diff
@@ -281,7 +283,8 @@ typedef enum {
  *  @endparblock
  */
 typedef void(*ucc_reduction_dtype_mapper_t)(void *invec, void *inoutvec,
-                                            ucc_count_t *count, ucc_datatype_t dtype);
+                                            ucc_count_t *count, ucc_datatype_t dtype,
+                                            ucc_reduction_op_t op);
```
manjugv, line 287: @alex--m This is passed to the library initialization. So, how does the library user know, what operation to map. I'm missing something here.

alex--m: This is the callback function used by UCC to ask runtime (e.g. MPI) to handle a reduction operation during a call to MPI_Reduce/Allreduce(). To do that, UCC must specify which operation type is required (e.g. SUM), in addition to the operand type already specified (e.g. INT).

vspetrov: @manjugv @alex--m i've looked again at our user defined reductions flow, and .. i lost . where do we pass the "opaque" ptr to the upper level data type? why do we have 2 user reduction callbacks: 1 on coll_args 1 on lib params? why on lib_params it  called "dtype_mapper" ?

vspetrov: @alex--m why do we need any callback for pre-defined types like "INT" - we will handle this internally, right? We only need those callbacks for the user-defined (non standard) data types.

alex--m: @vspetrov Regarding INT - yes, we can handle this and a few other primitive types, but for a general datatype we will need this callback.

Regarding the callback in coll_args - I'm not sure myself, IIRC @manjugv added those to support custom operations defined by the user (not included in `ucc_reduction_op_t`). I don't see a reason to pass this one MPI callback for reduction (where we pass the opaque datatype ptr) on every collective call - makes much more sense to me if we pass it once during lib_init.

manjugv: Case 1: Reductions using the user-defined reduction operation on standard datatypes. In this case, the library needs only one callback, which is `ucc_userdefined_reduction_op_t`

Case 2: Reductions using the user-defined reduction operation on user-defined datatypes. In this case, the library uses both callbacks `ucc_userdefined_reduction_op_t` and `ucc_reduction_dtype_mapper_t`. The `ucc_reduction_dtype_mapper_t `is one per library instance and `ucc_userdefined_reduction_op_t` is per invocation. 

Let’s talk in the WG today If there are cases that need to be addressed.

alex--m:
We decided we'll follow-up next time, with a different approach to address the custom op issue.

PR #21, opened 2020-10-16
TEST: added Google Test (gtest)

alex--m:


alex--m:
Error is related to UCC's common code, so this PR will wait until that issue is resolved (I'm assuming If I fix it I'll cause a contfict with Val's PRs...):

````
make[1]: Entering directory '/home/runner/work/ucc/ucc/src'
  CC       core/libucc_la-ucc_lib.lo
  CC       core/libucc_la-ucc_constructor.lo
  CC       core/libucc_la-ucc_global_opts.lo
core/ucc_global_opts.c:16:48: error: missing braces around initializer [-Werror=missing-braces]
 ucc_config_field_t ucc_global_config_table[] = {
                                                ^
cc1: all warnings being treated as errors
````

**Comment by alex--m** [comment]
@vspetrov - I fixed the braces-intialization issue, looks like a gcc/clang issue after all...

**Comment by vspetrov** [comment]
@alex--m from this PR what is vanilla googletest code, what is UCS changes and what is UCC changes?
We want to review it in order to merge. But we'd like to skip reviewing google part of course. also @manjugv

**Comment by alex--m** [comment]
@vspetrov I have to admit: I almost didn't add any NEW code. This is code taken from UCX's gtest - see the license headers. For example, "gtest-all.cc" is "GTEST-native", "main.cc" is from UCX, and I just added that dummy "tl_test.cc". I'm not sure how I would review that either....

File: test/gtest/common/test.h [hunk 1]
```diff
@@ -0,0 +1,250 @@
...
+    virtual ~test_base();
+
+protected:
+    typedef enum {
+        NEW, RUNNING, SKIPPED, ABORTED, FINISHED
+    } state_t;
+
+    typedef std::vector<ucc_global_config_t> config_stack_t;
+
+    void SetUpProxy();
```
vspetrov, line 52: what is the user acse for Proxy ? How is it used?

test/gtest/common/test.h
```diff
@@ -0,0 +1,250 @@
...
+#define UCC_MT_TEST_F(test_fixture, test_name, num_threads, ...) \
+  UCC_TEST_(test_fixture, test_name, \
+            ::testing::internal::GetTypeId<test_fixture>(), \
+            num_threads, 0, "", __VA_ARGS__)
+
+
+/*
+ * Helper macro
+ */
+#define UCC_TEST_P_(test_case_name, test_name, num_threads, \
```
vspetrov, line 190: What is the difference between UCC_TEST_F and UCC_TEST_P? When should i use either of those?

test/gtest/configure.m4
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (C) Mellanox Technologies Ltd. 2001-2019.  ALL RIGHTS RESERVED.
+# Copyright (C) Huawei Technologies Co., Ltd. 2020.  All rights reserved.
+#
+# See file LICENSE for terms.
+#
+
+AC_LANG_PUSH([C++])
+
+CHECK_COMPILER_FLAG([-fno-tree-vectorize], [-fno-tree-vectorize],
```
vspetrov, line 10: why we need to check if compiler supports --diag_suppress 236 and fno-tree-vectoeize? Does not look needed?

vspetrov (approved):
@alex--m i just have a couple of generic questions. Overall looks ok to me. Lets proceed with this.

test/gtest/common/test.h
SeyedMir: If you're still wondering about this, and if this is similar to UCX, then UCC_TEST_P is the parameterized version of UCC_TEST_F. You can use UCC_TEST_P when you want to run the exact same test with different values of an input parameter.
The value of the parameter can be accessed by `GetParam()` macro in the body of UCC_TEST_P. At the end of your test file, you should use one of `INSTANTIATE_TEST_CASE_P()` macros which instantiate all parameterized tests with a given list of parameter values that are passed to `INSTANTIATE_TEST_CASE_P()`.
Hope this helps.

test/gtest/common/test_obj_size.cc
```diff
@@ -0,0 +1,37 @@
...
+
+#if ENABLE_DEBUG_DATA
+    UCC_TEST_SKIP_R("Debug data");
+#elif defined (ENABLE_STATS)
+    UCC_TEST_SKIP_R("Statistic enabled");
+#elif UCC_ENABLE_ASSERT
+    UCC_TEST_SKIP_R("Assert enabled");
+#else
+
+    EXPECTED_SIZE(ucc_global_config_t, 72);
```
Sergei-Lebedev, line 33: Is size of global config still 72?

test/gtest/common/test_obj_size.cc
```diff
@@ -0,0 +1,37 @@
...
+
+UCC_TEST_F(test_obj_size, size) {
+
+#if ENABLE_DEBUG_DATA
+    UCC_TEST_SKIP_R("Debug data");
+#elif defined (ENABLE_STATS)
+    UCC_TEST_SKIP_R("Statistic enabled");
+#elif UCC_ENABLE_ASSERT
+    UCC_TEST_SKIP_R("Assert enabled");
+#else
```
Sergei-Lebedev, line 31: Do we really need all these checks? And will they really affect size of global config?

test/gtest/common/test.h
vspetrov: thanks!

test/gtest/common/test.h
alex--m: In general the Proxy is there to collect errors and warnings instead of giving it to gtest - which would likely fail on an error and stop everything...

test/gtest/configure.m4
alex--m: I traced this to https://github.com/openucx/ucx/commit/6621f7019f318df8eda4020828497578e94849dd - where it was added in UCX. probably to fix some compiler-specific problem. I can remove it - but then we will likely run into the same problem. What's your opinion? @vspetrov

test/gtest/common/test_helpers.h
```diff
@@ -0,0 +1,218 @@
...
+        ucs_status_t _status = (_expr); \
+        EXPECT_EQ(UCC_OK, _status) << "Error: " << ucs_status_string(_status); \
+    } while (0)
+
+
+#define ASSERT_UCC_OK(_expr, ...) \
+    do { \
+        ucs_status_t _status = (_expr); \
+        if ((_status) != UCC_OK) { \
+            UCC_TEST_ABORT("Error: " << ucs_status_string(_status)  __VA_ARGS__); \
```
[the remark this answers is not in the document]
alex--m: correct.

test/gtest/common/test_helpers.h
```diff
@@ -0,0 +1,218 @@
...
+        GTEST_MESSAGE_(ss.str().c_str(), ::testing::TestPartResult::kFatalFailure); \
+        throw ucc::test_abort_exception(); \
+    } while(0)
+
+
+/* UCS error check */
+#define EXPECT_UCC_OK(_expr) \
+    do { \
+        ucs_status_t _status = (_expr); \
+        EXPECT_EQ(UCC_OK, _status) << "Error: " << ucs_status_string(_status); \
```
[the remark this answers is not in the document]
alex--m: correct.

test/gtest/common/test_obj_size.cc
alex--m: size of global config is just a dummy - I simply didn't want to leave an empty function... once this patch is approved - I can start working on real tests.

test/gtest/common/test_obj_size.cc
alex--m: This is also kind of a placeholder - I'll be using it in tests I write in the future (I hope you all will too :) ). The reason NOT to run the object size test is you often want to compare to some CACHELINE_SIZE, but only in release build (because builds with statistics or debug typically add stuff to the struct).

Sergei-Lebedev (approved):

vspetrov:
@alex--m , the PR is ready to merge, plz rebase on top of master

Issue #19, opened 2020-10-15
UCG needs local/global rank/size information


alex--m:
This is a special case of topology information (#13), likely easier to accomplish.

For reference, UCG parameters require the following:
````
    /**
     * Information about other processes running UCX on the same node, used for
     * the UCG - Group operations (e.g. MPI collective operations). This includes
     * both the total number of processes (including myself) and a zero-based
     * index of my process, guaranteed to be unique among the local processes
     * which this process will contact. One such pair refers strictly to the
     * peers on the same host, and the other pair refers to the total amount
     * of peers for communication across the network. Typically the process with
     * index #0 (in either pair) performs special duties in group-aware
     * transports, and those transports need this information on every process.
     *
     * @note Both fields are indicated be the same bit in @ref field_mask.
     */
    struct {
    uint32_t                           num_local;
    uint32_t                           local_idx;
    uint32_t                           num_global;
    uint32_t                           global_idx;
    } peer_info;
````
Full disclose: this is NOT part of the upstream UCP version, but rather a modified UCP I've been using for UCG.

Currently, the OMPI-based implementation satisfies this requirement as follows:
````
    ucp_params.peer_info.num_local  = ompi_process_info.num_local_peers + 1;
    ucp_params.peer_info.local_idx  = ompi_process_info.my_local_rank;
    ucp_params.peer_info.num_global = ompi_process_info.num_procs;
    ucp_params.peer_info.global_idx = ompi_process_info.myprocid.rank;
````
To clarify, the reason this code has `ucp_params` is that this information is passed to UCP (and UCT), but is used exclusively for collective operations and not P2P.

vspetrov:
outdated

PR #22, opened 2020-10-16
UCG component draft

alex--m:
This is a draft describing how initial UCG integration could look like. It can also serve as reference for issues like #19

### DIFF
--- a/config/m4/ucx.m4
+++ b/config/m4/ucx.m4
@@ -8,6 +8,7 @@ UCX_MIN_REQUIRED_MAJOR=1
 UCX_MIN_REQUIRED_MINOR=9
 AS_IF([test "x$ucx_checked" != "xyes"],[
     ucx_happy="no"
+    ucg_happy="no"
 
     AC_ARG_WITH([ucx],
             [AS_HELP_STRING([--with-ucx=(DIR)], [Enable the use of UCX (default is guess).])],
@@ -57,6 +58,20 @@ AS_IF([test "x$ucx_checked" != "xyes"],[
             ucx_happy="no"
         ])
 
+        AC_CHECK_HEADERS([ucg/api/ucg.h],
+        [
+            AC_CHECK_LIB([ucg], [ucg_cleanup],
+            [
+                ucg_happy="yes"
+            ],
+            [
+                ucg_happy="no"
+            ], [-lucg])
+        ],
+        [
+            ucg_happy="no"
+        ])
+
         AS_IF([test "x$ucx_happy" = "xyes"],
         [
             AS_IF([test "x$check_ucx_dir" != "x"],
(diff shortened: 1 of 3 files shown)


config/m4/ucx.m4
```diff
@@ -80,7 +95,17 @@ AS_IF([test "x$ucx_checked" != "xyes"],[
                 AC_SUBST(UCS_LDFLAGS, "-L$check_ucx_libdir")
             ])
 
-            AC_SUBST(UCX_LIBADD, "-lucp -lucs -lucm")
+            AS_IF([test "x$ucg_happy" = "xyes"],
+            [
+                ucg_major=$(cat $check_ucx_dir/include/ucg/api/ucg_version.h | grep -Po "UCG_API_MAJOR\s+\K\d+")
+                ucg_minor=$(cat $check_ucx_dir/include/ucg/api/ucg_version.h | grep -Po "UCG_API_MINOR\s+\K\d+")
+                AC_MSG_RESULT([Detected UCG version: ${ucg_major}.${ucg_minor}])
+                AC_SUBST(UCX_LIBADD, "-lucg -lucp -lucs -lucm")
```
vspetrov, line 103: i don't think we want to add "-lucg" to global flags. It should be UCG_LIBADD parameter used only in CL/ucg. Main libucc.so should not depend on libucg (like it does not depend on say ibverbs, mcast or sharp).

alex--m: @vspetrov It's not added to global flags - only to UCX_LIBADD. This means every component using UCX will also use UCG (which is currently part of UCX, so it makes sense to me). I don't see which UCC component could use UCX without UCG (since UCG is not standalone, so it can't be a separate context anyway).

vspetrov: no. For example, TL/UCX - low level component. It will be linked against UCP only. Why it should depend on UCG. I think you are mixing 2 independent things here. only install/lib/xccl/xccl_cl_ucg.so should depend on libucg.so. So, it makes sense actually to put libucg detection into xucg.m4 . First we run ucx.m4 - if it is detected, then we can run xucg.m4 and try detecting it. If it is also detected, then you populate UCG_CPPFLAGS, UCG_LDFLAGS, UCG_LIBADD and use them in the makefile for CL/ucg.

vspetrov:
@manjugv can we close this? It is outdated and required "re-opening" anyways.

manjugv:
WG Discussion: Close for now and open if needed

lastweek:
Hi all, as an outsider and a newbie to the collective area, I'm a bit confused about the relationship between UCC and UCG.
How is UCC different from UCG? Is UCC an successor to UCG, or is UCC a superset to UCG?

my reference is the official ucc site and alex's 2019 slide: https://ucfconsortium.org/wp-content/uploads/2020/02/Alex_Margolin_Collectives_UCX_Hackathon_2019.pdf 

thanks!

PR #23, opened 2020-10-19
UCC/UTIL: Logger

vspetrov:
**What?**
Adds utils/ucc_log.h: set of macros that define logging utils for the global UCC scope: ucc_error, ucc_info, etc. The global verbose level is controlled by UCC_LOG_LEVEL. Any component that requires its own log level (to be profiled separately when needed) should define similar macros: e.g. cl_basic_error and add separate log_component entry in the config table.

**Why?**
general purpose logging

**How?**
Using ucs logging infra

### DIFF
--- a/src/core/ucc_constructor.c
+++ b/src/core/ucc_constructor.c
@@ -8,6 +8,7 @@
 #include "ucc_global_opts.h"
 #include "utils/ucc_malloc.h"
 #include "utils/ucc_component.h"
+#include "utils/ucc_log.h"
 
 #include <link.h>
 #include <dlfcn.h>
@@ -26,7 +27,8 @@ static int callback(struct dl_phdr_info *info, size_t size, void *data)
         component_path = (char *)ucc_malloc(pos + UCC_COMPONENT_LIBDIR_LEN + 1,
                                             "component_path");
         if (!component_path) {
-            //TODO add ucc_error(msg) when logging as added
+            ucc_error("failed to allocate %d bytes for component_path",
+                      pos + UCC_COMPONENT_LIBDIR_LEN + 1);
             return -1;
         }
         /* copying up to pos+1 due to ucs_strncpy_safe implementation specifics:
@@ -55,7 +57,7 @@ ucc_status_t ucc_constructor(void)
         status = ucc_config_parser_fill_opts(
             &ucc_global_config, ucc_global_config_table, "UCC_", NULL, 1);
         if (UCC_OK != status) {
-            //TODO add ucc_error(msg) when logging is added
+            ucc_error("failed to parse global options");
             return status;
         }
         if (strlen(ucc_global_config.component_path) == 0) {
@@ -64,15 +66,14 @@ ucc_status_t ucc_constructor(void)
     }
 
     if (!ucc_global_config.component_path) {
-        //TODO
-        //ucc_error, "Failed to get ucc library path. set UCC_COMPONENT_PATH.\n");
-        return UCC_ERR_NO_MESSAGE;
+        ucc_error("failed to get ucc components path");
+        return UCC_ERR_NOT_FOUND;
     }
-    if (UCC_OK != ucc_components_load("cl", &ucc_global_config.cl_framework)) {
-        //TODO
-        //ucc_error, "no CL components were found in the UCC_COMPONENT_PATH: %s\n",
-        //ucc_global_config.component_path);
-        return UCC_ERR_NO_MESSAGE;
+    status = ucc_components_load("cl", &ucc_global_config.cl_framework);
+    if (UCC_OK != status) {
+        ucc_error("no CL components were found in the UCC_COMPONENT_PATH: %s",
+                  ucc_global_config.component_path);
+        return status;
     }
     return UCC_OK;
 }
--- a/src/utils/ucc_compiler_def.h
+++ b/src/utils/ucc_compiler_def.h
@@ -10,10 +10,13 @@
 #include "api/ucc_status.h"
 #include <ucs/type/status.h>
 #include <ucs/sys/string.h>
+#include <ucs/debug/log_def.h>
 
 #define ucc_offsetof ucs_offsetof
 #define ucc_container_of ucs_container_of
 #define ucc_strncpy_safe ucs_strncpy_safe
+
+typedef ucs_log_component_config_t ucc_log_component_config_t;
 
 static inline ucc_status_t ucs_status_to_ucc_status(ucs_status_t status)
 {
--- a/src/utils/ucc_component.c
+++ b/src/utils/ucc_component.c
@@ -17,6 +17,7 @@
 #include <glob.h>
 #include <unistd.h>
 #include <string.h>
+#include <assert.h>
 
 static ucc_status_t ucc_component_load_one(const char *so_path,
                                            const char *framework_pattern,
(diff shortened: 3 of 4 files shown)


src/utils/ucc_component.c
```diff
@@ -74,25 +67,25 @@ ucc_status_t ucc_components_load(const char *framework_name,
...
 
-    full_pattern = (char *)ucc_malloc(strlen(ucc_global_config.component_path) +
-                                          strlen(framework_name) + 16,
-                                      "full_pattern");
+    pattern_size =
+        strlen(ucc_global_config.component_path) + strlen(framework_name) + 16;
+    full_pattern = (char *)ucc_malloc(pattern_size, "full_pattern");
     if (!full_pattern) {
-        //TODO ucc_error
+        ucc_error("failed to allocate %d bytes for full_pattern", pattern_size);
```
Sergei-Lebedev, line 88: pattern_size is of type size_t, so use either %zu or PRIu...

src/core/ucc_constructor.c
```diff
@@ -26,7 +27,8 @@ static int callback(struct dl_phdr_info *info, size_t size, void *data)
         component_path = (char *)ucc_malloc(pos + UCC_COMPONENT_LIBDIR_LEN + 1,
                                             "component_path");
         if (!component_path) {
-            //TODO add ucc_error(msg) when logging as added
+            ucc_error("failed to allocate %d bytes for component_path",
+                      pos + UCC_COMPONENT_LIBDIR_LEN + 1);
```
manjugv, line 31: return UCC_ERR_NO_MEMORY; ?

src/utils/ucc_component.c
```diff
@@ -36,24 +36,17 @@ static ucc_status_t ucc_component_load_one(const char *so_path,
...
-                so_path);
+        ucc_debug("Failed to load UCC component library: %s", so_path);
         goto error;
     }
     iface = (ucc_component_iface_t *)dlsym(handle, iface_struct);
     if ((error = dlerror()) != NULL) {
-        //TODO ucc_error
-        fprintf(stderr, "Failed to get iface %s from %s object\n", iface_struct,
-                so_path);
+        ucc_error("Failed to get iface %s from %s object", iface_struct,
```
manjugv, line 44: Let's start with small letters. "Failed" -> "failed"

src/utils/ucc_component.c
```diff
@@ -36,24 +36,17 @@ static ucc_status_t ucc_component_load_one(const char *so_path,
 
     handle = dlopen(so_path, RTLD_LAZY);
     if (!handle) {
-        //TODO use ucc_error
-        fprintf(stderr,
-                "Failed to load UCC component library: %s. "
-                "Check UCC_COMPONENT_PATH or LD_LIBRARY_PATH\n",
-                so_path);
+        ucc_debug("Failed to load UCC component library: %s", so_path);
```
manjugv, line 39: Let's start with small letters. "Failed" -> "failed"

manjugv:
More general comment: We should have some guidelines for writing error messages. Maybe, this is the time to introduce this. Say, for example, we should print component/collective type/algorithm and error message. Don't start with capitals (or be more consistent). For example, we could print "Error: UTILS - failed to ... " for the UTILS component. Error/Warn is appended by UCS in this case.  

Thoughts?

src/utils/ucc_compiler_def.h
```diff
@@ -10,11 +10,15 @@
...
 #include <ucs/type/status.h>
 #include <ucs/sys/string.h>
+#include <ucs/debug/log_def.h>
 
 #define ucc_offsetof ucs_offsetof
 #define ucc_container_of ucs_container_of
 #define ucc_strncpy_safe ucs_strncpy_safe
 
+typedef ucs_log_component_config_t ucc_log_component_config_t;
+#define UCC_LOG_LEVEL_WARN UCS_LOG_LEVEL_WARN
```
alex--m, line 20: should we do this for all UCS log levels? If yes - shouldn't it be part of the logger header?

src/utils/ucc_component.c
```diff
@@ -74,25 +67,25 @@ ucc_status_t ucc_components_load(const char *framework_name,
...
+        ucc_error("unsupported framework_name length: %s, len %d",
+                  framework_name, strlen(framework_name));
         return UCC_ERR_INVALID_PARAM;
     }
     sprintf(framework_pattern, "ucc_%s_", framework_name);
 
-    full_pattern = (char *)ucc_malloc(strlen(ucc_global_config.component_path) +
-                                          strlen(framework_name) + 16,
-                                      "full_pattern");
+    pattern_size =
```
alex--m, line 84: UCX ususally use asprintf for similar tasks - does it make sense we use it too?

src/utils/ucc_component.c
```diff
@@ -108,8 +101,11 @@ ucc_status_t ucc_components_load(const char *framework_name,
         if (n_loaded == ifaces_array_size) {
             ifaces_array_size += 8;
             ifaces = (ucc_component_iface_t **)ucc_realloc(
```
alex--m, line 103: I think it would be better to allocate once, after we know how many we need.

src/core/ucc_constructor.c
```diff
@@ -64,14 +66,12 @@ ucc_status_t ucc_constructor(void)
     }
 
     if (!ucc_global_config.component_path) {
-        //TODO
-        //ucc_error, "Failed to get ucc library path. set UCC_COMPONENT_PATH.\n");
+        ucc_error("failed to get ucc components path");
         return UCC_ERR_NO_MESSAGE;
```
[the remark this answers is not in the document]
vspetrov: yes

src/core/ucc_constructor.c
```diff
@@ -64,14 +66,12 @@ ucc_status_t ucc_constructor(void)
...
     }
     if (UCC_OK != ucc_components_load("cl", &ucc_global_config.cl_framework)) {
-        //TODO
-        //ucc_error, "no CL components were found in the UCC_COMPONENT_PATH: %s\n",
-        //ucc_global_config.component_path);
+        ucc_error("no CL components were found in the UCC_COMPONENT_PATH: %s",
+                  ucc_global_config.component_path);
         return UCC_ERR_NO_MESSAGE;
```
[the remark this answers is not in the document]
vspetrov: done

src/utils/ucc_component.c
vspetrov: changed to %zd since size_t is signed

src/core/ucc_constructor.c
vspetrov: this "callback" function is part of libdl interface - it has to return "int", not ucc_status_t.

src/utils/ucc_component.c
vspetrov: agree, missed this one.

src/utils/ucc_component.c
vspetrov: 👍

src/utils/ucc_compiler_def.h
vspetrov: moved UCC_LOG_LEVEL_WARN to the logger header. We can add other levels when it is needed (currently we only use UCC_LOG_LEVEL_WARN as default value for in global opts).

src/utils/ucc_component.c
vspetrov: asprintf is BSD extension. Looks like UCX uses ucm_asprintf which is internal implementation. Do you mean we should copy-paste that fn into UCC (we can replace the fn to actual ucm when it is exposed as public api) ?

src/utils/ucc_component.c
vspetrov: i will allocate large enough storage (globbuf.gl_pathc) - it will be larger than required. Then i will shrink it to the actual size with ucc_realloc. I can't compute the size in advance, since in order to do that we need to try loading those components first (and store the result somewhere).

src/utils/ucc_component.c
alex--m: I seem to remember they did, probably changed since I last looked. anyway, I don't think it's that important, so probably fine for now.

alex--m (approved):

Sergei-Lebedev (approved):

manjugv (approved):

PR #24, opened 2020-10-19
CORE: ucc_lib_config_read implementation

vspetrov:
**What?**
Adds implementation of the ucc_lib_config_read API. Currently the lib config table only holds the list of CLs to use. To be extended in the future.

**Why?**
API function implementation

**How?**
Using ucs_lib_config_read. The only specialty is the construction of the "full_prefix".
full_prefix for a UCC lib config is either just  (i)  "UCC_" - if not "env_prefix" is provided, or (ii) "UCC_AAA_" - where AAA is the value of "env_prefix". If user specifies env_prefix when performs ucc_lib_config_read he then can control the configuration of his UCC job specifically using set of params prefixed with UCC_<ENV_PREFIX>_ w/o affecting other UCC lib objects (potentially co-existing in the same runtime).

### DIFF
--- a/src/core/ucc_lib.c
+++ b/src/core/ucc_lib.c
@@ -1,6 +1,5 @@
 /**
  * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
- *
  * See file LICENSE for terms.
  */
 
@@ -9,6 +8,16 @@
 #include "ucc_global_opts.h"
 #include "api/ucc.h"
 #include "api/ucc_status.h"
+#include "ucc_lib.h"
+#include "utils/ucc_log.h"
+#include "utils/ucc_malloc.h"
+
+static ucc_config_field_t ucc_lib_config_table[] = {
+    {"CLS", "all", "Comma separated list of CL components to be used",
+     ucc_offsetof(ucc_lib_config_t, cls), UCC_CONFIG_TYPE_STRING_ARRAY},
+
+    {NULL}
+};
 
 ucc_status_t ucc_init_version(unsigned api_major_version,
                               unsigned api_minor_version,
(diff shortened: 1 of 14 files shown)


src/core/ucc_lib.c
```diff
@@ -18,3 +28,70 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+                                 ucc_lib_config_t **config_p)
+{
+    ucc_lib_config_t *config;
+    ucc_status_t      status;
+    int               full_prefix_len;
+    const char       *base_prefix = "UCC_";
+
+    config = ucc_malloc(sizeof(*config), "lib_config");
+    if (config == NULL) {
+        ucc_error("failed to allocated %d bytes for lib_config",
```
Sergei-Lebedev, line 42: * allocated -> allocate?
* sizeof returns size_t, so print type should be different

src/core/ucc_lib.c
```diff
@@ -18,3 +28,70 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+    /* full_prefix for a UCC lib config is either just
+          (i)  "UCC_" - if not "env_prefix" is provided, or
+          (ii) "UCC_AAA_" - where AAA is the value of "env_prefix".
+       Need to allocate enough space to build the proper prefix str: 2 more
+       chars for "_" and "\0". */
+    full_prefix_len =
+        strlen(base_prefix) + (env_prefix ? strlen(env_prefix) : 0) + 2;
+    config->full_prefix = ucc_malloc(full_prefix_len, "full_prefix");
+    if (!config->full_prefix) {
+        ucc_error("failed to allocated %d bytes for full_prefix",
```
Sergei-Lebedev, line 56: allocated->allocate, sizeof

src/core/ucc_lib.c
```diff
@@ -18,3 +28,70 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+        snprintf(config->full_prefix, full_prefix_len, "%s_%s", env_prefix,
+                 base_prefix);
+    } else {
+        strcpy(config->full_prefix, base_prefix);
+    }
+
+    status = ucc_config_parser_fill_opts(config, ucc_lib_config_table,
+                                         config->full_prefix, NULL, 0);
+    if (status != UCC_OK) {
+        ucc_error("faild to read UCC lib config");
```
Sergei-Lebedev, line 71: faild->failed

src/core/ucc_lib.c
```diff
@@ -18,3 +28,70 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+        status = UCC_ERR_NO_MEMORY;
+        goto err;
+    }
+    /* full_prefix for a UCC lib config is either just
+          (i)  "UCC_" - if not "env_prefix" is provided, or
+          (ii) "UCC_AAA_" - where AAA is the value of "env_prefix".
+       Need to allocate enough space to build the proper prefix str: 2 more
+       chars for "_" and "\0". */
+    full_prefix_len =
+        strlen(base_prefix) + (env_prefix ? strlen(env_prefix) : 0) + 2;
```
Sergei-Lebedev, line 53: is it should be +1 instead of +2 at the end if env_prefix is null?

Sergei-Lebedev (commented):
Maybe also add ucc_lib_config_modify?  just to make PR complete

src/core/ucc_lib.c
vspetrov: yes, but do you want process that case separately? i thought 1 byte does not worth extra hassle.

Sergei-Lebedev: sure, let's leave it as is

vspetrov:
> Maybe also add ucc_lib_config_modify? just to make PR complete

@Sergei-Lebedev agree. Done.

src/core/ucc_lib.h
```diff
@@ -0,0 +1,19 @@
+/*
+* Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+* See file LICENSE for terms.
+*/
+
```
manjugv, line 5: Minor : indent

src/core/ucc_lib.c
```diff
@@ -18,3 +28,77 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+        ucc_error("failed to allocate %zd bytes for full_prefix",
+                  full_prefix_len);
+        status = UCC_ERR_NO_MEMORY;
+        goto err_free_config;
+    }
+    if (env_prefix) {
+        snprintf(config->full_prefix, full_prefix_len, "%s_%s", env_prefix,
+                 base_prefix);
+    } else {
+        strcpy(config->full_prefix, base_prefix);
```
manjugv, line 65: Though it is correct here, should we avoid using strcpy and instead use snprintf ? Also, should we use ucs_snprintf_zero eventually ?

src/core/ucc_lib.c
```diff
@@ -18,3 +28,77 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+
+    config = ucc_malloc(sizeof(*config), "lib_config");
+    if (config == NULL) {
+        ucc_error("failed to allocate %zd bytes for lib_config",
+                  sizeof(*config));
+        status = UCC_ERR_NO_MEMORY;
+        goto err;
+    }
+    /* full_prefix for a UCC lib config is either just
+          (i)  "UCC_" - if not "env_prefix" is provided, or
```
manjugv, line 48: Minor: If not -> if no

src/core/ucc_lib.c
```diff
@@ -18,3 +28,77 @@ ucc_status_t ucc_init_version(unsigned api_major_version,
...
+          (ii) "AAA_UCC" - where AAA is the value of "env_prefix".
+       Need to allocate enough space to build the proper prefix str: 2 more
```
manjugv, line 50: Minor: “Need to allocate enough space to build the proper prefix .. “ -> “Allocate space to build prefix str and two characters ("_" and “\0”)”  ?

manjugv:
All minor comments. Looks good!

Sergei-Lebedev:
there are still left some compiler warnings, check ucc build step in latest CI run

src/core/ucc_lib.h
vspetrov: done

vspetrov:
> there are still left some compiler warnings, check ucc build step in latest CI run

yes. i've fixed. And i added "-Wall -Werror" to the build.

src/core/ucc_lib.c
vspetrov: @manjugv  @Sergei-Lebedev  changed strcpy to ucc_strncpy_safe. Regarding snprintf: i've added ucc_snprintf_safe. However, the implementation in UCS (ucs_snprintf_safe) turned to have a bug - it truncates last char (i've reported that to UCX people). Hence, so far i've defined ucc_snprintf_safe to just snprintf from stdlib (which does guarantee '\0' appended btw).

Sergei-Lebedev (approved):

alex--m (approved):

PR #25, opened 2020-10-23
CORE: free parser in lib destructor

Sergei-Lebedev:


src/core/ucc_constructor.c
```diff
@@ -79,7 +79,8 @@ ucc_status_t ucc_constructor(void)
 
 __attribute__((destructor)) static void ucc_destructor(void)
 {
-    if (ucc_global_config.component_path_default) {
-        ucc_free(ucc_global_config.component_path_default);
```
vspetrov, line 83: why this ucc_free is removed?

Sergei-Lebedev: ucs config parser release iterates over all fields declared in config table and calls corresponding destructor. String field destructor is here https://github.com/openucx/ucx/blob/1c883359c954e6678e0a2dbfea4efbbedb029d60/src/ucs/config/parser.c#L105
It calls free for pointer no matter if it was parsed (allocated) or not.

vspetrov: but "component_path_default" is NOT declared in the config table. It is detected automatically. "component_path" is declared. am i missing smth?

Sergei-Lebedev: yes, component path default is not part of config table, but if component path is not provided by user then component path pointer assigned to component path default
https://github.com/openucx/ucc/blob/dfdfb594da7bd61af7cb2b74856b9e077c3b1bfd/src/core/ucc_constructor.c#L37

vspetrov: got it

vspetrov (approved):

manjugv (approved):

PR #27, opened 2020-10-30
TEST: run clang-format in CI

Sergei-Lebedev:
## What
Adds clang-format check for each pull request

## Why ?
Check code style automatically

## How ?
Run clang-format only for changes introduced in pull request. If clang-formats finds any format issues it prints patch that can be used to fix it

### DIFF
--- a/.github/workflows/codestyle.yaml
+++ b/.github/workflows/codestyle.yaml
@@ -2,11 +2,20 @@ name: Codestyle
 
 on: [pull_request]
 
+env:
+  GIT_CF: https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format
 jobs:
-  check-commit-title:
+  check-codestyle:
     runs-on: ubuntu-latest
-    name: Check commit title
+    name: Check code style
     steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends wget lsb-core software-properties-common
+        sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
+        sudo wget -O /usr/bin/git-clang-format $GIT_CF
+        sudo chmod +x /usr/bin/git-clang-format
     - name: Checking out repository
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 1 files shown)


alex--m (approved):

vspetrov:
@Sergei-Lebedev i'm not sure but this imho must be problematic. Each time i prepared a PR, i applied clang-format on it, but then i had to revert a bunch of "incorrect/bad" clang fixes. So, it is almost never "zero" diff from clang-format in the PR. So, the question: is this GH check just warns that there can be some change to formatting OR is it disallows the PR if the diff with clang is not empty?

Sergei-Lebedev:
> @Sergei-Lebedev i'm not sure but this imho must be problematic. Each time i prepared a PR, i applied clang-format on it, but then i had to revert a bunch of "incorrect/bad" clang fixes. So, it is almost never "zero" diff from clang-format in the PR. So, the question: is this GH check just warns that there can be some change to formatting OR is it disallows the PR if the diff with clang is not empty?
I understand it but at the same time it doesn't look good when we need to reformat code after code formatter, sounds like a double work. This check shouldn't block merge button though and will just show that test is not passed.
We can consider following options going forward:
1.  follow clang format in each and every case maybe excluding some files (e.g. api/ucc.h)
2.  in addition to 1 start using clang-format on/off for code snippets that we don't want to touch
3.  print clang format output as a hint but doesn't force PR author to follow it
WDYT?

vspetrov:
@Sergei-Lebedev @manjugv i think
1. is not really an option. There are always some weird formatting cases and excluding the whole file to fix means, that we will end up with virtually ALL files excluded.

2. We can do that. But we will have non pretty artifacts in the code. Most of the function definitions will have those (currently clang strangely aligns pointer arguments in functions).

3. Is essentially what i do now - ie, run clang + manual fixes. But if we see the hint it is still usefull for the corner cases. For example, author of the PR might manually fix smth that reviewers might consider incorrect. They can see that from  the CI. However, this means that reviewers/author should NOT forget looking into that output before approving the PR.


I think we can't entirely automate that w/o manual "double" work unless we agree to have some badly formatted pieces of code.

Sergei-Lebedev:
@vspetrov @manjugv I've changed clang-format test behavior, now if it fails it will not mark test red but still suggests patch to fix formatting, so it's now up to author/reviewers to not forget check test output

vspetrov (approved):

manjugv (approved):

PR #28, opened 2020-10-30
API: dtype/op/reduce changes

vspetrov:
## What
Fixes the support of custom user reduction in the UCC/API

## Why ?
UCC v1.0 is supposed to support custom user reduction (over pre-defined ucc datatype)

## How ?
- The reduction wrapper callback signature is fixed: now it takes ucc_datatype_t and "void* op" as well. op_t can store user defined context that defines the custom op. E.g. in the case of integration into ompi it will store ompi_op_t *.
- The custom reduction op is stored on the coll_args as void*

```
typedef void(*ucc_reduction_wrapper_t)(void *invec, void *inoutvec,
                                                                 ucc_count_t *count, ucc_datatype_t dtype,  void *op) {
	ompi_op_t *op = (ompi_op_t*)(op);
	ompi_datatype_t dt = ucc_dt_to_ompi_dt(dtype); // for pre-defined types - fixed translation table
	ompi_dt_reduce(invec, inoutvec, count, dt, op); //actual invocation for a callback
}

ompi/coll/ucc/mca_ucc_ops.c:
int mca_coll_ucc_allreduce(const void *sbuf, void *rbuf, int count, struct ompi_datatype_t *dtype,
                            struct ompi_op_t *op, struct ompi_communicator_t *comm,
                            mca_coll_base_module_t *module)
{
    mca_coll_ucc_module_t *ucc_module = (mca_coll_ucc_module_t*)module;
    ucc_coll_req_h          req;
    ucc_datatype_t         ucc_dt;
    ucc_reduction_op_t  ucc_op;

    ucc_dt = ompi_dtype_to_ucc_dtype(dtype); //const lookup table for pre-defined types
    if (OMP_OP_PREDEFINED(op)) {
         ucc_op = ompi_op_to_ucc_op(op); // const lookup table for pre-defined ops
    } else {
         ucc_op = ucc_op_make_userdefined(op);
    }
    ucc_op = ucc_op_make_user
    ucc_coll_op_args_t coll = {
        .coll_type = UCC_ALLREDUCE,
        .buffer_info = {
            src_buffer = (void*)sbuf,
            dst_buffer = (void*)rbuf,
            src_counts = (ucc_count_t*)(uintptr_t)count,
            dst_counts = (ucc_count_t*)(uintptr_t)count,
            src_datatype   = ucc_dt,
            dst_datatype   = ucc_dt,
        },
        .predefined_reduction_op = UCC_OP_USERDEFINED,
        .custom_reduction_op  = (void*)op,
    };

    COLL_UCC_CHECK(ucc_collective_init(&coll, &req, ucc_module->ucc_team));
    COLL_UCC_CHECK(ucc_collective_post(req));
    COLL_UCC_CHECK(coll_ucc_req_wait(req));
    return OMPI_SUCCESS;
}
```
inside UCC library:
reduce algorithm can checks if ucc_op is not predefined and calls:
lib->reduction_wrapper(buf1, buf2, count, ucc_dt,  custom_reduction_op), where ucc_dt and op from the coll_args.buffer_info.reduce (above)
the wrapper calls the fn above where user extracts ompi_op_t back from the ucc_op.

### DIFF
--- a/src/api/ucc.h
+++ b/src/api/ucc.h
@@ -263,27 +263,28 @@ typedef enum {
 /**
  *  @ingroup UCC_COLLECTIVES
  *
- *  @brief The reduction wrapper provides a method to map custom user types to higher level
- *  programming model datatypes
+ *  @brief The reduction wrapper provides an interface for the UCC library to invoke
+ *         user-defined custom reduction callback
  *
  *  @param [in]  invec          The input elements to be reduced by the user function
  *  @param [in]  inoutvec       The input elements to be reduced and output of the reduction
  *  @param [in]  count          The number of elements of type "dtype" to be reduced
- *  @param [in]  dtype          Datatype passed to the reduction operation
+ *  @param [in]  dtype          Datatype specified in the coll_args
+ *  @param [in]  custom_op      A pointer to the user defined reduction passed to the coll_args as custom_reduction_op
  *
  *
  *  @parblock
  *
  *  @b Description
  *
- *  This function is called by the UCC library before calling the user-defined reduction.
- *  Hence, the signature of this function is same as ucc_userdefined_reductions_op_t.
- *  It maps the custom user types to higher level programming model datatypes (such as MPI datatypes)
+ *  This function is called by the UCC library when it needs to perform a non-standard
+ *  user-defined reduction operaion during allreduce/reduce collective.
  *
  *  @endparblock
  */
-typedef void(*ucc_reduction_dtype_mapper_t)(void *invec, void *inoutvec,
-                                            ucc_count_t *count, ucc_datatype_t dtype);
+typedef void(*ucc_reduction_wrapper_t)(void *invec, void *inoutvec,
+                                       ucc_count_t *count, void *dtype,
+                                       void *custom_reduction_op);
 
 /**
  * @brief UCC library initialization parameters
@@ -333,12 +334,12 @@ enum ucc_lib_attr_field{
  *
  */
 typedef struct ucc_lib_params {
-    uint64_t                     mask;
-    ucc_thread_mode_t            thread_mode;
-    uint64_t                     coll_types;
-    uint64_t                     reduction_types;
-    ucc_coll_sync_type_t         sync_type;
-    ucc_reduction_dtype_mapper_t reduction_mapper;
+    uint64_t                mask;
+    ucc_thread_mode_t       thread_mode;
+    uint64_t                coll_types;
+    uint64_t                reduction_types;
+    ucc_coll_sync_type_t    sync_type;
+    ucc_reduction_wrapper_t reduction_wrapper;
 } ucc_lib_params_t;
 
 /**
@@ -1313,29 +1314,6 @@ typedef enum {
 } ucc_error_type_t;
 
 /**
- *  @ingroup UCC_COLLECTIVES
- *
- *  @brief The user-defined reduction function signature.
- *
- *  @param [in]  invec          The input elements to be reduced by the user function
- *  @param [in]  inoutvec       The input elements to be reduced and output of the reduction
- *  @param [in]  count          The number of elements of type "dtype" to be reduced
- *  @param [in]  dtype          Datatype passed to the reduction operation
- *
- *  @parblock
- *
- *  @b Description
- *
- *  @ref ucc_userdefined_reduction_op_t is a reduction operation signature for
- *  user-defined reductions. The signature closely follows the MPI signature.
- *
- *  @endparblock
- *
- */
-typedef void(*ucc_userdefined_reduction_op_t)(void *invec, void *inoutvec,
-                                              ucc_count_t *count,
-                                              ucc_datatype_t dtype);
-/**
  *  @ingroup UCC_COLLECTIVES_DT
  */
 enum ucc_coll_op_args_field {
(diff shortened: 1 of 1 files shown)


manjugv:
@vspetrov  Thanks for the example and PR. These are a tricky set of changes.

* Wrapper : Changes to the wrapper make sense. 
* I'm in favor of simplicity at the expense of extra field for user reduction and user datatype.
    * So, IMO, making ucc_reduction_op_t into an integer to save an extra field might not be worth it. From the readability perspective, it seems convoluted for a user API. 
   * Same for the datatype, we could hold an extra pointer for packed type/user type (as you had suggested) in the coll args or buffer info. 
* Buffer info: I am not opposed to this. However, we had separated buffer information fields initially, and WG preferred to have a unified buffer structure.  

Let's discuss in today's meeting, if we have time.

manjugv:
UCC WG (Nov 4): 
 - Keep the wrapper changes as proposed by Val
 - Keep datatypes and user defined function fields separate 
 - Buffer info changes - We are debating about usability and elegance. No final decision, yet.

vspetrov:
reverted most of changes keeping only wrapper_cb modifications.

vspetrov:
i still think that having different definitions for single/vector/reduce buffer info is better. see in the example above how i have to cast count to ucc_count_t* - imho very clumsy. additionally, we have a field "predefined_reductino_op" that we init to UCC_OP_USERDEFINED. Maybe we need to re-name it to just "reduction_op"? Otherwise it is weird, "predefined_op" = USERDEFINED.

alex--m (approved):
Approve because it addresses the reduction issue. some valuable comments like casting "count" (I agree with @vspetrov it doesn't look very good) we should probably address in a separate PR.

src/api/ucc.h
```diff
@@ -261,27 +261,28 @@ typedef enum {
...
+ *  This function is called by the UCC library when it needs to perform a non-standard
+ *  user-defined reduction operaion during allreduce/reduce collective.
  *
  *  @endparblock
  */
-typedef void(*ucc_reduction_dtype_mapper_t)(void *invec, void *inoutvec,
-                                            ucc_count_t *count, ucc_datatype_t dtype);
+typedef void(*ucc_reduction_wrapper_t)(void *invec, void *inoutvec,
+                                       ucc_count_t *count, ucc_datatype_t dtype,
+                                       void *custom_reduction_op);
```
manjugv, line 285: @vspetrov You want datatype to be void* instead of ucc_datatype_t for two reasons 1) OMPI is operating under ompi datatype and 2) future proof for supporting user-defined datatype (opaque). For (2) void* datatype is not sufficient but necessary.

vspetrov: @manjugv  What will i pass to void* argument when i call the ucc_reduction_wrapper callback inside UCC? User provides us with ucc_datatype_t. Will i just cast like this:
```
ucc_knomial_allreduce (...) {
    ... some code ...
    ucc_reduction_wrapper (sbuf, rbuf, &count, (void*)&ucc_datatype, custom_op);
}
```
And how in this case OMPI will treat this void* to implement the callback? We decided that user does NOT pass "ompi_datatype_t" into UCC right now. We don't a place for that. So, i HAVE to use ucc_datatype_t. I suggested the solution where all the information is "packed" into ucc_datatype_t (both pointer and ucc_datatype) you said you didn't like it. "void*" - does not work.

manjugv: @vspetrov Once you call the wrapper, you don't care about ucc datatype, right ? You only care about how to get to ompi datatype type. I modified your code snippet to show this. Let me know, if it looks right? Let’s discuss this in the WG tomorrow.

manjugv:
```C
/*
 * When user-defined op and user-defined datatype 
 * ompi/coll/ucc/mca_ucc_ops.c:
 */

typedef void(*ucc_reduction_wrapper_t)(void *invec, void *inoutvec,
                                                                 ucc_count_t *count, void *dtype,  void *op) {
	ompi_op_t *op = (ompi_op_t*)(op);
	ompi_datatype_t dt = (ompi_dt_t*)(dtype); // for pre-defined types - fixed translation table
	ompi_dt_reduce(invec, inoutvec, count, dt, op); //actual invocation for a callback
}


int mca_coll_ucc_allreduce(const void *sbuf, void *rbuf, int count, struct ompi_datatype_t *dtype,
                            struct ompi_op_t *op, struct ompi_communicator_t *comm,
                            mca_coll_base_module_t *module)
{
    mca_coll_ucc_module_t *ucc_module = (mca_coll_ucc_module_t*)module;
    ucc_coll_req_h         req;
    ucc_datatype_t         ucc_dt;
    ucc_reduction_op_t     ucc_op;

    ucc_dt = ompi_dtype_to_ucc_dtype(dtype); //const lookup table for pre-defined types
    ucc_op = ompi_op_to_ucc_op(op); // const lookup table for pre-defined ops
   
    ucc_op = ucc_op_make_user
    ucc_coll_op_args_t coll = {
        .coll_type = UCC_ALLREDUCE,
        .buffer_info = {
            reduce.src_buffer = (void*)sbuf,
            reduce.dst_buffer = rbuf,
            reduce.count      = count,
            reduce.datatype   = ucc_dt,
            reduce.op         = ucc_op,
        },
        if (UCC_OP_USERDEFINED)||(UCC_DT_OPAQUE == ucc_dt) 
        {
            .buffer_info.reduce.custom_reduction_op = (void*)op;
            .user_dtype = (void*)dtype; /* This is a new field */
        }
    };

    COLL_UCC_CHECK(ucc_collective_init(&coll, &req, ucc_module->ucc_team));
    COLL_UCC_CHECK(ucc_collective_post(req));
    COLL_UCC_CHECK(coll_ucc_req_wait(req));
    return OMPI_SUCCESS;
}

In the UCC library .. 

ucc_allreduce()
{

    if (coll_args.buffer_info->reduce.op == UCC_OP_USERDEFINED) {
        reduction_wrapper(inbuf, outbuf, count, user_dtype, custom_reduction_op);
    }
}
```

vspetrov:
@manjugv added void* and custom_dtype. plz re-review

manjugv (approved):

manjugv:
@Sergei-Lebedev Do you have any comments? If not, can you approve. Thanks!

Sergei-Lebedev (approved):

PR #30, opened 2020-11-02
API: protect headers for c++ usage

vspetrov:
## What
Adds extern "C" protection when c++ i used

## Why ?
Otherwise c++ app that uses libucc wouldn't link

### DIFF
--- a/src/api/ucc.h
+++ b/src/api/ucc.h
@@ -14,6 +14,8 @@
 #include <api/ucc_status.h>
 #include <stdio.h>
 
+BEGIN_C_DECLS
+
 /** Unified Collective Communications (UCC) Library Specification
  *
  *  UCC is a collective communication operations API and library that is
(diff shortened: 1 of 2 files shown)


Sergei-Lebedev (approved):

manjugv (approved):

alex--m:
@vspetrov why do we need this? UCC is a pure C library. If a C++ framework wants to use it, it should do:
````
extern "C" {
#include <ucc.h>
}
````

Might be a nitpick, but `extern "C"` is a C++ keyword (for using C libraries), so I think it should only be used in C++ code (e.g. gtest), not in UCC's (C) header.

alex--m:
Just for laughs - do we add `extern "FORTRAN"`* as well? :)
*[it's a real thing](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.4.0/com.ibm.zos.v2r4.cbclx01/func_extsc.htm)!

manjugv:
UCC WG (Nov 4):
- Val to discuss with Alex and action to be decided.

vspetrov:
@alex--m yes, on the one hand user can always add "extern "C"" themselves but the common practice is that c++ friendly c libraries would wrap their headers with the macro above. E.g., UCP does that. I think we should follow the same approach. THe app that uses ucp/ucc would just include <api/ucc.p> and <api/ucc.h> w/o extra macros.

alex--m (approved):
ok, I'm convinced :)

PR #35, opened 2020-12-01
Shared ucp_context/worker for ucc_context_t

vspetrov:
## What
Adds a minimal interface and implementation of the dynamic component that holds UCP context/worker that can be shared among other internal UCC components (CLs, TLs)

## Why ?
Per multiple WG discussions we want to share UCP ctx/worker between multiple Cls,TLs

## How ?

- Firstly, we don't want to have a direct link dependency of libucc.so on libucp.so. So, all the code that initialized UCP worker/context is put into a stand alone .so dynamically loaded component called "libucc_ucp_ctx.so". It is loaded via ucc_component interfaces during ucc_constructor.
- The component has to interface to create and destroy the ucc_ucp_ctx_handle_t
- This handle is allocate during ucc_context creation a stored on the ucc_context
- Any other component can later grab an instance of ucc_ucp_ctx_t (which is a pair of ucp_context_h and ucp_worker_h) via ucc_ucp_ctx_get/put calls. Each component that want to use shared ucp context must register itself into the enum of ucc_ucp_ctx_type_t and provide its enum value during ucc_ucp_ctx_get() call. This will make sure that the ref_counting is done properly and the tag space is separated between components (there are several bits reserved, see ucc_ucp_tag.h)

### DIFF
--- a/src/ucp_ctx/ucc_ucp_ctx.c
+++ b/src/ucp_ctx/ucc_ucp_ctx.c
@@ -1,0 +1,171 @@
+#include "ucc_ucp_ctx.h"
+#include "utils/ucc_malloc.h"
+#include "utils/ucc_log.h"
+#include <ucp/api/ucp.h>
+
+#include <sys/types.h>
+#include <unistd.h>
+
+typedef struct ucc_ucp_ctx_handle {
+    ucc_ucp_ctx_t      ctx;
+    int                ref_count;
+} ucc_ucp_ctx_handle_t;
+
+static void ucc_ucp_req_init(void *request)
+{
+    ucc_ucp_req_t *req = (ucc_ucp_req_t *)request;
+    req->status        = UCC_OPERATION_INITIALIZED;
+}
+
+static void ucc_ucp_req_cleanup(void *request)
+{
+}
+
+ucc_status_t ucc_ucp_ctx_create(const ucc_ucp_ctx_create_params_t *params,
+                                ucc_ucp_ctx_handle_t **ctx)
+{
+    ucc_status_t          ucc_status = UCC_OK;
+    ucc_ucp_ctx_handle_t *ctx_h;
+    ucp_worker_params_t   worker_params;
+    ucp_worker_attr_t     worker_attr;
+    ucp_params_t          ucp_params;
+    ucp_config_t         *ucp_config;
+    ucp_context_h         ucp_context;
+    ucp_worker_h          ucp_worker;
+    ucs_status_t          status;
+
+    UCC_STATIC_ASSERT(UCC_UCP_CTX_LAST - 1 < UCC_BIT(UCC_UCP_CTX_BITS));
+
+    ctx_h = ucc_malloc(sizeof(*ctx_h), "ucc_ucp_ctx");
+    if (!ctx_h) {
+        ucc_error("failed to allocate %zd bytes for ucc_ucp_ctx",
+                  sizeof(*ctx_h));
+        return UCC_ERR_NO_MESSAGE;
+    }
+    status = ucp_config_read(params->prefix, NULL, &ucp_config);
+    if (UCS_OK != status) {
+        ucc_error("failed to read ucp configuration, %s",
+                  ucs_status_string(status));
+        ucc_status = ucs_status_to_ucc_status(status);
+        goto err_cfg_read;
+    }
+
+    ucp_params.field_mask =
+        UCP_PARAM_FIELD_FEATURES | UCP_PARAM_FIELD_REQUEST_SIZE |
+        UCP_PARAM_FIELD_REQUEST_INIT | UCP_PARAM_FIELD_REQUEST_CLEANUP |
+        UCP_PARAM_FIELD_TAG_SENDER_MASK;
+    ucp_params.features          = UCP_FEATURE_TAG | UCP_FEATURE_RMA;
+    ucp_params.request_size      = sizeof(ucc_ucp_req_t);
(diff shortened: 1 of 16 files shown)


vspetrov:
@manjugv @Sergei-Lebedev @alex--m This PR is a start point for a discussion how to add Shared UCP worker/context into UCC. Note, it is NOT the "external" UCP ctx/worker (that we grab from the outside). It is created internally per ucc_context_create call (if needed). Plz see the description above and comment.

src/ucp_ctx/ucc_ucp_ctx.c
```diff
@@ -0,0 +1,183 @@
...
+{
+    ucc_ucp_req_t *req = (ucc_ucp_req_t *)request;
+    req->status        = UCC_OPERATION_INITIALIZED;
+}
+
+static void ucc_ucp_req_cleanup(void *request)
+{
+}
+
+ucc_status_t ucc_ucp_ctx_create(ucc_ucp_ctx_create_params_t *params,
```
Sergei-Lebedev, line 25: ucc_ucp_ctx_create_params_t *params -> const ucc_ucp_ctx_create_params_t *params

src/ucp_ctx/ucc_ucp_ctx.c
```diff
@@ -0,0 +1,183 @@
...
+
+    return UCC_OK;
+}
+
+ucc_status_t ucc_ucp_ctx_get(ucc_ucp_ctx_handle_t *ctx_handle,
+                             ucc_ucp_ctx_t **ctx, ucc_ucp_ctx_type_t ctx_type)
+{
+    /* protect tag space: don't allow the same ctx_type to be used
+       on the same ucp context */
+    if (ctx_handle->types_used[ctx_type] >= 0) {
```
Sergei-Lebedev, line 162: how it can become >=0?

src/ucp_ctx/ucc_ucp_tag.h
```diff
@@ -0,0 +1,45 @@
...
+#define UCC_UCP_CTX_BITS_OFFSET                                                \
+    (UCC_UCP_ID_BITS + UCC_UCP_SENDER_BITS + UCC_UCP_TAG_BITS)
+
+#define UCC_UCP_TAG_BITS_OFFSET (UCC_UCP_ID_BITS + UCC_UCP_SENDER_BITS)
+#define UCC_UCP_SENDER_BITS_OFFSET (UCC_UCP_ID_BITS)
+#define UCC_UCP_ID_BITS_OFFSET 0
+
+#define UCC_UCP_MAX_TAG ((((uint64_t)1) << UCC_UCP_TAG_BITS) - 1)
+#define UCC_UCP_MAX_SENDER ((((uint64_t)1) << UCC_UCP_SENDER_BITS) - 1)
+#define UCC_UCP_MAX_ID ((((uint64_t)1) << UCC_UCP_ID_BITS) - 1)
```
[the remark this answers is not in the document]
vspetrov: 👍

src/ucp_ctx/ucc_ucp_ctx.c
vspetrov: if some component grabs the instance of ucc_ucp_ctx_t with a given ucc_ucp_ctx_type_t and then another component will attempt to grab with the same ucc_ucp_ctx_type_t. It could potentially lead to tag collision. It can only happen if there is a devel error. I agree maybe it is too much (i was in doubt when added this). Ref count is enough. I'll remove this.

src/ucp_ctx/ucc_ucp_tag.h
```diff
@@ -0,0 +1,45 @@
...
+
+#define UCC_UCP_TAG_SENDER_MASK                                                \
+    ((((uint64_t)1) << (UCC_UCP_ID_BITS + UCC_UCP_SENDER_BITS)) - 1)
```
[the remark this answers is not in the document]
vspetrov: done

manjugv:
@vspetrov At a high-level, does it make sense to generalize this infrastructure to share all types contexts - UCP worker, SHARP context ?

manjugv:
Some thoughts: 

- Wondering from the code organization perspective, where should we put the shared resources code. Given that this is not a one-off resource sharing, we might need it live in its directory. In the future, we might share algorithms, net patterns, and topology abstraction.

- You have relation from ucc_context -> ucc_ucp_ctx_handle_t -> ucc_ucp_ctx_t (which is inherited from ucc_context from functionality perspective). Though structure wise it is not circular, from the functionality perspective (ucc_ucp_ctx_create_params), it seems circular. 

- Is there an advantage to have UCP workers create/destroy here vs in the transport layer. I’m wondering if initializing and creating the context in TL makes sense. So, as we expand the context to other contexts like SHARP, SM, and NCCL, the contexts are created there in the TL and added to the shared data structure. Any TL or CL, requiring to share the context can fetch it from the shared data structure.

vspetrov:
@manjugv i agree with you. and after 2nd thought i realized we don't need any special framework for sharing a ctx. Contexts creation and abstraction is already there via CL (and TL which is going to be just a minor extension of CL) iface. So, ucc_context will allocate CL/TL framework which will contain TLs: ucp, sharp, mcast and possible others. We will just need to add flags that TL ctx is going to be shared and the callback for upper layers to grab it. The only disadvantage (lets discuss) is that lower level (TL UCP) should be kind of away of how many other users of its ucp_worker will be there  - even potentially at upper layers.

vspetrov:
decided to change the approach. closing.

PR #37, opened 2020-12-02
Ucc ctx create destroy

vspetrov:
## What
Adds ucc_context_create/destroy implementation
ucc_cl_context_t class
Minimal implementation for CL_BASIC

## Why ?
Part of API

## How ?

- Each ucc_context_t is a composition of cl_contexts. The logic is simple: just loop through all the CLs available on the lib_info object after filtering and create the corresponding cl contexts.
- If some CL requires shared UCP context then it is created as part of ucc_context_create. Then the shared UCP can be accessed through the ucc_ucp_ctx_get/put (see #35 )

### DIFF
--- a/src/cl/ucc_cl.h
+++ b/src/cl/ucc_cl.h
@@ -28,12 +28,15 @@ typedef struct ucc_cl_lib_config {
 } ucc_cl_lib_config_t;
 
 typedef struct ucc_cl_context_config {
-    ucc_cl_iface_t *iface;
     ucc_cl_lib_t   *cl_lib;
 } ucc_cl_context_config_t;
 
 extern ucc_config_field_t ucc_cl_lib_config_table[];
 extern ucc_config_field_t ucc_cl_context_config_table[];
+
+typedef struct ucc_cl_context_params {
+    const ucc_context_params_t *params;
+} ucc_cl_context_params_t;
 
 typedef struct ucc_cl_iface {
     ucc_component_iface_t          super;
@@ -42,11 +45,16 @@ typedef struct ucc_cl_iface {
     ucc_lib_attr_t                 attr;
     ucc_config_global_list_entry_t cl_lib_config;
     ucs_config_global_list_entry_t cl_context_config;
-    ucc_status_t                   (*init)(const ucc_lib_params_t *params,
-                                           const ucc_lib_config_t *config,
-                                           const ucc_cl_lib_config_t *cl_config,
-                                           ucc_cl_lib_t **cl_lib);
-    ucc_status_t                   (*finalize)(ucc_cl_lib_t *cl_lib);
+    ucc_status_t (*init)(const ucc_lib_params_t *params,
+                         const ucc_lib_config_t *config,
+                         const ucc_cl_lib_config_t *cl_config,
+                         ucc_cl_lib_t **cl_lib);
+    ucc_status_t (*finalize)(ucc_cl_lib_t *cl_lib);
+    ucc_status_t (*context_create)(ucc_cl_lib_t *cl_lib,
+                                   const ucc_cl_context_params_t *params,
+                                   const ucc_cl_context_config_t *config,
+                                   ucc_cl_context_t **cl_ctx);
+    ucc_status_t (*context_destroy)(ucc_cl_context_t *cl_ctx);
 } ucc_cl_iface_t;
 UCC_CLASS_DECLARE(ucc_cl_iface_t);
 
@@ -63,6 +71,8 @@ typedef struct ucc_cl_context {
     ucc_cl_lib_t *cl_lib;
 } ucc_cl_context_t;
 
+UCC_CLASS_DECLARE(ucc_cl_context_t, ucc_cl_lib_t *);
+
 #define UCC_CL_IFACE_NAME_PREFIX(_NAME)                 \
     .name   = UCC_PP_MAKE_STRING(CL_ ## _NAME),         \
     .prefix = UCC_PP_MAKE_STRING(CL_ ## _NAME ##_)
@@ -73,15 +83,17 @@ typedef struct ucc_cl_context {
         .table  = ucc_cl_ ## _name ## _ ## _cfg ## _config_table,       \
         .size   = sizeof(ucc_cl_ ## _name ## _ ## _cfg ## _config_t)}
 
-#define UCC_CL_IFACE_DECLARE(_name, _NAME, _priority)           \
-    ucc_cl_ ## _name ## _iface_t ucc_cl_ ## _name = {           \
-        UCC_CL_IFACE_CFG(lib, _name, _NAME),                    \
-        UCC_CL_IFACE_CFG(context, _name, _NAME),                \
-        .super.super.name = UCC_PP_MAKE_STRING(basic),          \
-        .super.type       = UCC_CL_ ## _NAME,                   \
-        .super.priority   = _priority,                          \
-        .super.init       = ucc_cl_ ## _name ## _lib_init,      \
-        .super.finalize   = ucc_cl_ ## _name ## _lib_finalize,  \
+#define UCC_CL_IFACE_DECLARE(_name, _NAME, _priority)                          \
+    ucc_cl_##_name##_iface_t ucc_cl_##_name = {                                \
+        UCC_CL_IFACE_CFG(lib, _name, _NAME),                                   \
+        UCC_CL_IFACE_CFG(context, _name, _NAME),                               \
+        .super.super.name      = UCC_PP_MAKE_STRING(basic),                    \
+        .super.type            = UCC_CL_##_NAME,                               \
+        .super.priority        = _priority,                                    \
+        .super.init            = ucc_cl_##_name##_lib_init,                    \
+        .super.finalize        = ucc_cl_##_name##_lib_finalize,                \
+        .super.context_create  = ucc_cl_##_name##_context_create,              \
+        .super.context_destroy = ucc_cl_##_name##_context_destroy,             \
     }
 
 #endif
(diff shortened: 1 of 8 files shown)


src/core/ucc_context.c
```diff
@@ -204,3 +205,84 @@ void ucc_context_config_print(const ucc_context_config_h config, FILE *stream,
...
+    cl_ctx_params.params = params;
+
+    for (i = 0; i < num_cls; i++) {
+        cl_lib = config->configs[i]->cl_lib;
+        status = cl_lib->iface->context_create(cl_lib, &cl_ctx_params,
+                                               config->configs[i], &cl_ctx);
+        if (UCC_OK != status) {
+            ucc_error("failed to create cl context for %s",
+                      cl_lib->iface->super.name);
+            goto error_ctx_create;
```
Sergei-Lebedev, line 251: Is it a fatal error and we cannot proceed? I would rather consider to print warn and continue and if no contexts were opened after the loop then print error

src/core/ucc_context.c
```diff
@@ -204,3 +205,84 @@ void ucc_context_config_print(const ucc_context_config_h config, FILE *stream,
...
+        }
+        ctx->cl_ctx[i] = cl_ctx;
+    }
+    ucc_debug("created ucc context %p", ctx);
```
Sergei-Lebedev, line 255: pls add lib name to debug print

Sergei-Lebedev (commented):
clang format output might be relevant
```diff
--- a/src/cl/ucc_cl.h
+++ b/src/cl/ucc_cl.h
@@ -45,15 +45,15 @@ typedef struct ucc_cl_iface {
     ucc_lib_attr_t                 attr;
     ucc_config_global_list_entry_t cl_lib_config;
     ucs_config_global_list_entry_t cl_context_config;
-    ucc_status_t (*init)(const ucc_lib_params_t *params,
-                         const ucc_lib_config_t *config,
+    ucc_status_t (*init)(const ucc_lib_params_t *   params,
+                         const ucc_lib_config_t *   config,
                          const ucc_cl_lib_config_t *cl_config,
-                         ucc_cl_lib_t **cl_lib);
+                         ucc_cl_lib_t **            cl_lib);
     ucc_status_t (*finalize)(ucc_cl_lib_t *cl_lib);
-    ucc_status_t (*context_create)(ucc_cl_lib_t *cl_lib,
+    ucc_status_t (*context_create)(ucc_cl_lib_t *                 cl_lib,
                                    const ucc_cl_context_params_t *params,
                                    const ucc_cl_context_config_t *config,
-                                   ucc_cl_context_t **cl_ctx);
+                                   ucc_cl_context_t **            cl_ctx);
     ucc_status_t (*context_destroy)(ucc_cl_context_t *cl_ctx);
 } ucc_cl_iface_t;
```

src/core/ucc_context.c
```diff
@@ -204,3 +205,84 @@ void ucc_context_config_print(const ucc_context_config_h config, FILE *stream,
...
+    num_cls = config->n_cl_cfg;
+    ctx     = ucc_malloc(sizeof(ucc_context_t), "ucc_context");
+    if (!ctx) {
+        ucc_error("failed to allocate %zd bytes for ucc_context",
+                  sizeof(ucc_context_t));
+        status = UCC_ERR_NO_MEMORY;
+        goto error;
+    }
+    ctx->lib = lib;
+    memcpy(&ctx->params, params, sizeof(ucc_context_params_t));
```
alex--m, line 231: This may cause an ABI issue: if an application was built with an older UCC than the one it's running with, then this memcpy will might access a bigger ucc_context_params_t than the application gives - potentially causing a segfault (unlikely). I think a 100% future-proof solution would be to detect the memcpy size based on the field mask in params.

src/core/ucc_context.c
alex--m: I agree, doesn't seem fair to have one CL problem effect the others.

src/core/ucc_context.c
```diff
@@ -83,7 +82,8 @@ find_cl_context_config(ucc_context_config_t *cfg, ucc_cl_type_t cl_type)
 {
     int i;
     for (i = 0; i < cfg->n_cl_cfg; i++) {
-        if (cfg->configs[i] && cl_type == cfg->configs[i]->iface->type) {
+        if (cfg->configs[i] &&
+            cl_type == cfg->configs[i]->cl_lib->iface->type) {
```
alex--m, line 86: pls add ( )

src/cl/ucc_cl.h
```diff
@@ -73,15 +83,17 @@ typedef struct ucc_cl_context {
...
+    ucc_cl_##_name##_iface_t ucc_cl_##_name = {                                \
+        UCC_CL_IFACE_CFG(lib, _name, _NAME),                                   \
+        UCC_CL_IFACE_CFG(context, _name, _NAME),                               \
+        .super.super.name      = UCC_PP_MAKE_STRING(basic),                    \
+        .super.type            = UCC_CL_##_NAME,                               \
+        .super.priority        = _priority,                                    \
+        .super.init            = ucc_cl_##_name##_lib_init,                    \
+        .super.finalize        = ucc_cl_##_name##_lib_finalize,                \
+        .super.context_create  = ucc_cl_##_name##_context_create,              \
+        .super.context_destroy = ucc_cl_##_name##_context_destroy,             \
```
alex--m, line 96: some compilers hate ',' before '}', I suggest removing

src/core/ucc_context.c
vspetrov: @Sergei-Lebedev @alex--m I agree with you. One corner case: what if user specified the list of CLs explicitly? - e.g, basic,opt and, say, opt can not create ctx - is it an error then, or still just warn?

In case list of CLS is NOT specified, do we need to print warn or just "info" and proceed w/o additional output?

vspetrov:
@Sergei-Lebedev regarding the CLANG: i face this every time - this is the case we discussed but somewhat never clearly stated. Do we follow clang format for function args alignment? We can. Just note a weird "*" positioning.
```
+    ucc_status_t (*context_create)(ucc_cl_lib_t *                 cl_lib,
                                    const ucc_cl_context_params_t *params,
                                    const ucc_cl_context_config_t *config,
-                                   ucc_cl_context_t **cl_ctx);
```

It is kind of non-consistent: *params: clang keeps the '*' near the arg, for the cl_lib it is near the arg type with extra "space". 

I don't care honestly: if we all agree to just follow clang there - lets do it. @alex--m @manjugv ?

src/core/ucc_context.c
vspetrov: @alex--m by "detect based on field mask" - do you mean copying fields one-by-one checking if the bit is set?

src/core/ucc_context.c
Sergei-Lebedev: I think if user explicitly sets CL list then print warning for each CL that cannot be created. If user doesn't set CL list all failed CL simply ignored. In both cases print error if no CL opened.

alex--m: I'd even go further than Sergei and say if user specified - failure should abort. If not - warn and continue...

src/core/ucc_context.c
alex--m: I did something similar in xUCG: https://github.com/openucx/xucg/blob/531d0a7a40ab2775ba0d9123ba9acb41140ff60a/base/ucg_context.c#L226 

Not sure it's very elegant, but it gets the job done (and only calls memcpy once), and in retrospect maybe a per-field copy is actually the right thing to do here.

src/core/ucc_context.c
vspetrov: What is "abort" ? I guess returning not OK from ucc_context_create is fatal error from our perspective. But still user should decide how to process it: OMPI may fallback to another coll component ,for example.

alex--m: by abort I mean stop running (because we can't fill the user's request), which is what seems to happen today (but not what Sergei suggested).

manjugv:
> @Sergei-Lebedev regarding the CLANG: i face this every time - this is the case we discussed but somewhat never clearly stated. Do we follow clang format for function args alignment? We can. Just note a weird "*" positioning.
> 
> ```
> +    ucc_status_t (*context_create)(ucc_cl_lib_t *                 cl_lib,
>                                     const ucc_cl_context_params_t *params,
>                                     const ucc_cl_context_config_t *config,
> -                                   ucc_cl_context_t **cl_ctx);
> ```
> 
> It is kind of non-consistent: _params: clang keeps the '_' near the arg, for the cl_lib it is near the arg type with extra "space".
> 
> I don't care honestly: if we all agree to just follow clang there - lets do it. @alex--m @manjugv ?

We capture with the example on how to handle it. 
https://github.com/openucx/ucc/blob/master/docs/codestyle.md

src/core/ucc_context.c
manjugv: Given that this is prevalent in UCC, it might be worth to create a copy routine. Between, how does UCX handle this?

src/core/ucc_context.c
manjugv: Seems like both Alex and Sergey have captured the issue very well.

src/core/ucc_context.c
vspetrov: btw, should be info or debug? i'm not sure. Do we have any rule of thumb what is info/debug?

src/core/ucc_context.c
alex--m: UCX doesn't copy params at any point. It goes field-by-field and sets internal variables/structures based on params (not calling memcpy).

src/core/ucc_context.c
alex--m: feels like 'debug' to me - this doesn't really provide any information the user can benefit from... to me, info feels like non-trivial stuff, like the transports/components acutally selected during the run (something the user would have a hard time deducing).

vspetrov:
closing in favor of #41 (as discussed in WG)

PR #39, opened 2020-12-07
BUILD: Fixing make docs

manjugv:
## What
Make docs depends on UCX

## Why ?
No need to depend on UCX, when requesting only to build the doxygen spec

## How ?
Disable check for UCX

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -142,10 +142,10 @@ AS_IF([test "x$with_docs_only" = xyes],
      fi
 
      CHECK_UCX
-        AC_MSG_RESULT([UCX support: $ucx_happy])
-        if test $ucx_happy != "yes"; then
-            AC_MSG_ERROR([UCX is not available])
-        fi
+     AC_MSG_RESULT([UCX support: $ucx_happy])
+     if test $ucx_happy != "yes"; then
+         AC_MSG_ERROR([UCX is not available])
+     fi
     ]) # Docs only
 
 
--- a/configure.ac
+++ b/configure.ac
@@ -79,12 +79,6 @@ CFLAGS="$CFLAGS_save"
 # Additional m4 files
 #
 m4_include([config/m4/ax_prog_doxygen.m4])
-m4_include([config/m4/compiler.m4])
-m4_include([config/m4/gtest.m4])
-m4_include([config/m4/sysdep.m4])
-m4_include([config/m4/ucx.m4])
-
-m4_include([test/gtest/configure.m4])
 
 AC_ARG_WITH([docs_only],
         AS_HELP_STRING([--with-docs-only],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -117,30 +111,45 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+     else
+	    AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_INFO], [Highest log level])
+	    CFLAGS="$CFLAGS -O3 -DNDEBUG"
+     fi
+
+     CHECK_UCX
+        AC_MSG_RESULT([UCX support: $ucx_happy])
+        if test $ucx_happy != "yes"; then
+            AC_MSG_ERROR([UCX is not available])
+        fi
```
Sergei-Lebedev, line 148: alignment, should be on the same vertical line as CHECK_UCX

Sergei-Lebedev (approved):

vspetrov (approved):

configure.ac
```diff
@@ -117,30 +111,45 @@ AS_IF([test "x$with_docs_only" = xyes],
...
-AC_MSG_RESULT([UCX support: $ucx_happy])
-if test $ucx_happy != "yes"; then
-   AC_MSG_ERROR([UCX is not available])
-fi
+     if test $enable_debug = yes; then
+	    AC_DEFINE([ENABLE_DEBUG], [1], [Enable debugging code])
+	    CFLAGS="$CFLAGS -O0 -g3"
+	    AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_TRACE_POLL], [Highest log level])
+     else
+	    AC_DEFINE([UCS_MAX_LOG_LEVEL], [UCS_LOG_LEVEL_INFO], [Highest log level])
```
alex--m, line 140: Probably not the highest log level

configure.ac
```diff
@@ -117,30 +111,45 @@ AS_IF([test "x$with_docs_only" = xyes],
      UCC_DX_ENABLE_CHECK([man])
      UCC_DX_ENABLE_CHECK([pdf])
      AM_CONDITIONAL([DOCS_ONLY], [true])
+     AM_CONDITIONAL([HAVE_GTEST], [false])
+     AM_CONDITIONAL([HAVE_CXX11], [false])
+     AM_CONDITIONAL([HAVE_GNUXX11], [false])
+     AM_CONDITIONAL([HAVE_GLIBCXX_NOTHROW], [false])
+     AM_CONDITIONAL([HAVE_AARCH64_THUNDERX2], [false])
```
alex--m, line 118: Do we use these arch-specific definitions today?

manjugv: Yep, it is coming from config/m4/compiler.m4. It is because we are starting to use UCS.

configure.ac
manjugv: I’m not sure what you mean here. I guess when in debug mode MAX_LOG_LEVEL is the highest level and in non-debug mode, LOG_LEVEL_INFO is the highest level. Is the description an issue, or the value? 
(Btw, this PR is not adding this line, just moving the line around. @vspetrov  can fill in better.)

alex--m: Ok, I initially thought that was a copy-paste mistake, saying that INFO is the highest log level possible (which it isn't)...

alex--m (approved):

PR #40, opened 2020-12-08
Proposal for exposing SET/GET for CL objects, e.g. a UCX context 

alex--m:
This is what we talked about during the last meeting we held - about allowing UCC to either re-use an existing CL-specific context or share the one it created during initialization.

Also, I noticed that currently the API file seems to have some non-ascii characters - I fixed it on the way (my IDE was adamant about it).


alex--m:
looks like UCX community can live with UCX object sharing... just need to be careful about it.
I'm now officially putting it up for review.

manjugv:
Dependent on the discussion of #48

alex--m:
> 
> 
> Dependent on the discussion of #48

No, I don't see it this way. 
This proposal was submitted prior to #48, and yet received no feedback before you presented your alternative suggestion. If you think there's a use-case not covered by this proposal - please write so here.

manjugv:
(Back from the holidays and break)

In the WG meeting, the WG provided initial feedback. Adding that info and adding more

1. Split the PR into editorial (non-ASCII character) changes and resource sharing part. We are in agreement with the editorial changes.
2. Add details about the semantics of the functionality to the Doxygen documentation.
3. Add details about motivation, how it is done, and what is done (PR template)

My main concern is that the proposal and solution are designed around addressing the lack of certain interfaces/features in UCX. My preference is to design a solution that evolves irrespective of UCX interfaces and serves the purpose of resource sharing. 

PR #48 is an attempt to comprehensively address resource sharing with lazy resource association for UCC.  PR #48 enables resource sharing at different granularities - both fine-grained and coarse-grained. With the same single interface, it enables the sharing of resources at the library, context, team, or endpoint level. I want to stress that it only enables but is not required to specify it. We could specify as we see deemed necessary. So, it is future proof, without much API churn.

PR #48 also enables dynamic resource association. This means there is a clear way to associate and disassociate resources and can be clearly specified. For example, when the p2p library creates and destroys the resources, there is a clear way to remove that association. In this PR, I’m not sure how to do that. (Maybe you already know this. It is not clear from the PR writeup.)

Though our immediate use case is sharing UCX workers, one can foresee such requirements in resource sharing. For example, we might want to share shared memory regions, GPU streams, and endpoints.

My intention is to make a case for general resource sharing than push PR #48 as it is (that is why the other PR is not proposed for review yet). I’m not fixated on the exact solution, yet.

As WG, let's decide whether to support
1. Sharing UCX worker
2. General purpose resource sharing 
3. Design general resource API and then support only UCX worker sharing at this point
Based on (1), (2), or (3) we could choose this PR, variation of PR #48, or some other solution.

manjugv:
UCC WG (01/20/21): WG decided to delay the discussion on this PR and PR #48 until the discussion on OMPI context sharing discussion is concluded.

manjugv:
@alex--m  Can you please make a separate PR for the non-ASCII character fix? That part we can merge.

alex--m:
UCX WG (03/02/21): UCX developers condition the OMPI patch upon the merge of a resource sharing API in UCC. I'll separate the ASCII fix and rebase over the latest master, and we can get back to reviewing.

alex--m:
WG (02/10/21): We weren't able to reach consensus regarding the resource-sharing API.

@manjugv feels the API must impose more semantics and limitations on the user, whereas this PR is just passing a void* and leaves all the responsibility to the internal component using it. 

On the other hand, I'm against imposing more restrictions and limitations in the objects without a clear benefit for it. I feel this would complicate and limit the code inside UCC (I prefer the flexibility/future-proofing of a generic pointer). We could only agree to some minimal semantics, for instance the passed object (e.g. UCP worker) should remain valid until UCC is finalized.

@vspetrov brought up the issue of tag-space, namely that aside from the UCP worker object we would need to pass more information anyway - the tags UCC could use, for example, seeing as there's a potential conflict between the tags used by UCC and the tags used by the user's MPI application.

I should point out here that this conflict is true for both this PR and #48, since UCP will not protect from conflicts between UCC and other components, namely the MPI application itself. For reference, @yosefe mentioned the following hazards:
- Tag : tags space collision
- AM : ID space collision
- RMA : fence/flush scope
- Stream : poll loop of endpoints

@bureddy suggested we could go back to our alternative approach - which is to ask UCP to resolve this problem: one way would be to ask UCP to fetch the UCP worker (instead of having the UCC init code pass it explicitly), and another way would be to ask UCP to partition/multiplex the "conflict spaces" (e.g. Tag space and AM ID space).

Finally, we've decided that provided we have data on the benefits of resource sharing (I'll work on that), we, as a working group, would **make a request to the UCX community to provide a new (UCP) API to prevent the potential conflict between two components using the same UCP worker**.

yosefe:
I dont see a way to separate namespaces without either increasing headers sizes or allocating more hw resources. The former is not acceptable, and the latter defeats the purpose (??) of worker sharing.
Whatever component decides to share its ucx objects is also responsible for coordinating the id space, it's outside the scope of ucx itself.

manjugv:
> @manjugv feels the API must impose more semantics and limitations on the user, whereas this PR is just passing a void* and leaves all the responsibility to the internal component using it.

I would not state that way. :) For me, the library interface is a contract between the user and the library implementer. This property is essential for the library to be portable and also implementers to support various network transports and for many other reasons. My concern with the PR is that we are under specifying it.

shamisp:
> I dont see a way to separate namespaces without either increasing headers sizes or allocating more hw resources. The former is not acceptable, and the latter defeats the purpose (??) of worker sharing.
> Whatever component decides to share its ucx objects is also responsible for coordinating the id space, it's outside the scope of ucx itself.

1. If you allocate 2x resources in one place and reduce 10x in another it still may make sense. For example, if I add one bit into tag space and reduce number of QPs by x1000 why not. 
2. If this helps, we can potentially simplify ID coordination by providing a query on worker or context object to report what tag space is unused / available. Regardless, I think it will introduce at least one branch (likely) in critical path to check validity of the tag.

yosefe:
> * If you allocate 2x resources in one place and reduce 10x in another it still may make sense. For example, if I add one bit into tag space and reduce number of QPs by x1000 why not.

Ok, agree with this, but to evaluate a specific suggestion we need some example of "increasing namespace by X improves something by Y"

> * If this helps, we can potentially simplify ID coordination by providing a query on worker or context object to report what tag space is unused / available. Regardless, I think it will introduce at least one branch (likely) in critical path to check validity of the tag.

For active messages, we can have a query API to check if a specific ID is used, or even allocate an available id; it seems generic enough.
For tags, we don't really allocate them today, the sender/receiver can just use any tags they want. What do we propose here?

yosefe:
How about passing a UCP task mask along with the UCP worker object in the UCC API to indicate which tag namespace available for each user (coll/p2p)?
Or using active messages API which should make it easier to share namespaces (since we expect to have only a small number of allocated IDs by each user)?

vspetrov:
> How about passing a UCP tag mask along with the UCP worker object in the UCC API to indicate which tag namespace available for each user (coll/p2p)?
Do you mean that upper level (OMPI) will define the tag mask that specifies the bits that UCC can safely  set in the tag ? This could work unless the space we are given is too small. 

> Or using active messages API which should make it easier to share namespaces (since we expect to have only a small number of allocated IDs by each user)?
AM approach is much easier from the perspective of sharing. I have only 1 question: does this mean UCC will have to build rndvz protocol on top of AM ?

shamisp:
@yosefe according to MPI spec:
`Tag values range from 0 to the value returned for MPI_TAG_UB, inclusive. These values are
guaranteed to be unchanging during the execution of an MPI program. In addition, the tag
upper bound value must be at least 32767. An MPI implementation is free to make the
value of MPI_TAG_UB larger than this; for example, the value 2^30 − 1 is also a valid value
for MPI_TAG_UB. The attribute MPI_TAG_UB has the same value on all processes of MPI_COMM_WORLD`

We can use the MSB (sign) to make negative values for collectives.  Considering that `typedef uint64_t                       ucp_tag_t;` it should give us some space.

shamisp:
> > How about passing a UCP tag mask along with the UCP worker object in the UCC API to indicate which tag namespace available for each user (coll/p2p)?
> 
> Do you mean that upper level (OMPI) will define the tag mask that specifies the bits that UCC can safely set in the tag ? This could work unless the space we are given is too small.
> 

Well, if you use very large number of communicators you can run into all kinda of issues.

> > Or using active messages API which should make it easier to share namespaces (since we expect to have only a small number of allocated IDs by each user)?,
> 
> AM approach is much easier from the perspective of sharing. I have only 1 question: does this mean UCC will have to build rndvz protocol on top of AM ?

It can be royal pain, imho. It is much easier to go send/recv route since most of the algorithms already implemented this way.

yosefe:
> It can be royal pain, imho. It is much easier to go send/recv route since most of the algorithms already implemented this way.

AM API already supports rndv protocol inside UCX. It actually shares 98% of code with the tag-based Rndv implementation.
I would suggest looking into this API (regardless of object sharing) since it provides additional flexibility and can avoid recv memcopy in some cases.

shamisp:
@yosefe Depends how the existing algorithms are implemented (I guess UCC collectives are ported from somewhere else) it might be painful to port from send/recv to active message.

vspetrov:
@manjugv can we close this? It is outdated and does not contain any solution. We can open "issue" to track that, but the way i see it we don't have a solution in near term

manjugv:
WG Discussion: Close for now and open if needed

PR #41, opened 2020-12-09
ucc base iface

vspetrov:
## What
ucc_base_iface.h/c contains definitions of the interfaces common
for CL and TL frameworks. Each framework will specify its own
implementation (this PR adds implementation for CL).

- ucc_context_create/destroy implementation
- cl context create/destroy iface callbacks
- ucc_cl_context_t class
- minimal ctx create/destroy implementation for cl_basic

## Why ?
core api


src/cl/basic/cl_basic_context.c
```diff
@@ -0,0 +1,27 @@
...
+#include "cl/ucc_cl_log.h"
+#include "utils/ucc_malloc.h"
+
+UCC_CLASS_INIT_FUNC(ucc_cl_basic_context_t,
+                    const ucc_base_context_params_t *params,
+                    const ucc_base_config_t *config)
+{
+    ucc_cl_context_config_t *cl_config =
+        ucc_derived_of(config, ucc_cl_context_config_t);
+    UCC_CLASS_CALL_SUPER_INIT(ucc_cl_context_t, cl_config->cl_lib);
```
Sergei-Lebedev, line 17: Minor, I think self->super->super.lib looks better here rather than cl_config->cl_lib

src/cl/basic/cl_basic_context.c
```diff
@@ -0,0 +1,27 @@
...
+#include "cl_basic.h"
+#include "cl/ucc_cl_log.h"
+#include "utils/ucc_malloc.h"
+
+UCC_CLASS_INIT_FUNC(ucc_cl_basic_context_t,
+                    const ucc_base_context_params_t *params,
+                    const ucc_base_config_t *config)
+{
+    ucc_cl_context_config_t *cl_config =
+        ucc_derived_of(config, ucc_cl_context_config_t);
```
Sergei-Lebedev, line 16: Does compiler allow to discard const for config? Looks like ucc_cl_context_config_t  should be const pointer

src/cl/ucc_base_iface.h
```diff
@@ -0,0 +1,69 @@
...
+
+typedef struct ucc_base_context_config {
+    /* C standard prohibits empty structures */
+    char __dummy;
+} ucc_base_context_config_t;
+
+typedef struct ucc_base_context_params {
+    ucc_context_params_t params;
+} ucc_base_context_params_t;
+UCC_CLASS_DECLARE(ucc_base_context_params_t, const ucc_context_params_t *);
```
Sergei-Lebedev, line 46: why ucc_base_contetext_params_t is a class?

src/cl/ucc_base_iface.h
```diff
@@ -0,0 +1,69 @@
...
+typedef struct ucc_base_lib_iface {
+    ucc_status_t (*init)(const ucc_base_lib_params_t *params,
+                         const ucc_base_config_t *config, ucc_base_lib_t **lib);
+    void         (*finalize)(ucc_base_lib_t *lib);
+} ucc_base_lib_iface_t;
+
+typedef struct ucc_base_context_config {
+    /* C standard prohibits empty structures */
+    char __dummy;
+} ucc_base_context_config_t;
```
Sergei-Lebedev, line 41: where do we use it? cl context config derived from base config

src/cl/basic/cl_basic_context.c
vspetrov: not sure what do you mean. self->super->super.lib is not defined there. In turn it is being set in the superclass constructor from cl_config->cl_lib

src/cl/basic/cl_basic_context.c
vspetrov: fixed (compiler allowed probably due to derived_of casting, it is strange though)

src/cl/ucc_base_iface.h
vspetrov: 👍

src/cl/ucc_base_iface.h
vspetrov: 👍

src/cl/basic/cl_basic_context.c
Sergei-Lebedev: you right, thanks

vspetrov:
@Sergei-Lebedev anything else here? @manjugv @alex--m plz have a look

src/cl/ucc_base_iface.h
```diff
@@ -0,0 +1,63 @@
...
+                                  ucc_config_global_list_entry_t *cfg_entry,
+                                  ucc_base_config_t **config);
+
+static inline void ucc_base_config_release(ucc_base_config_t *config)
+{
+    ucc_config_parser_release_opts(config, config->cfg_entry->table);
+    free(config);
+}
+
+#endif
```
manjugv, line 63: Do we need a distinction between ucc_cl and ucc_base_iface ? It is not clear to me what purpose would it serve to differentiate between the two ?

src/core/ucc_context.c
```diff
@@ -204,3 +190,100 @@ void ucc_context_config_print(const ucc_context_config_h config, FILE *stream,
...
+    ctx     = ucc_malloc(sizeof(ucc_context_t), "ucc_context");
+    if (!ctx) {
+        ucc_error("failed to allocate %zd bytes for ucc_context",
+                  sizeof(ucc_context_t));
+        status = UCC_ERR_NO_MEMORY;
+        goto error;
+    }
+    ctx->lib = lib;
+    ucc_copy_context_params(&ctx->params, params);
+    ucc_copy_context_params(&b_params.params, params);
```
manjugv, line 228: Do you need to copy? Do you expect the structures to change in future?

manjugv:
In general I like the direction. My high-level comments. 

* Do we need a distinction between ucc_cl and ucc_base_iface ? It is not clear to me what purpose would it serve to differentiate between the two ? 

* From the code organization perspective, we should probably have a base directory, thoughts ?

src/cl/basic/cl_basic_context.c
```diff
@@ -0,0 +1,27 @@
...
+                    const ucc_base_config_t *config)
+{
+    const ucc_cl_context_config_t *cl_config =
+        ucc_derived_of(config, ucc_cl_context_config_t);
+    UCC_CLASS_CALL_SUPER_INIT(ucc_cl_context_t, cl_config->cl_lib);
+    cl_info(cl_config->cl_lib, "initialized cl context: %p", self);
+    return UCC_OK;
+}
+
+UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_context_t)
```
manjugv, line 22: Minor: Should we use INIT and FINALIZE, instead of INIT and CLEANUP.  Probably, you are following UCX style.

manjugv (dismissed):

vspetrov:
> In general I like the direction. My high-level comments.
> 
> * Do we need a distinction between ucc_cl and ucc_base_iface ? It is not clear to me what purpose would it serve to differentiate between the two ?
> * From the code organization perspective, we should probably have a base directory, thoughts ?

@manjugv  The key idea why we have "base iface" is that we have 2 slightly different derivatives: CL interface and TL interface. CL is in the PR right now. TL is ready in my local repo i'll push soon. Those 2 inherit from the common "base"

src/cl/basic/cl_basic_context.c
vspetrov: I would keep macros as is. We reuse UCS class functionality. We could rename. But for "finalize" for class destructor sounds a bit weird imho.

test/gtest/core/test_context.cc
```diff
@@ -0,0 +1,47 @@
...
+    for (int i = 0; i < n_ctxs; i++) {
+        EXPECT_EQ(UCC_OK, ucc_context_create(lib_h, &ctx_params, ctx_config, &ctx_h));
+        ctxs.push_back(ctx_h);
+    }
+
+    std::shuffle(ctxs.begin(), ctxs.end(), std::default_random_engine());
+    for (auto ctx_h : ctxs) {
+        EXPECT_EQ(UCC_OK, ucc_context_destroy(ctx_h));
+    }
+}
```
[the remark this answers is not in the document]
vspetrov: ./test/gtest/gtest --gtest_filter=test_context_config* 
or 
./test/gtest/gtest --gtest_filter=test_context*

It is wildcard-style matching based on the test-fixture name.

src/core/ucc_context.c
vspetrov: There are 2 copies here: 1st - we cache user input on ucc_context_t. Main reason: there is OOB in params, that we will need. 2nd copy: to pass params down into CL/TL interfaces we need to pack it into "base_params" according to iface. THose copies are super-tiny and hence super fast and in the control path. should be ok i think.

src/core/ucc_context.c
```diff
@@ -204,3 +190,100 @@ void ucc_context_config_print(const ucc_context_config_h config, FILE *stream,
...
+    return UCC_OK;
+
+error_ctx_create:
+    for (i = i - 1; i >= 0; i--) {
+        config->configs[i]->cl_lib->iface->context.destroy(
+            &ctx->cl_ctx[i]->super);
+    }
+    free(ctx->cl_ctx);
+error_ctx:
+    free(ctx);
```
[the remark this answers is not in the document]
vspetrov: 👍

src/cl/basic/cl_basic.c
```diff
@@ -0,0 +1,41 @@
...
+     UCC_CONFIG_TYPE_TABLE(ucc_cl_lib_config_table)},
+
+    {NULL}
+};
+
+static ucs_config_field_t ucc_cl_basic_context_config_table[] = {
+    {"", "", NULL, ucc_offsetof(ucc_cl_basic_context_config_t, super),
+     UCC_CONFIG_TYPE_TABLE(ucc_cl_context_config_table)},
+
+    {"TEST_PARAM", "5", "For dbg test purpuse : don't commit",
```
alex--m, line 22: did you change your mind about committing? :)

src/cl/ucc_base_iface.c
```diff
@@ -0,0 +1,36 @@
...
+    {"LOG_LEVEL", "warn",
+     "UCC logging level. Messages with a level higher or equal to the "
+     "selected will be printed.\n"
+     "Possible values are: fatal, error, warn, info, debug, trace, data, func, "
+     "poll.",
+     ucc_offsetof(ucc_base_config_t, log_component), UCC_CONFIG_TYPE_LOG_COMP},
+
+    {NULL}};
+
+ucc_status_t ucc_base_config_read(const char *full_prefix,
```
alex--m, line 15: This looks awfully familiar... does this mean every "framework", e.g. CL/TL/..., will have this kind of function? could we merge all these functions into one?

src/cl/ucc_cl.c
```diff
@@ -99,3 +94,37 @@ ucc_status_t ucc_parse_cls_string(const char *cls_str,
...
+    return UCC_OK;
+}
+
+UCC_CLASS_CLEANUP_FUNC(ucc_cl_context_t)
+{
+}
+
+UCC_CLASS_DEFINE(ucc_cl_context_t, void);
+
+ucc_status_t ucc_cl_context_config_read(ucc_cl_lib_t *cl_lib,
```
alex--m, line 109: Same comment as before - maybe I'm missing something, but isn't it easier to have just one config read function?

vspetrov: Replying to both comments: essentially there is one common function - ucc_base_config_read  which reads config using ucs. CL/TL versions are just slightly different: they take config tables from different structs and take slightly different args (as defined by tl and cl iface). So, essentially the functions in CL/TL are convenience implementations with more clean implementation (specific to CL or TL)

src/cl/basic/cl_basic.c
vspetrov: it is convenient for testing so far. i'll remove this as soon as we add 1st meaningful parameter. if there are objections i can remove it right now.

vspetrov:
@manjugv applied changes we talked about: moved base iface to "src/components". cl and tl are moved there as well as we discussed. @alex--m i've replied to your comments, plz respond. @Sergei-Lebedev anything else?

alex--m (approved):

manjugv (approved):

manjugv:
@vspetrov It looks good. Do you want to move ucc_base_iface.c/h to base directory ?

Sergei-Lebedev (approved):
Agree with @manjugv, move base iface to base directory, otherwise looks good to me

vspetrov:
> @vspetrov It looks good. Do you want to move ucc_base_iface.c/h to base directory ?

done

PR #42, opened 2020-12-13
Fix include installation path to include '/ucc'

alex--m:
Currently, the path is just `$includedir/api/ucc.h`, so if `$includedir` is `/usr/include` - the installed header will be at `/usr/include/api/ucc.h` ... makes much more sense to have it in `/usr/include/ucc/api/ucc.h` (which is also what UCX does, BTW).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -38,4 +38,4 @@ libucc_la_SOURCES =        \
 	utils/ucc_component.c  \
 	cl/ucc_cl.c
 
-libucc_ladir = $(includedir)
+libucc_ladir = $(includedir)/ucc


alex--m:
There's a problem with this solution when we try Open-MPI's `--with-ucc=/opt/ucc` detection (and will happen in other cases as well):
````
configure:173918: /mnt/c/Users/a00467620/workspace/spack/opt/spack/linux-ubuntu18.04-skylake/gcc-10.1.0/gcc-10.1.0-sklchayarldyfbxagu34m5pkf37lrewj/bin/gcc -c -g -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing -mcx16   -I/home/alex/workspace/ompi/3rd-party/hwloc-2.2.0/include -I/home/alex/workspace/ompi/3rd-party/hwloc-2.2.0/include -I/home/alex/workspace/ompi/3rd-party/openpmix/include -I/home/alex/workspace/ompi/3rd-party/openpmix/include -I/home/alex/workspace/ucc/build/include conftest.c >&5
In file included from conftest.c:641:
/home/alex/workspace/ucc/build/include/ucc/api/ucc.h:12:10: fatal error: api/ucc_def.h: No such file or directory
   12 | #include <api/ucc_def.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
````

The problem is that our source tree has `src/api/ucc.h` - so the source files call `#include <api/ucc_def.h>`, and this doesn't work if I move `/usr/include/api/ucc.h` to be `/usr/include/ucc/api/ucc.h` (because the `/ucc` is missing).

I can only see two options:

1. change the source tree to make `#include <ucc/api/ucc_def.h>` work (this is what UCX have)
2. stick to installation under `/usr/include/api/ucc.h`, which looks pretty bad to me

@manjugv @vspetrov @Sergei-Lebedev  - Any thoughts?

vspetrov:
@alex--m I totally agree with you - we should install into /include/ucc. This is how i've fixed it: https://github.com/vspetrov/ucc/commit/10418893f8a79645ff7bc4544f5cdb1b28a8c01c 
plz comment.

Basically, i've moved src/api -> src/ucc/api and changed nobase_dist_libucc_la_HEADERS accordingly (adding prefix ucc/).

alex--m:
@vspetrov I read the patch and I agree with it - but it looks like it's part of a large PR... and I need it to move forward. Sure, I can apply this locally, but I prefer we merge it asap. Would you mind creating a dedicated PR for it? (the new PR would replace this one)

vspetrov:
will do now

vspetrov:
@alex--m  opened PR: https://github.com/openucx/ucc/pull/44

vspetrov:
resolved via #44 , closing

PR #43, opened 2020-12-14
TL framework

vspetrov:
## What
Adds TL Framework and interface
Adds TL/UCP lib init/finalize, context create/destroy

## Why ?
TLs are internal "transport" layers. Their interface is very similar to CL and it is inherited from the "base" interface (similarly to CL)

### DIFF
--- a/src/components/base/ucc_base_iface.h
+++ b/src/components/base/ucc_base_iface.h
@@ -38,6 +38,10 @@ typedef struct ucc_base_lib_iface {
 
 typedef struct ucc_base_context_params {
     ucc_context_params_t params;
+    int                  estimated_num_eps;
+    int                  estimated_num_ppn;
+    ucc_thread_mode_t    thread_mode;
+    const char          *prefix;
 } ucc_base_context_params_t;
 
 typedef struct ucc_base_context {
@@ -60,5 +64,33 @@ static inline void ucc_base_config_release(ucc_base_config_t *config)
     ucc_config_parser_release_opts(config, config->cfg_entry->table);
     ucc_free(config);
 }
+
+#define UCC_IFACE_NAME_PREFIX(_F, _NAME, _cfg)                                 \
+    .name   = UCC_PP_MAKE_STRING(_F##_NAME) " " UCC_PP_MAKE_STRING(_cfg),      \
+    .prefix = UCC_PP_MAKE_STRING(_F##_NAME##_)
+
+#define UCC_IFACE_CFG(_F, _f, _cfg, _name, _NAME)                              \
+    .super._f##_cfg##_config = {                                               \
+        UCC_IFACE_NAME_PREFIX(_F, _NAME, _cfg),                                \
+        .table = ucc_##_f##_name##_##_cfg##_config_table,                      \
+        .size  = sizeof(ucc_##_f##_name##_##_cfg##_config_t)}
+
+#define UCC_BASE_IFACE_DECLARE(_F, _f, _name, _NAME, ...)                      \
+    ucc_##_f##_name##_iface_t ucc_##_f##_name = {                              \
+        UCC_IFACE_CFG(_F, _f, lib, _name, _NAME),                              \
+        UCC_IFACE_CFG(_F, _f, context, _name, _NAME),                          \
+        .super.super.name = UCC_PP_MAKE_STRING(_name),                         \
+        __VA_ARGS__                                                            \
+        .super.lib.init   = UCC_CLASS_NEW_FUNC_NAME(ucc_##_f##_name##_lib_t),  \
+        .super.lib.finalize =                                                  \
+            UCC_CLASS_DELETE_FUNC_NAME(ucc_##_f##_name##_lib_t),               \
+        .super.context.create =                                                \
+            UCC_CLASS_NEW_FUNC_NAME(ucc_##_f##_name##_context_t),              \
+        .super.context.destroy =                                               \
+        UCC_CLASS_DELETE_FUNC_NAME(ucc_##_f##_name##_context_t)};              \
+    UCC_CONFIG_REGISTER_TABLE_ENTRY(&ucc_##_f##_name.super._f##lib_config,     \
+                                    &ucc_config_global_list);                  \
+    UCC_CONFIG_REGISTER_TABLE_ENTRY(&ucc_##_f##_name.super._f##context_config, \
+                                    &ucc_config_global_list)
 
 #endif
--- a/src/components/cl/ucc_cl.h
+++ b/src/components/cl/ucc_cl.h
@@ -63,32 +63,8 @@ typedef struct ucc_cl_context {
 
 UCC_CLASS_DECLARE(ucc_cl_context_t, ucc_cl_lib_t *);
 
-#define UCC_CL_IFACE_NAME_PREFIX(_NAME, _cfg)                                  \
-    .name   = UCC_PP_MAKE_STRING(CL_##_NAME) " " UCC_PP_MAKE_STRING(_cfg),     \
-    .prefix = UCC_PP_MAKE_STRING(CL_##_NAME##_)
-
-#define UCC_CL_IFACE_CFG(_cfg, _name, _NAME)                                   \
-    .super.cl_##_cfg##_config = {                                              \
-        UCC_CL_IFACE_NAME_PREFIX(_NAME, _cfg),                                 \
-        .table = ucc_cl_##_name##_##_cfg##_config_table,                       \
-        .size  = sizeof(ucc_cl_##_name##_##_cfg##_config_t)}
+#define UCC_CL_IFACE_EXT(_NAME) .super.type = UCC_CL_ ## _NAME,
 
 #define UCC_CL_IFACE_DECLARE(_name, _NAME)                                     \
-    ucc_cl_##_name##_iface_t ucc_cl_##_name = {                                \
-        UCC_CL_IFACE_CFG(lib, _name, _NAME),                                   \
-        UCC_CL_IFACE_CFG(context, _name, _NAME),                               \
-        .super.super.name = UCC_PP_MAKE_STRING(_name),                         \
-        .super.type       = UCC_CL_##_NAME,                                    \
-        .super.lib.init   = UCC_CLASS_NEW_FUNC_NAME(ucc_cl_##_name##_lib_t),   \
-        .super.lib.finalize =                                                  \
-            UCC_CLASS_DELETE_FUNC_NAME(ucc_cl_##_name##_lib_t),                \
-        .super.context.create =                                                \
-            UCC_CLASS_NEW_FUNC_NAME(ucc_cl_##_name##_context_t),               \
-        .super.context.destroy =                                               \
-            UCC_CLASS_DELETE_FUNC_NAME(ucc_cl_##_name##_context_t)};           \
-    UCC_CONFIG_REGISTER_TABLE_ENTRY(&ucc_cl_##_name.super.cl_lib_config,       \
-                                    &ucc_config_global_list);                  \
-    UCC_CONFIG_REGISTER_TABLE_ENTRY(&ucc_cl_##_name.super.cl_context_config,   \
-                                    &ucc_config_global_list)
-
+    UCC_BASE_IFACE_DECLARE(CL_, cl_, _name, _NAME, UCC_CL_IFACE_EXT(_NAME))
 #endif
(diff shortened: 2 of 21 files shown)


src/components/base/ucc_base_iface.h
```diff
@@ -38,6 +38,10 @@ typedef struct ucc_base_lib_iface {
 
 typedef struct ucc_base_context_params {
     ucc_context_params_t params;
+    int                  estimated_num_eps;
+    int                  estimated_num_ppn;
```
manjugv, line 42: How to get values for this, from config? Is this maximum? What happens, if I violate it? Is this relevant beyond UCP tl?

src/components/tl/ucp/tl_ucp.h
```diff
@@ -0,0 +1,48 @@
...
+/* Extern iface should follow the pattern: ucc_tl_<tl_name> */
+extern ucc_tl_ucp_iface_t ucc_tl_ucp;
+
+typedef struct ucc_tl_ucp_lib_config {
+    ucc_tl_lib_config_t super;
+} ucc_tl_ucp_lib_config_t;
+
+typedef struct ucc_tl_ucp_context_config {
+    ucc_tl_context_config_t super;
+    int                     test_param;
```
manjugv, line 26: ?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -0,0 +1,126 @@
...
+        ucp_params.field_mask |= UCP_PARAM_FIELD_ESTIMATED_NUM_PPN;
+        ucp_params.estimated_num_ppn = params->estimated_num_ppn;
+    }
+
+    if (params->estimated_num_eps > 0) {
+        ucp_params.field_mask |= UCP_PARAM_FIELD_ESTIMATED_NUM_EPS;
+        ucp_params.estimated_num_eps = params->estimated_num_eps;
+    }
+
+    status = ucp_init(&ucp_params, ucp_config, &ucp_context);
```
manjugv, line 64: I would have expected ucp_init to be in the library init. We might create multiple context create and you don’t want to init that multiple times? I’m missing something.

manjugv: Is the behavior of params the same as UCC params? .i.e., it is more of a hints rather than constraints?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -0,0 +1,126 @@
...
+        worker_params.thread_mode = UCS_THREAD_MODE_SINGLE;
+        break;
+    case UCC_THREAD_MULTIPLE:
+        worker_params.thread_mode = UCS_THREAD_MODE_MULTI;
+        break;
+    default:
+        /* unreachable */
+        ucc_assert(0);
+        break;
+    }
```
manjugv, line 86: If UCC context is created in SHARED mode, do you need UCS_THREAD_MODE_MULTIPLE, even in UCC_THREAD_SINGLE ? I don’t for sure either way, but it seems like a corner case that can cause problems. Thoughts?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -0,0 +1,126 @@
...
+
+    if (params->thread_mode == UCC_THREAD_MULTIPLE) {
+        worker_attr.field_mask = UCP_WORKER_ATTR_FIELD_THREAD_MODE;
+        ucp_worker_query(ucp_worker, &worker_attr);
+        if (worker_attr.thread_mode != UCS_THREAD_MODE_MULTI) {
+            tl_error(&self->super,
+                     "thread mode multiple is not supported by ucp worker");
+            status = UCC_ERR_NOT_SUPPORTED;
+            goto err_thread_mode;
```
manjugv, line 102: From the UCC perspective, it is ok to return single thread mode, even if multithreaded was requested.

manjugv (changes requested):
At this point, the CL and TL are very close. I expect this to evolve. However, we should document the role of each of these abstractions. Probably, we can add it as a part of tl. h and cl. h files - Doxygen style.

src/components/base/ucc_base_iface.h
vspetrov: 1. Various ways to get them:
- e.g., in xccl we compute that information (by performing 1 bootstrap allgather on a context)
- we can (and SHOULD) ask this info from the runtime, probably yes via context config

2. Yes, it is estimated max
3. It is perf optimization in UCP, nothing crashes if you don't pass it but perf difference is HUGE
4. Might be relevant for SHM TL as well (in hcoll we use that info to implement shmseg bcol efficiently)

src/components/tl/ucp/tl_ucp.h
vspetrov: Doxygen style - does this mean it will be part of document build? Do we want to expose that to users? 
C -style comment is not enough? 
What do you want me to describe: overall CL/TL idea or details for all the fields and params in the structs?

src/components/tl/ucp/tl_ucp_context.c
vspetrov: Yes, you would think that ucp_init should go to lib_init BUT there reason it is in context is:
ucp_init is called after ucp_config_read. If you want to create ucp worker on a specific NET iface you need to read ucp config, then modify "UCX_NET_DEVICES" value and then create ucp_context + worker. 
This kind of means that in order to represent another device you need another "ucp_context". This is why it is here. (same in xccl)

regarding 2: yes params are very similar, rather hints

src/components/tl/ucp/tl_ucp_context.c
vspetrov: @manjugv again i see that you use "UCC_CONTEXT_SHARED" in a different manner than you described to WG and in the manual. Previously, you were saying the SHARED has nothing to do with threading. It just means that multiple teams can be created from a ucc_context. If user wants to run collectives on those teams in a threaded env user HAS TO set THREAD MULTIPLE.

Plz, clarify again what UCC_CONTEXT_SHARED means?

src/components/tl/ucp/tl_ucp.h
```diff
@@ -0,0 +1,48 @@
...
+
+typedef struct ucc_tl_ucp_lib {
+    ucc_tl_lib_t super;
+} ucc_tl_ucp_lib_t;
+UCC_CLASS_DECLARE(ucc_tl_ucp_lib_t, const ucc_base_lib_params_t *,
+                  const ucc_base_config_t *);
+
+typedef struct ucc_tl_ucp_context {
+    ucc_tl_context_t super;
+    ucp_address_t   *worker_address;
```
Sergei-Lebedev, line 37: Why do you want to store ucp_adress in context? Isn't it better to query for it on demand?

src/components/tl/ucc_tl_log.h
```diff
@@ -0,0 +1,25 @@
...
+#define UCC_TL_LOG_H_
+#include "utils/ucc_log.h"
+#define ucc_log_component_tl(_tl_ctx, _level, fmt, ...)                        \
+    ucc_log_component(_level, ((ucc_base_context_t *)_tl_ctx)->lib->log_component, fmt, \
+                      ##__VA_ARGS__)
+
+#define tl_error(_tl_ctx, _fmt, ...)                                           \
+    ucc_log_component_tl(_tl_ctx, UCS_LOG_LEVEL_ERROR, _fmt, ##__VA_ARGS__)
+#define tl_warn(_tl_ctx, _fmt, ...)                                            \
+    ucc_log_component_tl(_tl_ctx, UCS_LOG_LEVEL_WARN, _fmt, ##__VA_ARGS__)
```
Sergei-Lebedev, line 17: I think it should be UCC_LOG_LEVEL_WARN  instead

src/components/tl/ucp/tl_ucp_context.c
vspetrov: @manjugv  This one is tricky. UCC defines THREAD mode as LIB object attribute. So, all the checks by user are done before creating the context. In contrast, ucp threading mode is defined (queried) on the WORKER, which is only created during context creation. So, if at this point i will return OK, saying to upper layer (CL) that i can only support THREAD_SINGLE (even if ucc_lib is MT) we can no longer change the level of threading support that has been already reported to user via lib_attr. 

Thoughts ? Does this mean that we need to move UCC_THREAD_MODE to the context ?

src/components/tl/ucc_tl_log.h
vspetrov: yes

src/components/tl/ucp/tl_ucp.h
vspetrov: agree, no need

src/components/tl/ucp/tl_ucp_context.c
manjugv: If the library is created in single-thread mode - all teams and contexts are in single-threaded mode. If the library is multithreaded, then the context can be either single or multi-threaded. I think we need to expose thread mode in context, so we have finer control.

src/components/tl/ucp/tl_ucp.h
vspetrov: after discussion with Manju: will add short desc

src/components/tl/ucp/tl_ucp_context.c
vspetrov: After discussion we decided it is not right to change UCC iface only because UCP is implemented that way. Ideally we should be able to query UCP if it can support THREAD_MULTIPLE w/o allocation of ucp_worker. Then the issue above will be resolved. So far we keep the error handling that way: reporting error if TL can't support required thread level

src/components/cl/ucc_cl.h
```diff
@@ -12,6 +12,18 @@
...
+  * into a single UCC object (lib, context, team).
+  * The interface very closely matches that of ucc.h with just
+  * the flexibility for possible extension. The interface is used
+  * internally by the ucc/core.
+  * CL components are loaded dynamically and their name should
+  * match the predefined pattern "ucc_cl_<name>.so".
+  * Users may control the CL selection with UCC_CLS lib
+  * parameter.
+  */
+
```
manjugv, line 26: CL (collective layer) is an internal collective interface reflecting the public UCC API and extensions to support modularity, the composition of multiple collective implementations, and functionality that bridges the gap between hardware implementation of communication primitives and the programming models. 

The CL layer will build upon TL for the communication transport requirements. The CL can include a basic implementation, which provides minimal functionality over the TL, or can provide more optimized implementation such as hierarchical implementation that leverages multiple TL components. 

The different implementations of CL are realized as different CL components. The CL components are loaded dynamically, and their names should match the predefined pattern “ucc_cl_<name>.so”. The CL that is used for a given application invocation can be selected with the UCC_CLS lib parameter.

src/components/tl/ucc_tl.h
```diff
@@ -0,0 +1,72 @@
...
+  * TL interface is internal to UCC library and is used directly
+  * by different CL components.
+  * Different CLs may share TL context: e.g., CL_BASIC and CL_HIER
+  * will both use TL_UCP and share its context.
+  * A TL component may not use another TL (i.e. horizontal
+  * dependency is not allowed).
+  * The selection of TLs in runtime is defined by the usage at CL
+  * level.
+  */
+
```
manjugv, line 26: TL (transport layer) is an internal interface that provides a basic implementation of collectives and p2p primitives. It differs from CL in that it focuses on the abstraction over hardware transport rather than the programming model. These collectives can leverage either p2p communication transport or collective transport. The primary example of p2p communication transport is UCX, and the primary example of collective transport is SHARP, MCAST, and shared memory.

manjugv (approved):
What about the suggestions for comments ? Otherwise, the PR is ready to go.

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -0,0 +1,126 @@
...
+        goto err_worker_create;
+    }
+
+    if (params->thread_mode == UCC_THREAD_MULTIPLE) {
+        worker_attr.field_mask = UCP_WORKER_ATTR_FIELD_THREAD_MODE;
+        ucp_worker_query(ucp_worker, &worker_attr);
+        if (worker_attr.thread_mode != UCS_THREAD_MODE_MULTI) {
+            tl_error(&self->super,
+                     "thread mode multiple is not supported by ucp worker");
+            status = UCC_ERR_NOT_SUPPORTED;
```
Sergei-Lebedev, line 101: should be ucc_status instead of status

Sergei-Lebedev (approved):

src/components/tl/ucc_tl.h
vspetrov: applied

src/components/cl/ucc_cl.h
vspetrov: applied

PR #48, opened 2020-12-16
Resource Sharing API

manjugv:
## What
Enables lazy resource association for UCC objects

## Why ?
Resource sharing is important for scalability and particularly important for systems with high core and thread count. 

## How ?
 This PR provides interfaces to lazily associate resources to UCC objects created/initialized by the third-party libraries. It allows association at various granularities including for teams, contexts, and libraries. The resources can be disassociated when the resources are finalized or destroyed by the third-party libraries.

### DIFF
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -626,15 +626,15 @@ typedef struct ucc_resource_params {
  *  @parblock
  *
  *  @b Description A local resource association operation for the UCC objects. The parameters
-+ *  specify the UCC objects to be associated with the resources. The resources are provided as
-+ *  a key/value/length triplet parameters. The values are user and implementation defined.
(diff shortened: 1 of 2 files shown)


alex--m (changes requested):
In general I'm not convinced with the underlying "lazy creation" approach. I think it makes for senseless stub creations, and it is really unclear at which point you actually finalize the creation of a context or a team. You never know - maybe half-way throughout the run someone will add a UCP worker? I strongly feel we should require all existing objects to be passed before creation.

On a related note, I put in the time to listen and review this proposal - I expect you extend the same courtesy to PR #40. I was posted before this one, and your choice to ignore it and present yours instead doesn't make #40 depend on it. @manjugv

vspetrov:
@manjugv what is the status here? do we need to keep it or can we close/move to issue ?

manjugv:
Closing for now. Need to update it anyways as we have a master branch that has progressed further if we pursue this approach.

PR #51, opened 2021-01-11
CORE: redefine ucc class functions

Sergei-Lebedev:
## What
Redefine UCC_CLASS_INIT_FUNC and UCC_CLASS_DECLARE 

## Why ?
Original versions of these macros in UCS define ucs_status_t as return code. It is not consistent with CL and TL implementation in which we return ucc_status_t

### DIFF
--- a/src/utils/ucc_class.h
+++ b/src/utils/ucc_class.h
@@ -10,13 +10,20 @@
 #include <ucs/type/class.h>
 #include "utils/ucc_compiler_def.h"
 
-#define UCC_CLASS_DECLARE         UCS_CLASS_DECLARE
 #define UCC_CLASS_DEFINE          UCS_CLASS_DEFINE
 #define UCC_CLASS_DELETE          UCS_CLASS_DELETE
-#define UCC_CLASS_INIT_FUNC       UCS_CLASS_INIT_FUNC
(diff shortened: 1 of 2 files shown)


src/utils/ucc_class.h
```diff
@@ -10,14 +10,21 @@
...
 #define UCC_CLASS_CLEANUP         UCS_CLASS_CLEANUP
 #define UCC_CLASS_NEW_FUNC_NAME   UCS_CLASS_NEW_FUNC_NAME
 
+#define UCC_CLASS_INIT_FUNC(_type, ...)                                        \
+    ucc_status_t _UCS_CLASS_INIT_NAME(_type)(_type *self,                      \
+                                             ucs_class_t *_myclass,            \
+                                             int *_init_count, ## __VA_ARGS__) \
+
+#define UCC_CLASS_DECLARE(_type, ...)                                          \
+    extern ucs_class_t _UCS_CLASS_DECL_NAME(_type);                            \
```
vspetrov, line 25: why ucs_class_t here? shouldn't it be ucc_class_t ?

vspetrov (approved):

manjugv (approved):

src/utils/ucc_class.h
Sergei-Lebedev: decided to leave it as it is

PR #52, opened 2021-01-12
CORE: init TL contexts

vspetrov:
## What
Adds TL context creation/destruction logic

## Why ?
Core functionality

## How ?
Everything is handled by CORE: 
- core ucc_lib/context queries CLs which TLs are required
- CORE allocates necessary TL libs/contexts
- CLs can then get a pointer to TL context based on type enum value

This is one of the alternative approaches (#53 ) and it is more clear, logical and convenient imo.

### DIFF
--- a/src/components/cl/basic/cl_basic_context.c
+++ b/src/components/cl/basic/cl_basic_context.c
@@ -11,9 +11,16 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_context_t,
                     const ucc_base_context_params_t *params,
                     const ucc_base_config_t *config)
 {
+    ucc_status_t status;
     const ucc_cl_context_config_t *cl_config =
         ucc_derived_of(config, ucc_cl_context_config_t);
     UCC_CLASS_CALL_SUPER_INIT(ucc_cl_context_t, cl_config->cl_lib);
+    status = ucc_tl_context_get(params->context, UCC_TL_UCP,
+                                &self->tl_ucp_ctx);
+    if (UCC_OK != status) {
+        cl_warn(cl_config->cl_lib, "TL UCP context is not available, CL BASIC can't proceed");
+        return UCC_ERR_NOT_FOUND;
+    }
     cl_info(cl_config->cl_lib, "initialized cl context: %p", self);
     return UCC_OK;
 }
@@ -21,6 +28,13 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_context_t,
 UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_context_t)
 {
     cl_info(self->super.super.lib, "finalizing cl context: %p", self);
+    ucc_tl_context_put(self->tl_ucp_ctx);
 }
 
 UCC_CLASS_DEFINE(ucc_cl_basic_context_t, ucc_cl_context_t);
+
+ucc_status_t ucc_cl_basic_get_context_attr(const ucc_base_context_t *context,
+                                           ucc_base_attr_t *attr)
+{
+    return UCC_OK;
+}
--- a/src/components/cl/ucc_cl.h
+++ b/src/components/cl/ucc_cl.h
@@ -80,8 +80,11 @@ typedef struct ucc_cl_context {
 
 UCC_CLASS_DECLARE(ucc_cl_context_t, ucc_cl_lib_t *);
 
-#define UCC_CL_IFACE_EXT(_NAME) .super.type = UCC_CL_ ## _NAME,
+typedef struct ucc_cl_lib_attr {
+    ucc_base_attr_t super;
+    uint64_t tls;
+} ucc_cl_lib_attr_t;
 
 #define UCC_CL_IFACE_DECLARE(_name, _NAME)                                     \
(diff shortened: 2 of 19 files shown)


src/components/cl/basic/cl_basic_context.c
```diff
@@ -11,16 +11,30 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_context_t,
...
+    ucc_tl_context_put(self->tl_ucp_ctx);
 }
 
 UCC_CLASS_DEFINE(ucc_cl_basic_context_t, ucc_cl_context_t);
+
+ucc_status_t ucc_cl_basic_get_context_attr(const ucc_base_context_t *context,
+                                           ucc_base_attr_t *attr)
+{
+    return UCC_OK;
+}
```
manjugv, line 40: I’m assuming this will be filled in future PRs.

src/components/cl/ucc_cl.h
```diff
@@ -80,8 +80,11 @@ typedef struct ucc_cl_context {
 
 UCC_CLASS_DECLARE(ucc_cl_context_t, ucc_cl_lib_t *);
 
-#define UCC_CL_IFACE_EXT(_NAME) .super.type = UCC_CL_ ## _NAME,
+typedef struct ucc_cl_lib_attr {
+    ucc_base_attr_t super;
+    int tls;
```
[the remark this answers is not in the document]
manjugv: How do you set the required TLs? Does the CL and TLs have some kind of identifiers?

src/core/ucc_context.c
```diff
@@ -226,6 +275,14 @@ ucc_status_t ucc_context_create(ucc_lib_h lib,
     ctx->lib = lib;
     ucc_copy_context_params(&ctx->params, params);
     ucc_copy_context_params(&b_params.params, params);
+    b_params.context = ctx;
+    status = ucc_create_tl_contexts(ctx, config);
```
manjugv, line 279: I’m wondering in this approach, where you initialize multiple CLs but you end up using only one CL, this will result in creating contexts for all TLs? Would wrapping TL under CL help overcome this? 
Did you explore that approach? Any obvious disadvantages?

manjugv (commented):
* Besides other comments, is there an opportunity to say not to open certain components? Like in Open MPI, where you could say not to open certain BTLs. 

* Probably for the files you touched, you need to update the copyright to reflect the year change.

src/components/cl/ucc_cl.h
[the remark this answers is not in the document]
vspetrov: yes CLs and TLs have both have enums. Some identifier is needed anyways. Originally we thought that for TLs we can use "char *". But since any TL (even vendor specific black box) requires integration into CL then there is no problem adding new enum id for any TL. And this way it is much more convenient.

src/components/cl/basic/cl_basic_context.c
vspetrov: > * Besides other comments, is there an opportunity to say not to open certain components? Like in Open MPI, where you could say not to open certain BTLs.
> * Probably for the files you touched, you need to update the copyright to reflect the year change.

Agree. We can add such a parameter to TL (to filter TLs based on user input).

src/core/ucc_context.c
vspetrov: yes, i've considered that approach: see closed/rejected PR #53 - we went over that approach with Sergey. Shortly, the downsides are the complexity of implementation and maintenance (need to keep hash-table based storage for tl lib and context objects in order to organize TL ctx sharing between CLs) and somewhat more complex cleanup flow. These complexities could be fine if we would get some real benefit from that approach, but looks like we actually don't. 

Regarding you example: we first filter CL lib objects based on user input, only after that we build the list of required TLs. So, the only corner case is : when some CL is supposed to be used but it fails creating its context AND it requested some TL which is not used by other CLs. In that case there is some TL context which is created w/o the need (so we will need to destroy it right away - this is TODO). However, this corner case still exist even in that alternative approach: consider some CL (say CL_HW - that implements a HW collectives) that needs a specific TL (say  sharp) and this CL allocates TL context internally when cl_context is being initialized. If the cl_context create fails for some reason then that CL has to cleanup allocated TL anyways. So, it is all the same but happening in different place.

Lastly, third argument: there is almost no reason for CL context to fail during init except for TL context init failure (this is where resources are allocated) unless some critical error (like OOM). 

So, i believe the approach taken in this PR is cleaner and has basically same corner cases as alternative approach.

vspetrov:
@manjugv @Sergei-Lebedev @alex--m please leave your comments, lets keep those PRs moving

manjugv:
> @manjugv @Sergei-Lebedev @alex--m please leave your comments, lets keep those PRs moving

@vspetrov I don't have any other comments on this PR.

Sergei-Lebedev (approved):
looks good to me

vspetrov:
> > @manjugv @Sergei-Lebedev @alex--m please leave your comments, lets keep those PRs moving
> 
> @vspetrov I don't have any other comments on this PR.

@manjugv if you don't have comments and you are good with the answers to others, then plz put +1 for the PR. @alex--m plz respond

manjugv (approved):

PR #53, opened 2021-01-12
CORE: init TL contexts

vspetrov:
Serves same purpose as #52 but with alternative implementation:

CORE knows nothing about TLs. CLs directly get TL context which is allocated internally and stored into a hash table.
Downsides (to discuss):
- Hash tables little bit messy
- Complicated TL cleanup logic (not implemented in this PR)
- how to implement local address packing function? - it is in CORE but CORE is not aware of TLs ..


vspetrov:
After discussion, closing in favor of #52

PR #54, opened 2021-01-18
Team create

vspetrov:
## What
Adds base team_create_post/test/destroy interface; cl and tl implementation; TL UCP non blocking team post/test implementation; gtest multiprocess env simulation + team_create_destroy test

### DIFF
--- a/src/components/cl/basic/cl_basic_team.c
+++ b/src/components/cl/basic/cl_basic_team.c
@@ -29,13 +29,27 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_team_t, ucc_base_context_t *cl_context,
 
 UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_team_t)
 {
-    ucc_cl_basic_context_t *ctx = UCC_CL_BASIC_TEAM_CTX(self);
-
-    if (self->tl_ucp_team) {
-        UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
-            ->team.destroy(&self->tl_ucp_team->super);
-    }
     cl_info(self->super.super.context->lib, "finalizing cl team: %p", self);
+}
+
+UCC_CLASS_DEFINE_DELETE_FUNC(ucc_cl_basic_team_t, ucc_base_team_t);
+UCC_CLASS_DEFINE(ucc_cl_basic_team_t, ucc_cl_team_t);
+
+ucc_status_t ucc_cl_basic_team_destroy(ucc_base_team_t *cl_team)
+{
+    ucc_cl_basic_team_t    *team = ucc_derived_of(cl_team, ucc_cl_basic_team_t);
+    ucc_cl_basic_context_t *ctx  = UCC_CL_BASIC_TEAM_CTX(team);
+    ucc_status_t            status;
+    if (team->tl_ucp_team) {
+        if (UCC_OK !=
+            (status = UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
+                          ->team.destroy(&team->tl_ucp_team->super))) {
+            return status;
+        }
+        team->tl_ucp_team = NULL;
+    }
+    UCC_CLASS_DELETE_FUNC_NAME(ucc_cl_basic_team_t)(cl_team);
+    return UCC_OK;
 }
 
 ucc_status_t ucc_cl_basic_team_create_test(ucc_base_team_t *cl_team)
@@ -52,5 +66,3 @@ ucc_status_t ucc_cl_basic_team_create_test(ucc_base_team_t *cl_team)
     }
     return status;
 }
-
-UCC_CLASS_DEFINE(ucc_cl_basic_team_t, ucc_cl_team_t);
--- a/src/components/cl/basic/cl_basic_team.c
+++ b/src/components/cl/basic/cl_basic_team.c
@@ -12,20 +12,45 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_team_t, ucc_base_context_t *cl_context,
 {
     ucc_cl_basic_context_t *ctx =
         ucc_derived_of(cl_context, ucc_cl_basic_context_t);
+    ucc_status_t     status;
+    ucc_base_team_t *b_team;
     UCC_CLASS_CALL_SUPER_INIT(ucc_cl_team_t, &ctx->super);
-    cl_info(cl_context->lib, "initialized cl team: %p", self);
-    return UCC_OK;
+    status = UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
+                 ->team.create_post(&ctx->tl_ucp_ctx->super, params, &b_team);
+    if (UCC_OK == status) {
+        self->tl_ucp_team = ucc_derived_of(b_team, ucc_tl_team_t);
+        cl_info(cl_context->lib, "posted cl team: %p", self);
+    } else {
+        self->tl_ucp_team = NULL;
+        cl_error(cl_context->lib, "tl ucp team create post failed");
+    }
+    return status;
 }
 
 UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_team_t)
 {
+    ucc_cl_basic_context_t *ctx = UCC_CL_BASIC_TEAM_CTX(self);
+
+    if (self->tl_ucp_team) {
+        UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
+            ->team.destroy(&self->tl_ucp_team->super);
+    }
     cl_info(self->super.super.context->lib, "finalizing cl team: %p", self);
 }
 
 ucc_status_t ucc_cl_basic_team_create_test(ucc_base_team_t *cl_team)
 {
-    /* ucc_cl_basic_team_t *team = ucc_derived_of(cl_team, ucc_cl_basic_team_t); */
-    return UCC_OK;
+    ucc_cl_basic_team_t    *team = ucc_derived_of(cl_team, ucc_cl_basic_team_t);
+    ucc_cl_basic_context_t *ctx  = UCC_CL_BASIC_TEAM_CTX(team);
+    ucc_status_t            status;
+    status = UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
+                 ->team.create_test(&team->tl_ucp_team->super);
+    if (UCC_OK == status) {
+        cl_info(ctx->super.super.lib, "initialized cl team: %p", team);
+    } else if (UCC_INPROGRESS != status) {
+        cl_error(ctx->super.super.lib, "failed to create tl ucp team");
+    }
+    return status;
 }
 
 UCC_CLASS_DEFINE(ucc_cl_basic_team_t, ucc_cl_team_t);
--- a/src/components/cl/basic/cl_basic_team.c
+++ b/src/components/cl/basic/cl_basic_team.c
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ * Copyright (C) Mellanox Technologies Ltd. 2020-2021.  ALL RIGHTS RESERVED.
  *
  * See file LICENSE for terms.
  */
@@ -17,14 +17,14 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_team_t, ucc_base_context_t *cl_context,
     UCC_CLASS_CALL_SUPER_INIT(ucc_cl_team_t, &ctx->super);
     status = UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
                  ->team.create_post(&ctx->tl_ucp_ctx->super, params, &b_team);
-    if (UCC_OK == status) {
-        self->tl_ucp_team = ucc_derived_of(b_team, ucc_tl_team_t);
-        cl_info(cl_context->lib, "posted cl team: %p", self);
-    } else {
+    if (UCC_OK != status) {
         self->tl_ucp_team = NULL;
         cl_error(cl_context->lib, "tl ucp team create post failed");
+        return status;
     }
-    return status;
+    self->tl_ucp_team = ucc_derived_of(b_team, ucc_tl_team_t);
+    cl_info(cl_context->lib, "posted cl team: %p", self);
+    return UCC_OK;
 }
 
 UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_team_t)
--- a/src/components/cl/basic/cl_basic_team.c
+++ b/src/components/cl/basic/cl_basic_team.c
@@ -1,0 +1,31 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "cl_basic.h"
+#include "utils/ucc_malloc.h"
+
+UCC_CLASS_INIT_FUNC(ucc_cl_basic_team_t, ucc_base_context_t *cl_context,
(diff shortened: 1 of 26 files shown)


manjugv:
@vspetrov Is this ready for review?

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -0,0 +1,243 @@
...
+    if (!st) {
+        tl_error(ctx->super.super.lib,
+                 "failed to allocate %zd bytes for tl_ucp_addr_storage",
+                 sizeof(*st));
+        return UCC_ERR_NO_MEMORY;
+    }
+    if ((NULL == ctx->worker_address) &&
+        (UCS_OK != ucp_worker_get_address(ctx->ucp_worker, &ctx->worker_address,
+                                          &ctx->ucp_addrlen))) {
+        tl_error(ctx->super.super.lib, "failed to get ucp worker address");
```
Sergei-Lebedev, line 32: add free(st) or goto err_addrlens

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -0,0 +1,243 @@
...
+    int                  i;
+    ucc_team_oob_coll_t *oob = &storage->oob;
+    if (storage->state == UCC_TL_UCP_ADDR_EXCHANGE_COMPLETE) {
+        return UCC_OK;
+    }
+    status = oob->req_test(storage->oob_req);
+    if (UCC_INPROGRESS == status) {
+        return status;
+    } else if (UCC_OK != status) {
+        tl_error(storage->ctx->super.super.lib, "failed during oob req test");
```
Sergei-Lebedev, line 77: free oob request

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -0,0 +1,243 @@
...
+static ucc_status_t ucc_tl_ucp_team_preconnect(ucc_tl_ucp_team_t *team)
+{
+    ucc_tl_ucp_context_t *ctx = UCC_TL_UCP_TEAM_CTX(team);
+    int                   i;
+    ucc_status_t          status;
+    for (i = 0; i < team->size; i++) {
+        status = ucc_tl_ucp_connect_ep(ctx, team, team->addr_storage->addresses,
+                                       team->addr_storage->max_addrlen, i);
+        if (UCC_OK != status) {
+            return status;
```
Sergei-Lebedev, line 199: Do we need to close all already opened endpoints in case of failure?

test/gtest/core/test_multiproc.cc
```diff
@@ -0,0 +1,203 @@
...
+    case test_multiproc::AG_COMPLETE:
+        return UCC_OK;
+    default:
+        break;
+    }
+    return UCC_INPROGRESS;
+}
+
+ucc_status_t req_free(void *request)
+{
```
Sergei-Lebedev, line 111: Don't we need to reset AG state to AG_INIT?

src/components/cl/basic/cl_basic_team.c
```diff
@@ -0,0 +1,68 @@
...
+    status = UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
+                 ->team.create_post(&ctx->tl_ucp_ctx->super, params, &b_team);
+    if (UCC_OK == status) {
+        self->tl_ucp_team = ucc_derived_of(b_team, ucc_tl_team_t);
+        cl_info(cl_context->lib, "posted cl team: %p", self);
+    } else {
+        self->tl_ucp_team = NULL;
+        cl_error(cl_context->lib, "tl ucp team create post failed");
+    }
+    return status;
```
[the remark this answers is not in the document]
vspetrov: 👍

src/components/tl/ucp/tl_ucp.c
```diff
@@ -20,12 +20,13 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
     {"", "", NULL, ucc_offsetof(ucc_tl_ucp_context_config_t, super),
      UCC_CONFIG_TYPE_TABLE(ucc_tl_context_config_table)},
 
-    {"TEST_PARAM", "5", "For dbg test purpuse : don't commit",
-     ucc_offsetof(ucc_tl_ucp_context_config_t, test_param),
+    {"PRECONNECT", "1",
```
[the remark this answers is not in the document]
vspetrov: in hcoll we have this as INT and the value actually defines the size of the team when you would the "preconnect". For example, hcoll's default is 1024, meaning up to 1024 procs we do preconnect, for larger scale, we do on-demand. Wdyt?

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -0,0 +1,243 @@
...
+#include "tl_ucp.h"
+#include "tl_ucp_ep.h"
+#include "utils/ucc_malloc.h"
+enum {
+    UCC_TL_UCP_ADDR_EXCHANGE_MAX_ADDRLEN,
+    UCC_TL_UCP_ADDR_EXCHANGE_GATHER,
+    UCC_TL_UCP_ADDR_EXCHANGE_COMPLETE,
+};
+
+ucc_status_t ucc_tl_ucp_addr_exchange_start(ucc_tl_ucp_context_t *ctx,
```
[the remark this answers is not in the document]
vspetrov: addr exchange will also be used in the context create (if some conditions are met: oob for context is provided, context rank is provided, etc). So i made those funcs non-static on purpose. I'll move them to a separate .c module for clarity.

src/components/tl/ucp/tl_ucp_team.c
vspetrov: 👍

src/components/tl/ucp/tl_ucp_team.c
vspetrov: i think we should. 👍

test/gtest/core/test_multiproc.cc
vspetrov: great catch 👍

src/components/tl/ucp/tl_ucp.c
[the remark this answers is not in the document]
Sergei-Lebedev: I see, yes it sounds reasonable

src/components/cl/basic/cl_basic_team.c
```diff
@@ -0,0 +1,68 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
```
manjugv, line 2: 2021

src/components/tl/ucp/Makefile.am
```diff
@@ -6,7 +6,10 @@ sources =            \
 	tl_ucp.h         \
 	tl_ucp.c         \
 	tl_ucp_lib.c     \
-	tl_ucp_context.c
+	tl_ucp_context.c \
+	tl_ucp_team.c    \
+	tl_ucp_ep.h      \
+    tl_ucp_ep.c
```
manjugv, line 12: indent

src/components/tl/ucp/tl_ucp.c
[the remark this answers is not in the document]
manjugv: I'm assuming that the value can be configurable for the user.

src/components/tl/ucp/tl_ucp.h
```diff
@@ -23,7 +24,7 @@ typedef struct ucc_tl_ucp_lib_config {
 
 typedef struct ucc_tl_ucp_context_config {
     ucc_tl_context_config_t super;
-    int                     test_param;
+    int                     preconnect;
```
manjugv, line 27: I'm not sure what is the role of pre-connect here: flag or storing the threshold value? If it is the threshold value, I think preconnect should be part of the team rather than the context. The context can be used for any team size, and if your intent to wire based on the team size, then it makes sense that it is a part of the team structure.

src/components/tl/ucp/tl_ucp.h
```diff
@@ -32,16 +33,51 @@ typedef struct ucc_tl_ucp_lib {
...
+    int              preconnect;
+    ucc_tl_ucp_ep_close_state_t ep_close_state;
 } ucc_tl_ucp_context_t;
 UCC_CLASS_DECLARE(ucc_tl_ucp_context_t, const ucc_base_context_params_t *,
                   const ucc_base_config_t *);
 
+typedef struct ucc_tl_ucp_team {
+    ucc_tl_team_t super;
+    ucc_status_t  status;
+    int           context_ep_storage; /*< The flag indicates whether ucp endpoints
```
manjugv, line 52: indent

src/components/tl/ucp/tl_ucp.h
manjugv: Also, do you want it is as uint32/64 - similar to team size?

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -0,0 +1,74 @@
...
+        ep = &team->eps[rank];
+    }
+    if (*ep) {
+        /* Already connected */
+        return UCC_OK;
+    }
+    ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
+    ep_params.address    = address;
+
+    status = ucp_ep_create(ctx->ucp_worker, &ep_params, ep);
```
manjugv, line 26: Nitpick: Does create ep connect (exchange information) the endpoints or just create? If only creates should we call it create?

src/components/tl/ucp/tl_ucp_team.c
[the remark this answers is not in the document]
manjugv: Agree. It makes sense to move it to a separate file.

src/core/ucc_team.c
```diff
@@ -0,0 +1,216 @@
...
+    team->num_contexts = num_contexts;
+    team->contexts =
+        ucc_malloc(sizeof(ucc_context_t *) * num_contexts, "ucc_team_ctx");
+    if (!team->contexts) {
+        ucc_error("failed to allocate %zd bytes for ucc team contexts array",
+                  sizeof(ucc_context_t) * num_contexts);
+        status = UCC_ERR_NO_MEMORY;
+        goto err_ctx_alloc;
+    }
+    memcpy(team->contexts, contexts, sizeof(ucc_context_t *) * num_contexts);
```
manjugv, line 109: Though not performance critical, I was wondering why not use ucs_memcpy* ?

test/gtest/core/test_multiproc.cc
```diff
@@ -0,0 +1,206 @@
...
+    ucc_lib_config_release(lib_config);
+}
+
+void test_process::init_ctx()
+{
+    EXPECT_EQ(UCC_OK, ucc_context_config_read(lib_h, NULL, &ctx_config));
+    ctx_params.mask     = UCC_CONTEXT_PARAM_FIELD_TYPE;
+    ctx_params.ctx_type = UCC_CONTEXT_EXCLUSIVE;
+    EXPECT_EQ(UCC_OK,
+              ucc_context_create(lib_h, &ctx_params, ctx_config, &ctx_h));
```
manjugv, line 24: Can we move resource creation to common code? As we add more tests, we don’t have to repeat (for example here context creation) the resource creation for each new interface. This might help gtest being relatively faster in the future.

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -0,0 +1,74 @@
...
+
+ucc_status_t ucc_tl_ucp_connect_ep(ucc_tl_ucp_context_t *ctx,
+                                   ucc_tl_ucp_team_t *team, char *addr_array,
+                                   size_t max_addrlen, int rank)
+{
+    ucp_address_t  *address = (ucp_address_t *)(addr_array + max_addrlen * rank);
+    ucp_ep_params_t ep_params;
+    ucs_status_t    status;
+    ucp_ep_h       *ep;
+    if (team->context_ep_storage) {
```
manjugv, line 12: FYI, this would be applicable only in the case of shared context, and you might want to check it.

manjugv (commented):
Overall, it looks good to me. There are few minor changes and some clarification questions.

manjugv:
@alex--m Ping ..

src/components/tl/ucp/tl_ucp.c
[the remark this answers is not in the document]
vspetrov: yes, it is configurable by user with UCC_TL_UCP_PRECONNECT env var

src/components/tl/ucp/tl_ucp_team.c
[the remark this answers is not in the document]
vspetrov: done

src/components/tl/ucp/tl_ucp_team.c
vspetrov: done

src/components/tl/ucp/tl_ucp.h
vspetrov: I've update the PR. Preconnect is now a threshold. Threshold that defines the number of ranks in the UCC team/context below which the team/context enpoints will be preconnected during  corresponding team/context create call. We can only store the parameter on the  context. Firstly we don't have interface to set parameters per team. And since user anyways can only set one parameter value for a job then it is constant for all teams anyways. The actual behavior will depend on other parameters and settings. Basically, if UCP endpoints are stored on a team then preconnect applies to the team. If eps are stored on a context then the threshold will define if the whole tl ucp context gets preconnected

src/core/ucc_team.c
vspetrov: why to use it ? do we want to overwrite default memcpy with ucc_memcpy throughout the code? we can

test/gtest/core/test_multiproc.cc
vspetrov: It kind of IS in the common code right now. There is a "test_process" class that has the corresponding members. Whenever one wants to add another test with "multiproc" simulation env it can derive from it.

src/components/tl/ucp/tl_ucp_ep.c
vspetrov: This is correct. I have this as TODO at line 20 in tl_ucp_team.c. There i initialize context_ep_storage either to 1 or 0. The thing is : the check is more complicated then just context being shared or not. In order to user context ep storage we need the context to be "collective" (ie oob provided), we need the context to have defined EP with EP_RANGE_CONTIG (ie context level rank, or kind of world_rank) and we need the mappers from team_rank->context_rank provided at the team creation time. I plan to add this logic later because i think we might need some more work there (need to check all the cases when different mappings are provided and when not, and how to handle latter cases)

src/core/ucc_team.c
manjugv: In general for performance portability across the architectures and in this case for uniformity. As I understand, folks are always optimizing the performance in UCS.

src/components/tl/ucp/tl_ucp_ep.c
vspetrov: it is actually connect

src/core/ucc_team.c
vspetrov: so, Manju, basically ucs does not have ucs_memcpy. There are 2 flavours of memcpy there: ucs_memcpy_nontemporal and ucs_memcpy_relaxed with arch specific implementations. And UCX itself use those only in a couple of places. All the rest of UCX code uses just memcpy - which is way more portable actually (libc on pretty much any system has memcpy). So, i don't think we should use some exotic memcpy in the places like above - out of critical path, generic copy w/o any specific semantics.

manjugv (approved):

Sergei-Lebedev (approved):

alex--m (approved):

PR #55, opened 2021-01-19
CORE: added memory component interface

Sergei-Lebedev:
## What
Memory components (MC) interface to work with different kinds of memory

## Why ?
Provides functionality for CLs and TLs to allocate memory and perform reduce for predefined datatypes and operations.

## How ?
Each MC is loadable object to avoid dependency on third party libraries (e.g. libcudart). There is exactly one instance of MC per library. Functions in core provides proxy to access each MC.

### DIFF
--- a/src/components/mc/base/ucc_mc_base.c
+++ b/src/components/mc/base/ucc_mc_base.c
@@ -1,0 +1,25 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2020-2021.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "ucc_mc_base.h"
+
+ucc_config_field_t ucc_mc_config_table[] = {
+    {"LOG_LEVEL", "warn",
+     "UCC logging level. Messages with a level higher or equal to the "
+     "selected will be printed.\n"
+     "Possible values are: fatal, error, warn, info, debug, trace, data, func, "
+     "poll.",
+     ucc_offsetof(ucc_mc_config_t, log_component), UCC_CONFIG_TYPE_LOG_COMP},
+
+    {NULL}};
+
+const char *ucc_memory_type_names[] = {
+    [UCC_MEMORY_TYPE_HOST]         = "host",
+    [UCC_MEMORY_TYPE_CUDA]         = "cuda",
+    [UCC_MEMORY_TYPE_CUDA_MANAGED] = "cuda-managed",
+    [UCC_MEMORY_TYPE_ROCM]         = "rocm",
+    [UCC_MEMORY_TYPE_ROCM_MANAGED] = "rocm-managed",
+    [UCC_MEMORY_TYPE_LAST]         = "unknown"};
--- a/src/components/mc/base/ucc_mc_base.h
+++ b/src/components/mc/base/ucc_mc_base.h
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2020-2021.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#ifndef UCC_MC_BASE_H_
+#define UCC_MC_BASE_H_
+
+#include "config.h"
+#include "utils/ucc_component.h"
+#include "utils/ucc_class.h"
+#include "utils/ucc_parser.h"
+#include "core/ucc_global_opts.h"
+
+typedef enum ucc_memory_type {
+    UCC_MEMORY_TYPE_HOST,         /**< Default system memory */
+    UCC_MEMORY_TYPE_CUDA,         /**< NVIDIA CUDA memory */
+    UCC_MEMORY_TYPE_CUDA_MANAGED, /**< NVIDIA CUDA managed memory */
+    UCC_MEMORY_TYPE_ROCM,         /**< AMD ROCM memory */
+    UCC_MEMORY_TYPE_ROCM_MANAGED, /**< AMD ROCM managed system memory */
+    UCC_MEMORY_TYPE_LAST,
+    UCC_MEMORY_TYPE_UNKNOWN = UCC_MEMORY_TYPE_LAST
+} ucc_memory_type_t;
+
+/**
+ * Array of string names for each memory type
+ */
+extern const char *ucc_memory_type_names[];
+
+typedef struct ucc_mc_config {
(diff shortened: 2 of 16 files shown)


src/core/ucc_mc.c
```diff
@@ -0,0 +1,135 @@
...
+    int            i, n_mcs;
+    ucc_mc_base_t *mc;
+    ucc_status_t   status;
+
+    memset(mc_ops, 0, UCC_MEMORY_TYPE_LAST * sizeof(ucc_mc_ops_t *));
+    n_mcs = ucc_global_config.mc_framework.n_components;
+    for (i = 0; i < n_mcs; i++) {
+        mc = ucc_derived_of(ucc_global_config.mc_framework.components[i],
+                            ucc_mc_base_t);
+        if (mc->ref_cnt == 0) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: I thought about possible future use case when user can ask to explicitly enable/disable specific memory component per lib. For example create one ucc_lib with only cpu mc and another with cpu and gpu mc. Does it make sense?

src/core/ucc_mc.c
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+    return UCC_OK;
+}
+
+ucc_status_t ucc_mc_alloc(void **ptr, size_t size, ucc_memory_type_t mem_type)
+{
+    ucc_status_t status = UCC_OK;
+
+    if (NULL == mc_ops[mem_type]) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: I think it is possible, e.g. tl sharp might have a pool of preregistered  host and gpu buffers. And if cuda is not available ucc_mc_alloc will return NOT_SUPPORTED. Alternatively we can add ucc_mc_is_available(ucc_memory_type_t mt) to check if given mc is there

src/components/mc/base/ucc_mc_base.c
```diff
@@ -0,0 +1,25 @@
...
+
+    {NULL}};
+
+const char *ucc_memory_type_names[] = {
+    [UCC_MEMORY_TYPE_HOST]         = "host",
+    [UCC_MEMORY_TYPE_CUDA]         = "cuda",
+    [UCC_MEMORY_TYPE_CUDA_MANAGED] = "cuda-managed",
+    [UCC_MEMORY_TYPE_ROCM]         = "rocm",
+    [UCC_MEMORY_TYPE_ROCM_MANAGED] = "rocm-managed",
+    [UCC_MEMORY_TYPE_LAST]         = "unknown"};
```
manjugv, line 25: We can add the types as we had the components. 

 In the future, you might have x86, aarch64, and DPUs. So for the host type, you might want a more specific identifier.

src/components/mc/base/ucc_mc_base.h
```diff
@@ -0,0 +1,56 @@
...
+
+typedef enum ucc_memory_type {
+    UCC_MEMORY_TYPE_HOST,         /**< Default system memory */
+    UCC_MEMORY_TYPE_CUDA,         /**< NVIDIA CUDA memory */
+    UCC_MEMORY_TYPE_CUDA_MANAGED, /**< NVIDIA CUDA managed memory */
+    UCC_MEMORY_TYPE_ROCM,         /**< AMD ROCM memory */
+    UCC_MEMORY_TYPE_ROCM_MANAGED, /**< AMD ROCM managed system memory */
+    UCC_MEMORY_TYPE_LAST,
+    UCC_MEMORY_TYPE_UNKNOWN = UCC_MEMORY_TYPE_LAST
+} ucc_memory_type_t;
```
manjugv, line 24: Same as above, we can add as we add the components

manjugv (commented):
As discussed at the WG, if we want to support copy and flush routines in the future, we might want to rethink the code placement and/or rename.

src/components/mc/base/ucc_mc_base.c
Sergei-Lebedev: Do you mean rename cpu mem component right now in this pull request?

test/gtest/core/test_mc.cc
```diff
@@ -0,0 +1,41 @@
...
+
+UCC_TEST_F(test_mc, init_is_required)
+{
+    void *ptr;
+
+    EXPECT_NE(UCC_OK, ucc_mc_alloc(&ptr, 1, UCC_MEMORY_TYPE_HOST));
+    EXPECT_NE(UCC_OK, ucc_mc_free(ptr, UCC_MEMORY_TYPE_HOST));
+}
+
+UCC_TEST_F(test_mc, can_alloc_and_free_host_mem)
```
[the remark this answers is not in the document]
Sergei-Lebedev: added this test and immediately found a bug

src/core/ucc_mc.c
```diff
@@ -0,0 +1,135 @@
...
+
+ucc_status_t ucc_mc_finalize()
+{
+   ucc_memory_type_t  mt;
+   ucc_mc_base_t     *mc;
+
+    for (mt = UCC_MEMORY_TYPE_HOST; mt < UCC_MEMORY_TYPE_LAST; mt++) {
+        if (NULL != mc_ops[mt]) {
+            mc = ucc_container_of(mc_ops[mt], ucc_mc_base_t, ops);
+            mc->ref_cnt--;
```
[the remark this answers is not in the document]
Sergei-Lebedev: discussed offline, will proceed with refcnt per component to have support for config option to choose mem component per library

src/components/mc/base/ucc_mc_base.c
manjugv: IMO, it might be easier now than later. Is there a reason not to do it now?

manjugv:
@alex--m Do you have any comments on this PR?

vspetrov (approved):

src/components/mc/base/ucc_mc_base.c
Sergei-Lebedev: One reason why I'm not sure about renaming cpu mc to x86-cpu mc is that it's not entirely clear how we will implement architecture specific code. One approach is to keep everything cpu specific in cpu mc so that it will be able to detect and use appropriate implementation. In that case UCC_MEMORY_TYPE_HOST means not specific architecture but runtime system memory. Another approach is to divide cpu mc into separate memory components but in that case we will have a lot of code duplication.

manjugv (approved):

PR #59, opened 2021-01-25
Collective init

vspetrov:
## What
Implements ucc_collective_init/post/test/finalize

## Why ?
Core API

## How ?
Firstly, the ucc_coll_req_t is defined in ucc_def.h to have ucc_status_t status field. All the internally allocated requests (during collective_init) will inherit from ucc_coll_req_t. Then ucc_collective_test is just inline "status == UCC_OK".
Second: we define ucc_coll_task_t and ucc_schedule_t. Both eventually inherit from ucc_coll_req_t. So both task and schedule can be used as user request. Why? - because, in case of, for example, basic CL we don't need any complicated schedules and might want to have as thin layer as possible on top of TL UCP (returning tl_ucp-task_t to user). At the same time we can still build schedules (both at CL and TL levels) and use them with the same interface.
Thirdly, to reduce number of branches in a fast path (collective_post, test, finalize) collreq has pointers to internal post/finalize implementations. Those pointers are initialized during collective_init (by the owner of a request - CL or TL).
Finally, the ucc_coll_task and schedule will be progressed via internal progress_queue interface - this is done in the next PR.

Additionally this PR adds:
mpool infrastructure, tl_ucp task allocation and init using mpool, cl basic implementation of coll_init using tl_ucp_team, 
mpi based test - simple infra and barrier test stubs

### DIFF
--- a/src/core/ucc_coll.c
+++ b/src/core/ucc_coll.c
@@ -28,8 +28,8 @@ ucc_status_t ucc_collective_init(ucc_coll_op_args_t *coll_args,
     /* TO discuss: maybe we want to pass around user pointer ? */
     memcpy(&op_args.args, coll_args, sizeof(ucc_coll_op_args_t));
     cl_team = ucc_select_cl_team(coll_args, team);
-    status = UCC_CL_TEAM_IFACE(cl_team)->coll.init(&op_args, &cl_team->super,
-                                                   &task);
+    status =
+        UCC_CL_TEAM_IFACE(cl_team)->coll.init(&op_args, &cl_team->super, &task);
     if (status != UCC_OK) {
         //TODO more descriptive error msg
         ucc_error("failed to init collective");
--- a/src/core/ucc_coll.c
+++ b/src/core/ucc_coll.c
@@ -9,6 +9,7 @@
 #include "components/cl/ucc_cl.h"
 #include "utils/ucc_malloc.h"
 #include "utils/ucc_log.h"
+#include "schedule/ucc_schedule.h"
 
 static ucc_cl_team_t *ucc_select_cl_team(ucc_coll_op_args_t *coll_args,
                                          ucc_team_t *team)
@@ -38,3 +39,16 @@ ucc_status_t ucc_collective_init(ucc_coll_op_args_t *coll_args,
     *request = &task->super;
     return UCC_OK;
 }
+
+ucc_status_t ucc_collective_post(ucc_coll_req_h request)
+{
+    ucc_coll_task_t *task = ucc_derived_of(request, ucc_coll_task_t);
+    return task->post(request);
+}
+
+ucc_status_t ucc_collective_finalize(ucc_coll_req_h request)
+{
+    ucc_coll_task_t *task = ucc_derived_of(request, ucc_coll_task_t);
+    return task->finalize(request);
+}
+
--- a/src/core/ucc_coll.c
+++ b/src/core/ucc_coll.c
@@ -1,0 +1,40 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#include "config.h"
+#include "ucc_team.h"
+#include "ucc_context.h"
+#include "components/cl/ucc_cl.h"
+#include "utils/ucc_malloc.h"
+#include "utils/ucc_log.h"
+
+static ucc_cl_team_t *ucc_select_cl_team(ucc_coll_op_args_t *coll_args,
+                                         ucc_team_t *team)
+{
+    /* TODO: collective CL selection logic will be there.
+       for now just return 1st CL on a list */
+    return team->cl_teams[0];
+}
+
+ucc_status_t ucc_collective_init(ucc_coll_op_args_t *coll_args,
+                                 ucc_coll_req_h *request, ucc_team_h team)
+{
+    ucc_cl_team_t          *cl_team;
+    ucc_base_coll_op_args_t op_args;
+    ucc_status_t            status;
+    ucc_coll_task_t        *task;
+    /* TO discuss: maybe we want to pass around user pointer ? */
+    memcpy(&op_args.args, coll_args, sizeof(ucc_coll_op_args_t));
+    cl_team = ucc_select_cl_team(coll_args, team);
+    status = UCC_CL_TEAM_IFACE(cl_team)->coll.init(&op_args, &cl_team->super,
+                                                   &task);
+    if (status != UCC_OK) {
+        //TODO more descriptive error msg
+        ucc_error("failed to init collective");
+        return status;
+    }
+    *request = &task->super;
+    return UCC_OK;
+}
--- a/src/schedule/ucc_schedule.c
+++ b/src/schedule/ucc_schedule.c
@@ -1,0 +1,78 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+#include "ucc_schedule.h"
+#include "utils/ucc_compiler_def.h"
+
+void ucc_event_manager_init(ucc_event_manager_t *em)
+{
+    int i;
+    for (i = 0; i < UCC_EVENT_LAST; i++) {
+        em->listeners_size[i] = 0;
+    }
+}
+
+void ucc_event_manager_subscribe(ucc_event_manager_t *em,
+                                 ucc_event_t event,
+                                 ucc_coll_task_t *task)
+{
+    em->listeners[event][em->listeners_size[event]] = task;
+    em->listeners_size[event]++;
+}
+
+void ucc_coll_task_init(ucc_coll_task_t *task)
+{
+    task->super.status = UCC_OPERATION_INITIALIZED;
+    ucc_event_manager_init(&task->em);
+}
+
+ucc_status_t ucc_event_manager_notify(ucc_event_manager_t *em,
+                                      ucc_event_t event)
+{
+    ucc_coll_task_t *task;
+    ucc_status_t     status;
+    int              i;
+
+    for (i = 0; i < em->listeners_size[event]; i++) {
+        task   = em->listeners[event][i];
+        status = task->handlers[event](task);
+        if (status != UCC_OK) {
+            return status;
+        }
+    }
+    return UCC_OK;
+}
+
+static ucc_status_t ucc_schedule_completed_handler(ucc_coll_task_t *task)
+{
+    ucc_schedule_t *self = ucc_container_of(task, ucc_schedule_t, super);
+    self->n_completed_tasks += 1;
+    if (self->n_completed_tasks == self->n_tasks) {
+        self->super.super.status = UCC_OK;
+    }
+    return UCC_OK;
+}
+
+void ucc_schedule_init(ucc_schedule_t *schedule, ucc_context_t *ctx)
+{
+    ucc_coll_task_init(&schedule->super);
+    schedule->super.handlers[UCC_EVENT_COMPLETED] = ucc_schedule_completed_handler;
+    schedule->n_completed_tasks = 0;
+    schedule->ctx               = ctx;
+    schedule->n_tasks           = 0;
+}
+
+void ucc_schedule_add_task(ucc_schedule_t *schedule, ucc_coll_task_t *task)
+{
+    ucc_event_manager_subscribe(&task->em, UCC_EVENT_COMPLETED, &schedule->super);
+    task->schedule = schedule;
+    schedule->n_tasks++;
+}
+
+ucc_status_t ucc_schedule_start(ucc_schedule_t *schedule)
+{
+    schedule->super.super.status = UCC_INPROGRESS;
+    return ucc_event_manager_notify(&schedule->super.em,
+                                    UCC_EVENT_SCHEDULE_STARTED);
+}
--- a/src/schedule/ucc_schedule.c
+++ b/src/schedule/ucc_schedule.c
@@ -5,25 +5,27 @@
 #include "ucc_schedule.h"
 #include "utils/ucc_compiler_def.h"
 
-void ucc_event_manager_init(ucc_event_manager_t *em)
+ucc_status_t ucc_event_manager_init(ucc_event_manager_t *em)
 {
     int i;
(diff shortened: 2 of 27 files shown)


src/schedule/ucc_schedule.c
```diff
@@ -0,0 +1,79 @@
...
+{
+    ucc_coll_task_t *task;
+    ucc_status_t     status;
+    int              i;
+
+    for (i = 0; i < em->listeners_size[event]; i++) {
+        task   = em->listeners[event][i];
+        status = task->handlers[event](task);
+        if (status != UCC_OK) {
+            return status;
```
Sergei-Lebedev, line 40: add error message that task failed

src/schedule/ucc_schedule.c
```diff
@@ -0,0 +1,79 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+#include "ucc_schedule.h"
+#include "utils/ucc_compiler_def.h"
+
+void ucc_event_manager_init(ucc_event_manager_t *em)
```
Sergei-Lebedev, line 8: return ucc_status_t

src/schedule/ucc_schedule.c
```diff
@@ -0,0 +1,79 @@
...
+{
+    int i;
+    for (i = 0; i < UCC_EVENT_LAST; i++) {
+        em->listeners_size[i] = 0;
+    }
+}
+
+void ucc_event_manager_subscribe(ucc_event_manager_t *em, ucc_event_t event,
```
Sergei-Lebedev, line 16: return ucc_status_t

src/schedule/ucc_schedule.c
```diff
@@ -0,0 +1,79 @@
...
+                                 ucc_coll_task_t *task)
+{
+    em->listeners[event][em->listeners_size[event]] = task;
+    em->listeners_size[event]++;
+}
+
+void ucc_coll_task_init(ucc_coll_task_t *task)
```
Sergei-Lebedev, line 23: ucc_status_t

src/schedule/ucc_schedule.c
```diff
@@ -0,0 +1,79 @@
...
+    int i;
+    for (i = 0; i < UCC_EVENT_LAST; i++) {
+        em->listeners_size[i] = 0;
+    }
+}
+
+void ucc_event_manager_subscribe(ucc_event_manager_t *em, ucc_event_t event,
+                                 ucc_coll_task_t *task)
+{
+    em->listeners[event][em->listeners_size[event]] = task;
```
Sergei-Lebedev, line 19: check for overflow if listeners_size > max listeners

src/schedule/ucc_schedule.c
```diff
@@ -0,0 +1,79 @@
...
+{
+    ucc_schedule_t *self = ucc_container_of(task, ucc_schedule_t, super);
+    self->n_completed_tasks += 1;
+    if (self->n_completed_tasks == self->n_tasks) {
+        self->super.super.status = UCC_OK;
+    }
+    return UCC_OK;
+}
+
+void ucc_schedule_init(ucc_schedule_t *schedule, ucc_context_t *ctx)
```
Sergei-Lebedev, line 56: return ucc_status_s

src/ucc/api/ucc_def.h
```diff
@@ -66,7 +66,15 @@ typedef struct ucc_team*            ucc_team_h;
...
-typedef struct ucc_coll_req*        ucc_coll_req_h;
+typedef struct ucc_coll_req* ucc_coll_req_h;
+typedef ucc_status_t (*ucc_coll_post_fn_t)(ucc_coll_req_h request);
+typedef ucc_status_t (*ucc_coll_finalize_fn_t)(ucc_coll_req_h request);
+typedef struct ucc_coll_req {
+    ucc_status_t           status;
+    ucc_coll_post_fn_t     post;
+    ucc_coll_finalize_fn_t finalize;
+} ucc_coll_req_t;
+
```
Sergei-Lebedev, line 77: why is it part of API? do you think user wants to access post and finalize functions?

vspetrov: only to make implementations of ucc_collective_post and finalize static inlines and remove 1 extra fn call. We can this private and then move impl of ucc_collectve_post/finalize into ucc/core but we will have a fn call. Same logic applies to "status" field in the ucc_coll_req_t. If we make it part of API we can implement ucc_collective_test as just inlilne "status == UCC_OK" otherwise fn call again. Wdyt?

Sergei-Lebedev: the only issues I can think of with this approach is that we will need to track possible API changes, which means adding fields mask to coll_req, but that's probably fine

src/core/ucc_coll.c
```diff
@@ -0,0 +1,40 @@
...
+
+ucc_status_t ucc_collective_init(ucc_coll_op_args_t *coll_args,
+                                 ucc_coll_req_h *request, ucc_team_h team)
+{
+    ucc_cl_team_t          *cl_team;
+    ucc_base_coll_op_args_t op_args;
+    ucc_status_t            status;
+    ucc_coll_task_t        *task;
+    /* TO discuss: maybe we want to pass around user pointer ? */
+    memcpy(&op_args.args, coll_args, sizeof(ucc_coll_op_args_t));
```
manjugv, line 29: Making a copy might be expensive for some types of collectives that have data buffers proportional to the number of ranks/pes in the collectives. Good to have a discussion on this.

src/schedule/ucc_schedule.h
```diff
@@ -0,0 +1,54 @@
...
+typedef ucc_status_t (*ucc_task_event_handler_p)(ucc_coll_task_t *task);
+
+typedef struct ucc_event_manager {
+    ucc_coll_task_t *listeners[UCC_EVENT_LAST][MAX_LISTENERS];
+    int              listeners_size[UCC_EVENT_LAST];
+} ucc_event_manager_t;
+
+typedef struct ucc_coll_task {
+    ucc_coll_req_t             super;
+    ucc_event_manager_t        em;
```
manjugv, line 29: - Why have em for every task - may be have one event manager per schedule? Particularly, do you expect multiple handlers (in this case max 8) for each event per task? 

- Another approach would be to have a common event manager for all schedules and tasks. Is there a reason to not do this way?

src/ucc/api/ucc_def.h
manjugv: I think we should not it not make it part of API and/or track it. Post/Finalize is a heavy weight operation, do you think a function call would make difference? In this case, maybe, pros doesn’t outweigh the cons.

test/mpi/test_mpi_barrier.c
```diff
@@ -0,0 +1,46 @@
...
+    /* for (i=0; i<size; i++) { */
+    /*     sleep_us = rand() % 1000; */
+    /*     usleep(sleep_us); */
+    /*     if (i == rank) { */
+    /*         printf("Rank %d checks in\n", rank); */
+    /*         fflush(stdout); */
+    /*         usleep(100); */
+    /*     } */
+    /*     do_barrier(ucc_world_team); */
+    /* } */
```
manjugv, line 42: I’m guessing you are bringing in this in your next PR ?

src/schedule/ucc_schedule.h
```diff
@@ -0,0 +1,54 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+#ifndef UCC_SCHEDULE_H_
```
manjugv, line 5: Given that this is core and basis for all collectives, we should document it.

src/schedule/ucc_schedule.h
```diff
@@ -0,0 +1,54 @@
...
+void ucc_event_manager_init(ucc_event_manager_t *em);
+void ucc_coll_task_init(ucc_coll_task_t *task);
+void ucc_event_manager_subscribe(ucc_event_manager_t *em, ucc_event_t event,
+                                 ucc_coll_task_t *task);
+ucc_status_t ucc_event_manager_notify(ucc_event_manager_t *em,
+                                      ucc_event_t event);
+void         ucc_schedule_init(ucc_schedule_t *schedule, ucc_context_t *ctx);
+void ucc_schedule_add_task(ucc_schedule_t *schedule, ucc_coll_task_t *task);
+ucc_status_t ucc_schedule_start(ucc_schedule_t *schedule);
+#endif
```
manjugv, line 54: When do you progress context, how do you get all related schedules? Is there a pointer on context?

src/schedule/ucc_schedule.c
vspetrov: i thought that handler that returns status would print it. think it is necessary?

src/schedule/ucc_schedule.c
vspetrov: why? it can't fail.

src/schedule/ucc_schedule.c
vspetrov: again, don't need. i'll add ucc-assert to check for MAX_LISTENERS. Other than that there is nothing to fail here. We will implement algorithms internally. So we now we should not overflow em.

I put some thought into it. On the one hand we can implement generic support for arbitrary number of listeners. We can put some preallocated number of listeners into em (like it is right now) and add parameter "num_listeners" to coll_task_init. Then if task knows it needs more listeners there will be dynamic memory allocation to hold them (malloc/free). Obviously we will also need a ucc_coll_task_cleanup fn to release it. Now the question is: if we do that, will we have any algorithms that really need it ? I mean, it is better writing the algorithm with less dependencies between tasks (not exceeding max number of listeners) rather than doing malloc/free in a critical path.

src/schedule/ucc_schedule.c
vspetrov: again, see above. lets discuss in WG anyways.

src/schedule/ucc_schedule.c
vspetrov: added ucc_assert

src/ucc/api/ucc_def.h
vspetrov: agree. moved those into internal coll_task. status is still part of public API - i think test is better to have w/o fn call - it is lightweight and called multiple times in a critical path.

src/core/ucc_coll.c
vspetrov: @manjugv So, this memcpy is actually just copying our coll_op_args (which is 120 bytes at the moment) regardless of the scale or coll type. However, it works only for the "non-vector" collectives, ie those that don't have multiple src,dest buffers or offset, dtypes arrays. In the latter case we will need a better solution. 

1st option that always works - deep copying the input - might be very expensive (though collectives like Alltoallv, allgatherv are very expensive on their own).
2nd option - we can have a couple of flags on the ucc_coll_op_args_t (we can use "mask" to store them, or actually add "flags" field). 1 such flag would be a hint, UCC_COLL_ARGS_PERSISTENT - this would tell the library that USER will guarantee that coll_op_args structure is not released or altered until the collective request is finalized. Another useful flag i have in mind is UCC_COLL_BLOCKING. This flag would tell the library that user invoking the collective will do the loop of progress/test right after the coll start and will not post any other collectives untils this one completes. This will be used for blocking MPI colls implementation and internally we will use higher polling settings in the TLs for this mode.

src/schedule/ucc_schedule.h
vspetrov: There is no such "other" approach actually if think about it and try writing the implementation on paper. This is well known "observer" pattern. In general each our task might have multiple dependencies AND it might depend on multiple other tasks as well. This is why each "task" has em - this way it tells other tasks (depending on it) that event happened.

Schedule is just a special kind of task - it subscribes to all other tasks in the schedule and basically accounting purposes (how many tasks from the schedule are completed).

Regarding first comment: yes currently MAX_TASK is constant for all the tasks even though most common case is just 1 or 2 listeners or even 0. See the comment above regarding making this "dynamic" and working for all possible number of listeners (imho does not worth bothering). currently task is 208 bytes. If i set MAX_LISTENERS to 4 it will be 144. Setting it less does not make any sense (in hierarchical bcast i will have task that 3 listeners). We can start with 4 and we always have the assertion in the code. So if someone will try to write an algorithm with more than 4 listeners (which is unlikely) we can consider increasing it. Or adding little bit "slower" path of dynamic listeners allocation for case like that. At any rate Cls and TLs will inherit from ucc_coll_task_t and i'm sure their additional request size will likely be larger.

test/mpi/test_mpi_barrier.c
vspetrov: yes, it is uncommented in #65

src/schedule/ucc_schedule.h
vspetrov: the context progressing, tasks queue is all added in next #60

src/schedule/ucc_schedule.h
vspetrov: i'll add basic description

src/schedule/ucc_schedule.c
Sergei-Lebedev: yes, I understand it can't fail. I thought that if in future we add something more complicated here return code can be helpful. And same for other functions.

src/schedule/ucc_schedule.h
manjugv: Yes, the task and dependencies/observer pattern is obvious for me. 

The question I was asking above was more of a leading question and also has two parts. I will ask differently: (1) How do you progress/process all tasks and schedules? If you have a centralized event manager (say per UCC context) you can look at all of the ready/completed tasks. With this approach, I'm not sure how you figure out the tasks that are ready during progress (Maybe I have to wait for the next set of PRs to answer this question. :)) (2) Having per task event manager might be expensive because either you overestimate or underestimate the listeners required. If you overestimate you waste memory, and if you underestimate (you might need dynamic allocation) as you suggest.

manjugv: As discussed in the WG, we will document the design.

src/schedule/ucc_schedule.c
vspetrov: done

Sergei-Lebedev (approved):

manjugv (approved):

PR #60, opened 2021-01-25
Context progress

vspetrov:
## What
Implements ucc_context_progress

## Why ?
core api

## How ?
ucc_context_progress consists of 2 parts.

- Firstly, any internal component may register a progress callback fn that it needs. For example, TL UCP needs to call ucp_worker_progress (even if there are no active tasks locally), mcast needs to check for potential NACKs in a progress etc. So, this PR adds a simple interface to register/deregister progress callbacks into core ucc_context_t.
- Secondly, the actual collective requests (implemented as schedules & tasks) are progressed on a ucc_context_t via progress_queue. This PR adds basic progress_queue interface and single threaded implementation (linked list).
- TL UCP : adds barrier fn stubs. Basically now we can run ucc_barrier test that triggers all the flow: collective_init -> CL_BASIC->TL_UCP tl_ucp_task is taken from the mpool and return to user; collective_post->tl_ucp_knomial_barrier_post: the task state is set to inprogress and task is queued into progress_queue; collective_test + ucc_context_progress: moves the collective state to completion; collective_finalize: returns tl_ucp task to mpool. The actual implementation of barrier itself (p2p send/recv) is missing so far.

### DIFF
--- a/src/components/tl/ucp/barrier/barrier_knomial.c
+++ b/src/components/tl/ucp/barrier/barrier_knomial.c
@@ -3,8 +3,7 @@
 #include "barrier.h"
 #include "core/ucc_progress_queue.h"
 
-ucc_status_t
-ucc_tl_ucp_barrier_knomial_progress(ucc_coll_task_t *coll_task)
+ucc_status_t ucc_tl_ucp_barrier_knomial_progress(ucc_coll_task_t *coll_task)
 {
     ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
     /* TODO implement main logic */
@@ -16,6 +15,7 @@ ucc_status_t ucc_tl_ucp_barrier_knomial_start(ucc_coll_req_t *req)
 {
     ucc_tl_ucp_task_t *task = ucc_derived_of(req, ucc_tl_ucp_task_t);
     /* TODO implement main logic */
-    ucc_progress_enqueue(UCC_TL_UCP_TEAM_CORE_CTX(task->team)->pq, &task->super);
+    ucc_progress_enqueue(UCC_TL_UCP_TEAM_CORE_CTX(task->team)->pq,
+                         &task->super);
     return UCC_OK;
 }
--- a/src/components/tl/ucp/barrier/barrier_knomial.c
+++ b/src/components/tl/ucp/barrier/barrier_knomial.c
@@ -1,0 +1,21 @@
+#include "config.h"
+#include "tl_ucp.h"
+#include "barrier.h"
+#include "core/ucc_progress_queue.h"
+
+ucc_status_t
+ucc_tl_ucp_barrier_knomial_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    /* TODO implement main logic */
+    task->super.super.status = UCC_OK;
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_ucp_barrier_knomial_start(ucc_coll_req_t *req)
+{
+    ucc_tl_ucp_task_t *task = ucc_derived_of(req, ucc_tl_ucp_task_t);
+    /* TODO implement main logic */
+    ucc_progress_enqueue(UCC_TL_UCP_TEAM_CORE_CTX(task->team)->pq, &task->super);
+    return UCC_OK;
+}
--- a/src/components/tl/ucp/barrier/barrier_knomial.c
+++ b/src/components/tl/ucp/barrier/barrier_knomial.c
@@ -11,9 +11,9 @@ ucc_status_t ucc_tl_ucp_barrier_knomial_progress(ucc_coll_task_t *coll_task)
     return UCC_OK;
 }
 
-ucc_status_t ucc_tl_ucp_barrier_knomial_start(ucc_coll_req_t *req)
+ucc_status_t ucc_tl_ucp_barrier_knomial_start(ucc_coll_task_t *coll_task)
 {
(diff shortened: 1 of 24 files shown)


vspetrov:
@manjugv @Sergei-Lebedev @alex--m any comments?

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -16,27 +17,25 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_req_t *request)
...
+    ucc_status_t          status;
+    memcpy(&task->args, &coll_args->args, sizeof(ucc_coll_op_args_t));
+    task->team           = tl_team;
+    task->super.finalize = ucc_tl_ucp_coll_finalize;
+    switch (coll_args->args.coll_type) {
+    case UCC_COLL_TYPE_BARRIER:
+        status = ucc_tl_ucp_barrier_init(task);
+        break;
+    default:
+        status = UCC_ERR_NOT_SUPPORTED;
```
Sergei-Lebedev, line 36: free task

src/components/tl/ucp/barrier/barrier_knomial.c
```diff
@@ -0,0 +1,21 @@
...
+
+ucc_status_t ucc_tl_ucp_barrier_knomial_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    /* TODO implement main logic */
+    task->super.super.status = UCC_OK;
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_ucp_barrier_knomial_start(ucc_coll_req_t *req)
```
Sergei-Lebedev, line 14: it looks like that every post function will enqueue task to the queue. Isn't it better to move enqueue to common place, e.g. ucc_collecitve_post?
why argument is of type ucc_coll_req_t?

src/core/ucc_context.c
```diff
@@ -352,7 +368,71 @@ ucc_status_t ucc_context_destroy(ucc_context_t *context)
...
 }
+
+ucc_status_t ucc_context_progress_register(ucc_context_t *ctx,
+                                           ucc_context_progress_fn_t fn,
+                                           void *progress_arg)
+{
+    int next_pos = ctx->progress_array_size;
+    if (next_pos == ctx->progress_array_max_size) {
+        ctx->progress_array_max_size += 8;
+        ctx->progress_array = ucc_realloc(ctx->progress_array,
```
Sergei-Lebedev, line 384: why not to use ucs list instead of array?

vspetrov: to have more locality? it is in the most critical path. with the array you fetch 8 pointers (and i don't expect to have more) with 1 cache line. With list you will be going through next into potentially different mem locations.

src/components/tl/ucp/barrier/barrier_knomial.c
vspetrov: 1. argument: does not really matter. i can change it to ucc_coll_task. the key here is that the argument to post would always inherit from ucc_coll_task which in turn from ucc_coll_req_t.
2. I thought that some TL/CL might want to bypass core progress, if, e.g., the task does not need "progress" function. Lets say it runs on some device, ie offloaded AND the task does not have any dependencies. Then you don't need to enqueue and can avoid overhead.

src/components/tl/ucp/tl_ucp_coll.c
vspetrov: 👍

src/components/tl/ucp/barrier/barrier_knomial.c
Sergei-Lebedev: 1. I think task is better for consistency with progress
2. Don't we need to poll for completion to know status even if there is no actual progress function?

manjugv (commented):
Overall looks good. 

Val, higher level question that we discussed in the WG : How do we address priority based progress - LIFO kinda of semantics?

src/core/ucc_context.c
vspetrov: after discussion with Sergey, agreed the benefits are immeasurable. Changed to ucc_list in favor of readability.

vspetrov:
> Overall looks good.
> 
> Val, higher level question that we discussed in the WG : How do we address priority based progress - LIFO kinda of semantics?

@manjugv we had that discussion internally, and here are some thoughts:
First and most important: we can't just progress one random task of the queue unless in the post-order: this will cause a deadlock. So, we only discussing here some prioritized progressing. How user can tell the library that some collective is desired to complete first?
Prioritized tasks. We can add a field to the coll_args : priority - say positive integer. Then internally in the context progress we can add some "fair-poll-count" kinda thing. Ie, we find high prio task and poll it more heavily. Obviously, this only applies to SW based tasks where progressing is required (ucp based, sm based). If the task is offloaded to HW there is no effect.

Regarding the use case of torch. It is not lifo if i understand correctly. The requirement there is to limit the number of outstanding collectives concurrently progressed to 1. We could introduce a env var for that: number_of_outstanding_colls_in_progress. Implementation effectively will make collective_post fn to just place coll request to the "posted_queue" and then context progress would do actual collective_post (ie starting a collective and placing into progress_queue).

IMHO, All the above should be part of separate PRs after the discussion.

Sergei-Lebedev (approved):

src/components/tl/ucp/barrier/barrier_knomial.c
vspetrov: the polling for completion is just checking the "status" field. If the TL implements everything (and does not need tasks/schedules) then it can just set COMPLETE flag. No need for queue/dequeue.

manjugv (approved):

alex--m (approved):

PR #65, opened 2021-01-29
TL/UCP: knomial barrier

vspetrov:
## What
Adds recursive knomial barrier implementation

## How ?
Generic recursive knomial collective pattern is added into src/coll_patterns/
TL/UCP send recv are implemented using ucp tagged nbx api . Note the way requests are handled in TL/UCP: we don't store ucp requests anywhere but rather test task completions via send/recv completion callbacks and corresponding set of counters: task->send_posted,send_completed,recv_posted,recv_completed

TL/UCP tag layout is changed a bit: it now has 2 more fields "scope" and "scope_id". Whenever an upper layer (CL or CORE) allocates a TL team it passes down its "scope" - integer value that defines a scope: CL will use its CL_TYPE as scope, CORE may  define its scope as CL_TYPE_LAST+1. Secondly, for each "scope" several tl teams can be created. For example, hierarchical CL will create several tl teams on different hierarchy levels. Each time different "scope_id" will be provided by CL when creating a TL/team - this will make tag separation.

Thus TL UCP team will use 3 values to identify a team uniquely: core_level ucc team_id (so far not initialized - todo for next pr), scope and scop_id. 

Some preliminary perf data. Obtained with the osu_barrier (with ompi/coll/ucc layer prototype https://github.com/vspetrov/ompi/tree/coll_ucc)
Setup: 8 nodes, x86, CX6
```
ppn=1
--
nnodes | hcoll (flat,   radix 2) | tuned | ucc (radix 2)
-- | -- | -- | --
2 | 1.07 | 1 | 0.97
4 | 2.06 | 2.09 | 1.28
6 | 3.4 | 2.4 | 3.21
8 | 2.52 | 3.23 | 2.4
```

```
ppn=28
--
nnodes | hcoll (flat, radix 2) | tuned | ucc (radix 2) | ucc (radix 4)
2           | 15.32                     | 5.45   | 4.99              | 4.47
4           |  16.72                    | 6.42   | 6.58              | 6.3
6           | 17.2                       | 7.97   | 7.12              | 6.63
8           | 17.85                     | 8.36   | 8.64              | 7.1
```
hcoll and tuned numbers here are just for some base-line comparison. Note, however, that hcoll is launched in a "flat topo" mode, ie w/o hierarchy and it is not optimized for that at all. Tuned has its own selection of algorithms. The most fair comparison is ucc_radix2 vs tuned with ppn=8. There the both use recursive doubling barrier, and demonstrate identical perf (up to sys noise). So, coll/ucc adds close to 0 overhead even compared to the implementation directly on top of PML.

fo nnodes=8 ppn=28, you have 224 ranks. So for radix=2 you expect: send/recv from extra rand + 7 complete send-recv steps in the group of 128 ranks: 7 steps ~ 1us each, + 1 more recv in the beginning and send in the end ~0.8 us each . You end up with the estimate 8.5 us which is very close to what we see.

@manjugv this is the baseline estimation that we are not doing anything wrong in tl/ucp implementation.


src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -0,0 +1,115 @@
...
+            ucp_request_cancel(UCC_TL_UCP_WORKER(team), ucp_status);    \
+            ucp_request_free(ucp_status);                               \
+            return UCC_ERR_NO_MESSAGE;                                  \
+        }                                                               \
+    } while(0)
+
+
+static inline ucc_status_t
+ucc_tl_ucp_send_nb(void *buffer, size_t msglen, ucs_memory_type_t mtype,
+                 int dest_group_rank, ucc_tl_ucp_team_t *team, uint32_t tag,
```
Sergei-Lebedev, line 55: Is tag argument needed? It's already there in task

src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -0,0 +1,115 @@
...
+    if (UCC_OK != ucp_status) {
+        UCC_TL_UCP_CHECK_REQ_STATUS();
+        task->send_posted++;
+    }
+    return UCC_OK;
+}
+
+static inline ucc_status_t
+ucc_tl_ucp_recv_nb(void *buffer, size_t msglen, ucs_memory_type_t mtype,
+                 int dest_group_rank, ucc_tl_ucp_team_t *team, uint32_t tag,
```
Sergei-Lebedev, line 89: tag?

src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -0,0 +1,115 @@
...
+                     ucs_status_string(UCS_PTR_STATUS(ucp_status)));    \
+            ucp_request_cancel(UCC_TL_UCP_WORKER(team), ucp_status);    \
+            ucp_request_free(ucp_status);                               \
+            return UCC_ERR_NO_MESSAGE;                                  \
+        }                                                               \
+    } while(0)
+
+
+static inline ucc_status_t
+ucc_tl_ucp_send_nb(void *buffer, size_t msglen, ucs_memory_type_t mtype,
```
Sergei-Lebedev, line 54: maybe change mtype from ucs_memory_type_t to ucc_memory_type_t and convert it here in-place? Otherwise we will have to convert it in algorithms

src/components/tl/ucp/tl_ucp_sendrecv.h
vspetrov: removed

src/components/tl/ucp/tl_ucp_sendrecv.h
vspetrov: 👍  did that

src/components/tl/ucp/tl_ucp.c
```diff
@@ -27,6 +29,16 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, preconnect),
      UCC_CONFIG_TYPE_UINT},
 
+    {"NPOLLS", "10",
+     "Number of ucp progress polling cycles for p2p requests testing",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, n_polls),
+     UCC_CONFIG_TYPE_UINT},
+
+    {"KN_BARRIER_RADIX", "4",
```
Sergei-Lebedev, line 37: I would suggest to follow some naming convention for parameters. For me it looks better something like
"COLLECTIVE_ALGORITHM_PARAMETER". For example BARRIER_KN_RADIX=4 and ALLTOALL_PAIRWISE_CHUNK=1 instead of KN_BARRIER_RADIX=4 and PAIRWISE_ALLTOALL_CHUNK=1

vspetrov: 👍

src/ucc/api/ucc.h
```diff
@@ -903,7 +903,7 @@ enum ucc_team_params_field {
     UCC_TEAM_PARAM_FIELD_OUTSTANDING_COLLS      = UCC_BIT(1),
     UCC_TEAM_PARAM_FIELD_EP                     = UCC_BIT(2),
     UCC_TEAM_PARAM_FIELD_EP_LIST                = UCC_BIT(3),
-    UCC_TEAM_PARAM_FIELD_EP_TYPE                = UCC_BIT(4),
+    UCC_TEAM_PARAM_FIELD_EP_RANGE               = UCC_BIT(4),
```
manjugv, line 906: ? I’m guessing it is because of naming consistency. If so, you might want to make params and attribs consistent.

src/components/base/ucc_base_iface.h
```diff
@@ -68,6 +68,14 @@ typedef struct ucc_base_context_iface {
...
 typedef struct ucc_base_team_params {
     ucc_team_params_t params;
+    int               scope; /* Scope that allocates the team. When TL team is created
+                                the scope would be a CL_TYPE. This provides a separation
+                                of teams created from different CLs with the same TL_TYPE */
+    int               scope_id; /* The id of the base_team in the specified scope. Use case:
+                                   a single CL team (e.g. basic) creates multiple TL teams
+                                   of the same type (e.g. several tl_ucp teams). In this
+                                   case CL would give those teams different scope_id. */
+    uint32_t          rank; /* Rank of a calling process in the TL/CL team */
```
manjugv, line 78: We use ep to mean a participant of the team externally. Do we want to call it team_ep instead of rank? Moving away from MPI kind of understand and constraints. I’m assuming this rank is unique to the team but not necessarily to the entire job.

src/components/tl/ucp/barrier/barrier_knomial.c
```diff
@@ -1,21 +1,150 @@
...
+
+        for (k = 1; k < radix; k++) {
+            peer = (myrank + k * radix_pow) % step_size +
+                   (myrank - myrank % step_size);
+            if (peer >= full_size)
+                continue;
+            ucc_tl_ucp_recv_nb(NULL, 0, UCC_MEMORY_TYPE_UNKNOWN, peer, team,
+                               task);
+        }
+        radix_pow *= radix;
```
manjugv, line 105: It seems like all sends/recvs from all phases are on the same task, is that correct?

Sergei-Lebedev (approved):

src/ucc/api/ucc.h
vspetrov: correct, will fix attribs as well

src/components/base/ucc_base_iface.h
vspetrov: it is internal interface, and this value is exactly "rank", not endpoint. This has nothing to do with MPI. It is a uniq process identifier within a team (not job) but it has the property: it is always contig and in the rang [0, team_size). It is just convenient to call it rank. It is part of internal interface because we always need such a value to implement any meaningful algorithm (except for some dummy algs where everybody just send everything to some random endpoint). Externally EP is supposed to be used for alternative way of team creations but to write algorithms we need "rank". If user does not give us one, we need to generate it in the "core" and then pass down to cl/tl.

src/components/tl/ucp/barrier/barrier_knomial.c
vspetrov: correct. we touched upon this several times long ago. basically we defined tasks and schedules. From that point, it is up to CL/TL to put whatever logic they want into a TASK. In TL/UCP we put "collective primitives" into task - ie, complete implementation of a collective over a tl_ucp_team is a task. This way we don't need to generate long schedules, allocate multiple task objects and move all this through the progress queue engine. There is an alternative: build schedule and move algorithm steps via progress_queue. This would be "event_driven" approach, aka, ucg-like. It is just another approach with its pros/cons. This could be another TL implementation if at some point we will see benefit from this.

manjugv: Right. With the infrastructure, it seems very straightforward to write algorithms in both approaches.

src/components/base/ucc_base_iface.h
manjugv: Can you please add "It is a uniq process identifier within a team (not job) but it has the property: it is always contig and in the rang [0, team_size). " to the documentation? It might be useful for readers.

manjugv (approved):

src/components/base/ucc_base_iface.h
vspetrov: manju, do you mean internal comment in the code? it is not part of the public api.

manjugv: Internal. Right next to the structure.

bureddy (approved):

src/components/base/ucc_base_iface.h
vspetrov: added

PR #66, opened 2021-02-02
CORE: properly init b_params for cl/tl_ctx_creat

vspetrov:
## What
estimated_num_eps/ppn were uninitialized and let to bugs in ucp init.

## Why ?
BugFix (found with gtest + valgrind)

### DIFF
--- a/src/core/ucc_context.c
+++ b/src/core/ucc_context.c
@@ -204,12 +204,12 @@ static inline void ucc_copy_context_params(ucc_context_params_t *dst,
 }
 
 static ucc_status_t ucc_create_tl_contexts(ucc_context_t *ctx,
-                                           ucc_context_config_t *ctx_config)
+                                           ucc_context_config_t *ctx_config,
+                                           ucc_base_context_params_t b_params)
 {
     ucc_lib_info_t *lib = ctx->lib;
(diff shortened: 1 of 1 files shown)


src/core/ucc_context.c
```diff
@@ -275,8 +271,12 @@ ucc_status_t ucc_context_create(ucc_lib_h lib,
     ctx->lib = lib;
     ucc_copy_context_params(&ctx->params, params);
     ucc_copy_context_params(&b_params.params, params);
-    b_params.context = ctx;
-    status = ucc_create_tl_contexts(ctx, config);
+    b_params.context           = ctx;
+    b_params.estimated_num_eps = 0; //TODO
```
manjugv, line 275: Eventually, how do you expect it to be filled?

manjugv (approved):

src/core/ucc_context.c
vspetrov: We can:
* get it from runtime via context_config_modify (first and simplest solution)
* compute inside libucc using oob if provided (e.g., if we build local info for hierarchies, it will be there)

Sergei-Lebedev (approved):

PR #67, opened 2021-02-03
CORE: update memory type detection in CUDA memory component 

SeyedMir:
## What
Update memory type detection in CUDA memory component  

## Why ?
cudaMemoryTypeUnregistered is a catchall category for any pointer CUDA doesn't know about. We also need to check for possible errors.

### DIFF
--- a/src/components/mc/cuda/mc_cuda.c
+++ b/src/components/mc/cuda/mc_cuda.c
@@ -70,7 +70,6 @@ static ucc_status_t ucc_mc_cuda_mem_type(const void *ptr,
 
 #if CUDART_VERSION >= 10000
     switch (attr.type) {
-    case cudaMemoryTypeUnregistered:
     case cudaMemoryTypeHost:
         *mem_type = UCC_MEMORY_TYPE_HOST;
         break;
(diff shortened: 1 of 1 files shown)


SeyedMir:
According to CUDA docs, cudaPointerGetAttributes() might fail.
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__UNIFIED.html#group__CUDART__UNIFIED_1gd89830e17d399c064a2f3c3fa8bb4390

cudaErrorInvalidValue means the pointer is not recognized by CUDA, but any other error return value would mean cudaPointerGetAttributes() failed.

Sergei-Lebedev:
@SeyedMir follow up WG discussion, I agree with changing return status for the case when cudaPointerGetAttributes returns  cudaMemoryTypeUnregistered but I'm not sure that ucc_mc_type should return error if one of its mc failed. I think we can just print error "failed to query CUDA for mem type" but return UCC_ERR_NOT_SUPPORTED

SeyedMir:
@Sergei-Lebedev OK, agreed. I updated the PR accordingly.

Sergei-Lebedev (approved):

vspetrov (approved):

alex--m (approved):

PR #69, opened 2021-02-09
CORE: vector reduction

Sergei-Lebedev:
## What
Adding reduction implementation for cpu and cuda components

## How ?
CPU component: simple loop.
CUDA component: bunch of kernels for different reduction operations, each kernel is a template function.


src/components/mc/cuda/kernel/mc_cuda_reduce.cu
```diff
@@ -0,0 +1,168 @@
...
+                                size_t count, ucc_datatype_t dt,
+                                ucc_reduction_op_t op)
+{
+    int th;
+    unsigned long bk;
+    cudaStream_t stream;
+
+    stream = ucc_mc_cuda.stream;
+    th = MC_CUDA_CONFIG->reduce_num_threads;
+    bk = (count + th - 1)/th;
```
vspetrov, line 135: just for my understanding: if user does not set required number of blocks, are we trying to get as many blocks as it is needed to have 1 op per thread ? Ie, total number of threads == count ?

src/components/mc/cuda/kernel/mc_cuda_reduce.cu
```diff
@@ -0,0 +1,168 @@
...
+            break;
+        case UCC_DT_FLOAT64:
+            ucc_assert(8 == sizeof(double));
+            DT_REDUCE_FLOAT(double, op, src1, src2, dst, count, stream, th, bk);
+            break;
+        default:
+            mc_error(&ucc_mc_cuda.super, "unsupported reduction type (%d)", dt);
+            return UCC_ERR_NOT_SUPPORTED;
+    }
+    CUDACHECK(cudaStreamSynchronize(stream));
```
vspetrov, line 162: maybe we need a split interface: reduce_nb + sync? can we try implementing overlap if we have such a nb reduce ?

src/core/ucc_mc.c
```diff
@@ -11,6 +11,15 @@
 
 static const ucc_mc_ops_t *mc_ops[UCC_MEMORY_TYPE_LAST];
 
+#define UCC_CHECK_MC_AVAILABLE(mc)                                             \
+    do {                                                                       \
+        if (NULL == mc_ops[mc]) {                                              \
+            ucc_error("no components supported memory type %s available",      \
```
vspetrov, line 17: minor: "no components supporting memory type %s are available"

src/core/ucc_mc.c
```diff
@@ -86,40 +96,22 @@ ucc_status_t ucc_mc_type(const void *ptr, ucc_memory_type_t *mem_type)
...
-
-    if (NULL == mc_ops[mem_type]) {
-        ucc_error("no components supported memory type %s available",
-                  ucc_memory_type_names[mem_type]);
-        return UCC_ERR_NOT_SUPPORTED;
-    }
-    status = mc_ops[mem_type]->mem_free(ptr);
-
-    return status;
+    UCC_CHECK_MC_AVAILABLE(mem_type);
```
vspetrov, line 113: do we need the check here? assert maybe is enough. if you call free this means you already succeeded allocating the ptr of this mem_type, so component should be available.

vspetrov (approved):

src/components/mc/cuda/kernel/mc_cuda_reduce.cu
Sergei-Lebedev: i think it can be useful. What reduce_nb should return? Some kind of request probably. What will return cpu reduce?

src/components/mc/cuda/kernel/mc_cuda_reduce.cu
Sergei-Lebedev: correct

manjugv (approved):

PR #70, opened 2021-02-09
TEST: fix clang-tidy warnings

Sergei-Lebedev:
## What
Fixes clang-tidy warnings

## How ?
Some functions are marked with NOLINT to suppress warnings because those functions are not fully implemented or not implemented at all

### DIFF
--- a/src/components/mc/cpu/mc_cpu.c
+++ b/src/components/mc/cpu/mc_cpu.c
@@ -42,6 +42,9 @@ static ucc_status_t ucc_mc_cpu_mem_free(void *ptr)
 static ucc_status_t ucc_mc_cpu_mem_type(const void *ptr,
                                         ucc_memory_type_t *mem_type)
 {
+    if (ptr == NULL) {
+        *mem_type = UCC_MEMORY_TYPE_HOST;
+    }
     /* not supposed to be used */
     mc_error(&ucc_mc_cpu.super, "host memory component shouldn't be used for"
                                 "mem type detection");
--- a/src/utils/ucc_class.h
+++ b/src/utils/ucc_class.h
@@ -55,6 +55,15 @@
         }                                                                      \
     }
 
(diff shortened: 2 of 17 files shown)


src/components/mc/cpu/mc_cpu.c
```diff
@@ -42,6 +42,9 @@ static ucc_status_t ucc_mc_cpu_mem_free(void *ptr)
 static ucc_status_t ucc_mc_cpu_mem_type(const void *ptr,
                                         ucc_memory_type_t *mem_type)
 {
+    if (ptr == NULL) {
```
vspetrov, line 45: i'm not sure about this one. mem_type check is in the critical path. you are checking one more branch just to silence the warning. It is user responsibility to pass proper pointers. Maybe just suppress with NOLINT here as well.

src/utils/ucc_class.h
```diff
@@ -55,6 +55,15 @@
         }                                                                      \
     }
 
+#define UCC_CLASS_CALL_BASE_INIT()                                             \
+    {                                                                          \
```
vspetrov, line 59: why double {} ?

vspetrov (approved):

src/components/mc/cpu/mc_cpu.c
Sergei-Lebedev: It doesn't matter because this function is not used, we can remove it.

src/utils/ucc_class.h
Sergei-Lebedev: just follows UCC_CLASS_CALL_SUPER_INIT style

src/components/mc/cpu/mc_cpu.c
vspetrov: 👍

alex--m (approved):

manjugv (approved):

PR #72, opened 2021-02-11
CORE: Add memory attributes query function

bureddy:


src/components/mc/base/ucc_mc_base.h
```diff
@@ -23,6 +23,48 @@ typedef enum ucc_memory_type {
     UCC_MEMORY_TYPE_UNKNOWN = UCC_MEMORY_TYPE_LAST
 } ucc_memory_type_t;
 
+/**
+ * UCC memory attributes field mask
+ */
+typedef enum ucc_mem_attr_field {
+    UCC_MEM_ATTR_FIELD_MEM_TYPE     = UCS_BIT(0), /**< Indicate if memory type
```
Sergei-Lebedev, line 30: UCC_BIT not UCS_BIT

src/components/mc/cuda/mc_cuda.c
```diff
@@ -77,46 +78,84 @@ static ucc_status_t ucc_mc_cuda_mem_free(void *ptr)
...
+        }
+#endif
+
+        if (mem_attr->field_mask & (UCC_MEM_ATTR_FIELD_ALLOC_LENGTH |
+                                    UCC_MEM_ATTR_FIELD_BASE_ADDRESS)) {
+            cu_err = cuMemGetAddressRange((CUdeviceptr*)&base_address,
+                    &alloc_length, (CUdeviceptr)ptr);
+            if (cu_err != CUDA_SUCCESS) {
+                mc_error(&ucc_mc_cuda.super,
+                         "ccuMemGetAddressRange(%p) error: %d(%s)",
```
Sergei-Lebedev, line 141: double "c" in cuMemGetAddressRange

src/components/mc/cuda/mc_cuda.c
```diff
@@ -77,46 +78,84 @@ static ucc_status_t ucc_mc_cuda_mem_free(void *ptr)
...
+    if (!(mem_attr->field_mask & (UCC_MEM_ATTR_FIELD_MEM_TYPE     |
+                                  UCC_MEM_ATTR_FIELD_BASE_ADDRESS |
+                                  UCC_MEM_ATTR_FIELD_ALLOC_LENGTH))) {
+        return UCC_OK;
     }
 
+    if (ptr == 0) {
+        mem_type = UCC_MEMORY_TYPE_HOST;
+    } else {
+        st = cudaPointerGetAttributes(&attr, ptr);
```
Sergei-Lebedev, line 101: I think we don't need to query memory type if it wasn't requested

src/core/ucc_mc.c
```diff
@@ -78,13 +78,33 @@ ucc_status_t ucc_mc_type(const void *ptr, ucc_memory_type_t *mem_type)
...
     ucc_status_t      status;
+    ucc_mem_attr_t    mem_attr;
 
     /* TODO: consider using memory type cache from UCS */
     /* by default assume memory type host */
     *mem_type = UCC_MEMORY_TYPE_HOST;
     for (mt = UCC_MEMORY_TYPE_HOST + 1; mt < UCC_MEMORY_TYPE_LAST; mt++) {
         if (NULL != mc_ops[mt]) {
-            status = mc_ops[mt]->mem_type(ptr, mem_type);
+            mem_attr.field_mask = UCC_MEM_ATTR_FIELD_MEM_TYPE;
```
Sergei-Lebedev, line 88: move it out from for loop?

src/components/mc/cpu/mc_cpu.c
```diff
@@ -70,11 +70,11 @@ static ucc_status_t ucc_mc_cpu_mem_free(void *ptr)
...
 }
 
-static ucc_status_t ucc_mc_cpu_mem_type(const void *ptr,
-                                        ucc_memory_type_t *mem_type)
+static ucc_status_t ucc_mc_cpu_mem_query(const void *ptr, size_t length,
+                                        ucc_mem_attr_t *mem_attr)
 {
     if (ptr == NULL) {
-        *mem_type = UCC_MEMORY_TYPE_HOST;
+        mem_attr->mem_type = UCC_MEMORY_TYPE_HOST;
```
Sergei-Lebedev, line 77: pls set length to zero to make linter happy
```
mc_cpu.c:73:66: warning: parameter 'length' is unused [misc-unused-parameters]
static ucc_status_t ucc_mc_cpu_mem_query(const void *ptr, size_t length,
```

Sergei-Lebedev (changes requested):

src/components/mc/cpu/mc_cpu.c
bureddy: where is this log show linter warnings? can we set Jenkins check fails when this happens so that we can look in it?

vspetrov: why length is needed at all?

src/components/mc/cuda/mc_cuda.c
```diff
@@ -77,46 +78,84 @@ static ucc_status_t ucc_mc_cuda_mem_free(void *ptr)
...
+                return UCC_ERR_NOT_SUPPORTED;
+            }
         }
     }
-    else if (attr.memoryType == cudaMemoryTypeHost) {
-        *mem_type = UCC_MEMORY_TYPE_HOST;
-    } else {
-        return UCC_ERR_NOT_SUPPORTED;
+
+    if (mem_attr->field_mask & UCC_MEM_ATTR_FIELD_MEM_TYPE) {
```
vspetrov, line 150: Generic question (maybe i'm paranoid). Isn't that a lot of branching which is not always needed? You kept separate APIs in the top component: for mem_type and mem_query. Maybe you could keep them separate internally in the mc as well. And mem_query could call mem_type if needed. wdyt?

src/components/mc/cpu/mc_cpu.c
bureddy: not needed here. but when querying base alloc length and MC do not support it, it will set what we pass to the query.

src/components/mc/cuda/mc_cuda.c
bureddy: I guess we already in slow path in querying attributes with CUDA api (slow ~-0.5 us) w/o mem type cache.
but,  i think I can remove these additional if () and move them to above where they computed.

vspetrov: btw, do we have plans to introduce mem_type cache?

bureddy: i don't think we need it for the first release. but maybe in the future.

src/components/mc/cpu/mc_cpu.c
Sergei-Lebedev: Linter results are here https://github.com/openucx/ucc/pull/72/checks?check_run_id=1882307244
We agreed to not block PR merge even if linter complains about something because there was a lot of warnings initially. Now all old warnings are fixed and we can change linter behavior.

Sergei-Lebedev (approved):

vspetrov (approved):

src/components/mc/cpu/mc_cpu.c
manjugv: > where is this log show linter warnings? can we set Jenkins check fails when this happens so that we can look in it?

I have set merge to block, if CI complains. It seems to be in clear now.

manjugv (approved):

manjugv:
@alex--m Any comments?

alex--m (approved):

manjugv:
@bureddy Can you rebase this PR? This has all approvals to merge.

PR #73, opened 2021-02-12
TL/UCP alltoall and alltoallv

Sergei-Lebedev:
## What
Adding alltoall and alltoallv implementation for UCP TL

## How ?
Pairwise algorithm

PR depends on https://github.com/openucx/ucc/pull/65
Question:
What is the proper way for user to set buffer length for alltoall, should we separate counts for altoall and alltoallv?

### DIFF
--- a/src/components/tl/ucp/alltoall/alltoall_pairwise.c
+++ b/src/components/tl/ucp/alltoall/alltoall_pairwise.c
@@ -23,34 +23,34 @@ static inline int get_send_peer(int rank, int size, int step)
 
 ucc_status_t ucc_tl_ucp_alltoall_pairwise_progress(ucc_coll_task_t *coll_task)
 {
-    ucc_tl_ucp_task_t  *task  = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
-    ucc_tl_ucp_team_t  *team  = task->team;
-    ptrdiff_t           sbuf  = (ptrdiff_t)task->args.src.info.buffer;
-    ptrdiff_t           rbuf  = (ptrdiff_t)task->args.dst.info.buffer;
-    int                 grank = team->rank;
-    int                 gsize = team->size;
-    int                 polls = 0;
-    int peer, chunk, nreqs;
-    size_t data_size;
+    ucc_tl_ucp_task_t *task  = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t *team  = task->team;
+    ptrdiff_t          sbuf  = (ptrdiff_t)task->args.src.info.buffer;
+    ptrdiff_t          rbuf  = (ptrdiff_t)task->args.dst.info.buffer;
+    int                grank = team->rank;
+    int                gsize = team->size;
+    int                polls = 0;
+    int                peer, posts, nreqs;
+    size_t             data_size;
 
-    chunk = UCC_TL_UCP_TEAM_CTX(team)->cfg.alltoall_pairwise_chunk;
-    nreqs = (chunk > gsize || chunk == 0) ? gsize: chunk;
+    posts     = UCC_TL_UCP_TEAM_LIB(team)->cfg.alltoall_pairwise_num_posts;
+    nreqs     = (posts > gsize || posts == 0) ? gsize : posts;
     data_size = (size_t)task->args.src.info.count *
                 ucc_dt_size(task->args.src.info.datatype);
     while ((task->send_posted < gsize || task->recv_posted < gsize) &&
-           (polls++ < task->n_polls+1)) {
+           (polls++ < task->n_polls)) {
         ucp_worker_progress(UCC_TL_UCP_TEAM_CTX(team)->ucp_worker);
         while ((task->recv_posted < gsize) &&
-              ((task->recv_posted - task->recv_completed) < nreqs)) {
+               ((task->recv_posted - task->recv_completed) < nreqs)) {
             peer = get_recv_peer(grank, gsize, task->recv_posted);
-            ucc_tl_ucp_recv_nb((void*)(rbuf + peer * data_size), data_size,
+            ucc_tl_ucp_recv_nb((void *)(rbuf + peer * data_size), data_size,
                                task->args.dst.info.mem_type, peer, team, task);
             polls = 0;
         }
         while ((task->send_posted < gsize) &&
-              ((task->send_posted - task->send_completed) < nreqs)) {
+               ((task->send_posted - task->send_completed) < nreqs)) {
             peer = get_send_peer(grank, gsize, task->send_posted);
-            ucc_tl_ucp_send_nb((void*)(sbuf + peer * data_size), data_size,
+            ucc_tl_ucp_send_nb((void *)(sbuf + peer * data_size), data_size,
                                task->args.src.info.mem_type, peer, team, task);
             polls = 0;
         }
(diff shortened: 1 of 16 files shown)


src/components/tl/ucp/alltoall/alltoall_pairwise.c
```diff
@@ -0,0 +1,79 @@
...
+    data_size = (size_t)task->args.buffer_info.src_counts *
+                ucc_dt_size(task->args.buffer_info.src_datatype);
+    while ((task->send_posted < gsize || task->recv_posted < gsize) &&
+           (polls++ < task->n_polls+1)) {
+        ucp_worker_progress(UCC_TL_UCP_TEAM_CTX(team)->ucp_worker);
+        while ((task->recv_posted < gsize) &&
+              ((task->recv_posted - task->recv_completed) < nreqs)) {
+            peer = get_recv_peer(grank, gsize, task->recv_posted);
+            ucc_tl_ucp_recv_nb((void*)(rbuf + peer * data_size), data_size,
+                               UCC_MEMORY_TYPE_UNKNOWN, peer, team, task);
```
vspetrov, line 47: //TODO replace with coll_op_args->buffer_info.mem_Type ? or task->mem_type?

src/ucc/api/ucc.h
```diff
@@ -160,7 +160,8 @@ typedef enum {
     UCC_COLL_TYPE_GATHER             = UCC_BIT(6),
     UCC_COLL_TYPE_SCATTER            = UCC_BIT(7),
     UCC_COLL_TYPE_FANIN              = UCC_BIT(8),
-    UCC_COLL_TYPE_FANOUT             = UCC_BIT(9)
+    UCC_COLL_TYPE_FANOUT             = UCC_BIT(9),
+    UCC_COLL_TYPE_ALLTOALLV          = UCC_BIT(10)
```
vspetrov, line 164: generic question: do we want any particular "fancier" ordering of this enum ? like alphabetical .. just a thought (looking for smth nice to read ))

vspetrov (approved):
Approving this PR, but we need to decide on the coll_op_args ASAP so that we don't have to change a lot afterwards. @manjugv @alex--m @bureddy @Sergei-Lebedev do we split buffer_info into "Vector"/"scalar" or not ? Where do we put mem_type and its mask value? Do we have a proposal?

manjugv:
> Approving this PR, but we need to decide on the coll_op_args ASAP so that we don't have to change a lot afterwards. @manjugv @alex--m @bureddy @Sergei-Lebedev do we split buffer_info into "Vector"/"scalar" or not ? Where do we put mem_type and its mask value? Do we have a proposal?

It makes sense to gate this on the #80

Sergei-Lebedev:
@vspetrov @manjugv @bureddy @alex--m 
fixed comments and updated after coll args PR landed, pls review

src/utils/ucc_math.h
```diff
@@ -12,4 +12,31 @@
 #define ucc_min(_a, _b) ucs_min((_a), (_b))
 #define ucc_max(_a, _b) ucs_max((_a), (_b))
 
+static inline size_t ucc_dt_size(ucc_datatype_t dt)
```
vspetrov, line 15: i remember you said it is better to use array map instead of switch branching. can't we use it here?

vspetrov: i still don't remember why we haven't defined datatype as uint64_t like UCX. In that case we could actually pack the size of the predefined datatype in the higher bits of the uint64_t and don't need any map/switch at all. hcoll does it this way as well btw. @manjugv i remember i was making a proposal like that. what was the reason to discard it ?

manjugv:
@Sergei-Lebedev Have you tested with benchmarks OSU/IMB or something? Also, it would be good to have some performance and correctness (data integrity checks) numbers.

manjugv:
@Sergei-Lebedev Also, in future, lets have separate PRs for different collectives/algorithms. So, it gets more attention.

src/utils/ucc_math.h
manjugv: As I mentioned in our earlier conversation, let's bring this as separate PR and discuss. Very likely, it was not a standalone PR and did not get the attention it deserves.

vspetrov:
> @Sergei-Lebedev Have you tested with benchmarks OSU/IMB or something? Also, it would be good to have some performance and correctness (data integrity checks) numbers.

@manjugv would you like to establish a kind of a process for PRs with respect to performance? - then, we probably need to define system/scale/benchmark requirements for numbers collection. Because, otherwise i don't understand what can be done with those collected numbers - do we want to compare them against smth? For example, the alltoall algorithm in this PR is "pairwise" which is almost the only algorithm existing for large msg alltoall (V): if we run this PR and compare vs ompi/tuned/alltoall vs hcoll/alltoall for large msgsize it will be almost 3 identical results - same algorithm, same transport behind. Except we don't have the proper UCP settings yet: estimated_num_eps and estimated_num_ppn.

Sergei-Lebedev:
@manjugv I've tested with both OSU and IMB
```
IMB:
#-----------------------------------------------------------------------------
# Benchmarking Alltoallv
# #processes = 32
#-----------------------------------------------------------------------------
       #bytes #repetitions  t_min[usec]  t_max[usec]  t_avg[usec]      defects
            0         1000        12.00        17.09        13.63         0.00
            1         1000        12.55        17.48        13.96         0.00
            2         1000        12.44        17.42        14.02         0.00
            4         1000        12.39        16.55        13.64         0.00
            8         1000        13.36        15.41        14.42         0.00
           16         1000        12.82        14.20        13.86         0.00
           32         1000        16.42        17.55        16.86         0.00
           64         1000        15.52        16.49        16.13         0.00
          128         1000        22.39        23.62        22.82         0.00
          256         1000        26.53        27.90        26.94         0.00
          512         1000        30.75        34.21        31.62         0.00
         1024         1000        46.67        50.19        47.62         0.00
         2048         1000        49.89        52.96        50.97         0.00
         4096         1000        87.83        91.89        89.77         0.00
         8192         1000       171.37       178.30       174.77         0.00
        16384         1000       149.16       152.54       150.43         0.00
        32768         1000       223.44       227.13       225.12         0.00
        65536          640       949.34       971.97       962.15         0.00
       131072          320      2527.94      2712.76      2626.53         0.00
       262144          160      4969.33      5198.17      5086.97         0.00
       524288           80     10379.55     10800.59     10611.17         0.00
      1048576           40     23485.07     24431.69     24024.43         0.00
      2097152           20     46459.83     47485.06     46811.39         0.00
      4194304           10     90962.75     94828.73     92633.73         0.00
```

OSU alltoallv, 32 Ranks, default settings
| Size     | UCC Avg Latency(us) | Tuned Avg Latency (us) |
|----------|---------------------|------------------------|
| 1024     | 43.77               | 52.46                  |
| 2048     | 51.65               | 59.90                  |
| 4096     | 92.28               | 99.20                  |
| 8192     | 183.47              | 176.55                 |
| 16384    | 137.25              | 124.09                 |
| 32768    | 269.52              | 201.50                 |
| 65536    | 1145.63             | 1012.30                |
| 131072   | 3229.17             | 2552.07                |
| 262144   | 5950.06             | 5570.63                |
| 524288   | 11363.70            | 11936.35               |
| 1048576  | 21924.19            | 24076.57               |
| 2097152  | 43396.38            | 51299.43               |
| 4194304  | 85847.04            | 97548.59               |
| 8388608  | 170860.88           | 185206.01              |
| 16777216 | 338052.28           | 372465.27              |

src/components/tl/ucp/alltoall/alltoall_pairwise.c
```diff
@@ -0,0 +1,81 @@
...
+ucc_status_t ucc_tl_ucp_alltoall_pairwise_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t *task  = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t *team  = task->team;
+    ptrdiff_t          sbuf  = (ptrdiff_t)task->args.src.info.buffer;
+    ptrdiff_t          rbuf  = (ptrdiff_t)task->args.dst.info.buffer;
+    int                grank = team->rank;
+    int                gsize = team->size;
+    int                polls = 0;
+    int                peer, posts, nreqs;
```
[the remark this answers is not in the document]
Sergei-Lebedev: done

src/utils/ucc_math.h
Sergei-Lebedev: I've added array instead of switch in ucc_dt_size as a temporary solution

vspetrov (approved):

manjugv:
> > @Sergei-Lebedev Have you tested with benchmarks OSU/IMB or something? Also, it would be good to have some performance and correctness (data integrity checks) numbers.
> 
> @manjugv would you like to establish a kind of a process for PRs with respect to performance? - then, we probably need to define system/scale/benchmark requirements for numbers collection. Because, otherwise i don't understand what can be done with those collected numbers - do we want to compare them against smth? For example, the alltoall algorithm in this PR is "pairwise" which is almost the only algorithm existing for large msg alltoall (V): if we run this PR and compare vs ompi/tuned/alltoall vs hcoll/alltoall for large msgsize it will be almost 3 identical results - same algorithm, same transport behind. Except we don't have the proper UCP settings yet: estimated_num_eps and estimated_num_ppn.

@vspetrov  the basic component will be the most stable CL for UCC. Our goal should be to implement simple but stable algorithms here. So, when other highly performant collectives fail to deliver the promise, the recommendation from UCC developers to the users is to switch to basic CL. This should work on every system configuration. 

Performance: We should record the performance numbers for each component and store it in contrib/some other directory. This will serve as a way to sanity check the UCC performance for users. 

Criteria: We need to establish the list of benchmarks and system confirmations for reporting the performance. We discussed this but did not finalize it, lets discuss this in the WG either this week or next.

manjugv (approved):

Issue #56, opened 2021-01-20
non blocking team destroy


vspetrov:
Smth for the WG to discuss.
Currently we defined the UCC team creation to be non-blocking: team_create_post + test.
The corresponding ucc_team_destroy is blocking call.

However, in some cases ucc_team_destroy will actually involve communication among the ranks of the team that is being destroyed. Examples: TL UCP team might have the UCP EPs connected that must be disconnected during destroy - this involves ucp up_close protocol which is non-local. Another example would be mcast group destruction which requires synchronizing flush over participating ranks.

I faced the issue when i was adding team_destroy into the gtest. There we simulate multiple ranks from the single process and obviously it is impossible then to destroy the team with blocking API (the very first rank in gtest will hang). i've implemented non-blocking team destruction internally (in CL/TL and Base interface) and use it in gtest. Currently the ucc api is kept the same: blocking team destruction (implemented as while (UCC_OK != ucc_team_destroy_nb(team))).

Question: do we want to define ucc_team_destroy as non-blocking call in UCC API? Probably there is not much use case for it, but this could make it more consistent.

manjugv:
Good point. I agree that it would nice to make team_destroy non-blocking and consistent with team_post. IIRC, we discussed this and we anticipated several race cases when we make team_destroy a non-blocking call.

vspetrov:
decided for team_destroy to be non-blocking with constrain:
- team_destroy & team_create should not overlap
- 1 team is being destroyed at a time

PR #74, opened 2021-02-12
API: Changing team destroy semantics from blocking to nonblocking

manjugv:
Addressing issue #56

### DIFF
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -1,7 +1,7 @@
 /**
  * @file ucc.h
  * @date 2020
- * @copyright Copyright (C) Mellanox Technologies Ltd. 2020.  ALL RIGHTS RESERVED.
+ * @copyright Copyright (C) Mellanox Technologies Ltd. 2020-2021.  ALL RIGHTS RESERVED.
  * @copyright Copyright (C) Huawei Technologies Co., Ltd. 2020.  ALL RIGHTS RESERVED.
  *
  * See file LICENSE for terms.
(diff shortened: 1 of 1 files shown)


vspetrov:
@manjugv 1 question: do we say that team_destroy "can be non-blockin" ie it may not complete upon call return, OR we say that it is guaranteed to be non-blocking ? What if some TLs/Cls can not support that? E.g., right now we have non-blocking team creation but NCCL can not support that since nccl comm creation is a blocking collective call.

manjugv:
> @manjugv 1 question: do we say that team_destroy "can be non-blockin" ie it may not complete upon call return, OR we say that it is guaranteed to be non-blocking ? What if some TLs/Cls can not support that? E.g., right now we have non-blocking team creation but NCCL can not support that since nccl comm creation is a blocking collective call.

Non-blocking : On return there is no guarantee that it is complete. If implementations block on non-blocking calls and complete, it is not wrong, but it is just a low-quality implementation.

vspetrov:
I see that we don't introduce ucc_team_destroy_nb api but rather declaring ucc_team_destroy new semantics. Am i right that the user how wants to destroy the team need to call:
```
    ucc_status_t status;
    while (UCC_INPROGRESS == (status = ucc_team_destroy_single(team))) {
    |   ;
    }
```
do we need some blocking helper fn for user in ucc.h?
smth like
``` 
static inline ucc_team_destroy_and_wait(team) {
while ()...
}
```

manjugv:
> I see that we don't introduce ucc_team_destroy_nb api but rather declaring ucc_team_destroy new semantics. Am i right that the user how wants to destroy the team need to call:
> 
> ```
>     ucc_status_t status;
>     while (UCC_INPROGRESS == (status = ucc_team_destroy_single(team))) {
>     |   ;
>     }
> ```
> 
> do we need some blocking helper fn for user in ucc.h?
> smth like
> 
> ```
> static inline ucc_team_destroy_and_wait(team) {
> while ()...
> }
> ```

You are correct Val, I was hoping we could do the earlier with the while loop. Helper function - should it be a part of API? we don't really add any value, if we implement in the library.

vspetrov:
> > I see that we don't introduce ucc_team_destroy_nb api but rather declaring ucc_team_destroy new semantics. Am i right that the user how wants to destroy the team need to call:
> > ```
> >     ucc_status_t status;
> >     while (UCC_INPROGRESS == (status = ucc_team_destroy_single(team))) {
> >     |   ;
> >     }
> > ```
> > 
> > 
> > do we need some blocking helper fn for user in ucc.h?
> > smth like
> > ```
> > static inline ucc_team_destroy_and_wait(team) {
> > while ()...
> > }
> > ```
> 
> You are correct Val, I was hoping we could do the earlier with the while loop. Helper function - should it be a part of API? we don't really add any value, if we implement in the library.

agree

vspetrov (approved):

manjugv:
@Sergei-Lebedev Any comments?

Sergei-Lebedev (approved):

PR #75, opened 2021-02-12
Tl ucp preconnect fix

vspetrov:
## What
Fixes tl/ucp team preconnect implementation. Additionally preconnect is set to 0 by-default (off). If RC TLS is used in UCP preconnect might result in O(N^2) RC QPs connection and will have a significant impact on the team create cost. Functionality is still used by some applications - it can be enabled manually there.

## Why ?
Need to complete actual p2p to guarantee a ucp wireup. 

## How ?
Preconnected is implemented as pairwise exchange with 0 len messages.

### DIFF
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -22,7 +22,7 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
     {"", "", NULL, ucc_offsetof(ucc_tl_ucp_context_config_t, super),
      UCC_CONFIG_TYPE_TABLE(ucc_tl_context_config_table)},
 
-    {"PRECONNECT", "1024",
+    {"PRECONNECT", "0",
      "Threshold that defines the number of ranks in the UCC team/context "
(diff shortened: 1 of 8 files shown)


alex--m (approved):

manjugv (approved):

Sergei-Lebedev (approved):
@vspetrov I remember that to force wireup to finish it's enough to call flush for each endpoint. Do you think it would be easier rather than send dummy message?

vspetrov:
> @vspetrov I remember that to force wireup to finish it's enough to call flush for each endpoint. Do you think it would be easier rather than send dummy message?

I'm not sure. Is it guaranteed by the ucp? I haven't found that in the ucp.h. If it is we can switch there from send/recv, to flush nbx. I assume perf wise it is more/less the same - the cost there is the connection setup not the send/recv ping-pong. Visually, i think flush would look nicer. @Sergei-Lebedev @alex--m thoughts?

vspetrov:
after discussion in WG decided to stick with send/recv - more clean and predictable behavior.

PR #77, opened 2021-02-15
TL/UCP: context create with OOB

vspetrov:
## What
Adds context eps storage and team->ctx ranks mappers

## Why ?
Resource usage optimization (used mostly in MPI env) which is available when user provides a set of additional input for ucc context creation

## How ?
The context level storage of endpoints can be used when: (1) User passes a "context_ep" (aka world_rank or context rank) of a process when creating ctx, (2) user passes an OOB for context creation, (3) user passes a flag UCC_CONTEXT_FLAG_TEAM_EP_MAP to ctx creation , the flag says that ep_map field will be provided by user as part of ucc_team_params_t for ucc_team_create_post (this is the API change).

### DIFF
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -7,6 +7,8 @@
 #include "tl_ucp.h"
 #include "tl_ucp_tag.h"
 #include "tl_ucp_coll.h"
+#include "tl_ucp_addr.h"
+#include "tl_ucp_ep.h"
 #include <limits.h>
 
 UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
@@ -94,10 +96,11 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
         }
     }
 
-    self->ucp_context = ucp_context;
-    self->ucp_worker  = ucp_worker;
+    self->ucp_context    = ucp_context;
+    self->ucp_worker     = ucp_worker;
     self->worker_address = NULL;
-
+    self->addr_storage   = NULL;
+    self->eps            = NULL;
     ucc_status = ucc_mpool_init(&self->req_mp, sizeof(ucc_tl_ucp_task_t),
                                 UCC_CACHE_LINE_SIZE, 8, UINT_MAX, NULL, NULL,
                                 "tl_ucp_req_mp");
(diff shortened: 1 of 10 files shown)


src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -106,13 +109,46 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
...
+                      self->ucp_worker);
+    if (UCC_OK != ucc_status) {
         tl_error(self->super.super.lib, "failed to register progress function");
         goto err_thread_mode;
     }
+    if ((params->params.mask & UCC_CONTEXT_PARAM_FIELD_OOB) &&
+        (params->params.mask & UCC_CONTEXT_PARAM_FIELD_EP) &&
+        (params->params.mask & UCC_CONTEXT_PARAM_FIELD_FLAGS) &&
+        (params->params.flags & UCC_CONTEXT_FLAG_TEAM_EP_MAP)) {
+        if (params->params.ep >= params->params.oob.participants) {
```
Sergei-Lebedev, line 123: We can use this value as estimated number of endpoints in this case, right?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -106,13 +109,46 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
...
+            tl_error(self->super.super.lib,
+                     "incorrect ctx ep %llu: out of oob.participants range %u",
+                     (long long unsigned)params->params.ep,
+                     params->params.oob.participants);
+            ucc_status = UCC_ERR_INVALID_PARAM;
+            goto err_thread_mode;
```
Sergei-Lebedev, line 129: Need one more label to unregister progress function or move ucc_context_progress_register to the end of function

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -126,7 +162,23 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
 
 UCC_CLASS_CLEANUP_FUNC(ucc_tl_ucp_context_t)
 {
+    ucc_status_t status;
     tl_info(self->super.super.lib, "finalizing tl context: %p", self);
+    if (self->eps) {
+        status = ucc_tl_ucp_close_eps(self, self->eps, self->size);
+        while (UCC_OK != status) {
+            if (status < 0) {
```
Sergei-Lebedev, line 170: Does this loop spin forever until error?  😀

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -126,7 +162,23 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
...
+                tl_error(self->super.super.lib,
+                         "failure during tl_ucp_close_eps, %s",
+                         ucc_status_string(status));
+                break;
+            }
+        }
+        ucc_free(self->eps);
+    }
+    if (self->addr_storage) {
+        ucc_tl_ucp_addr_storage_free(self->addr_storage);
```
Sergei-Lebedev, line 180: Maybe move if (self->addr_storage) inside ucc_tl_ucp_addr_storage_free, we have identical checks before it here and in team cleanup

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -37,8 +37,8 @@ ucc_status_t ucc_tl_ucp_connect_team_ep(ucc_tl_ucp_team_t *team, int team_rank)
 {
     ucc_tl_ucp_context_t *ctx = UCC_TL_UCP_TEAM_CTX(team);
     if (team->context_ep_storage) {
-        int ctx_rank = -1; //TODO map to ctx rank
-        ucc_assert(0);
+        ucc_assert(team->ep_map.type > 0);
+        uint32_t ctx_rank = ucc_ep_map_eval(team->ep_map, team_rank);
```
Sergei-Lebedev, line 41: ucc_ep_map_eval might return error (ctx_rank=-1). Don't we need to check it here?

src/components/tl/ucp/tl_ucp_ep.h
```diff
@@ -39,8 +40,7 @@ static inline ucc_status_t ucc_tl_ucp_get_ep(ucc_tl_ucp_team_t *team, int rank,
     } else {
         ctx = UCC_TL_UCP_TEAM_CTX(team);
         ucc_assert(ctx->eps);
-        int ctx_rank = -1; //TODO map to ctx rank
-        ucc_assert(0);
+        uint32_t ctx_rank = ucc_ep_map_eval(team->ep_map, rank);
```
Sergei-Lebedev, line 43: check for ctx_rank == -1?

src/utils/ucc_math.h
```diff
@@ -12,4 +12,26 @@
...
+        r = map.strided.start + rank*map.strided.stride;
+        break;
+    case UCC_EP_MAP_ARRAY:
+        r = *((uint32_t*)((ptrdiff_t)map.array.map + rank*map.array.elem_size));
+        break;
+    case UCC_EP_MAP_CB:
+        r = (uint32_t)map.cb.cb(rank, map.cb.cb_ctx);
+        break;
+    default:
+        r = -1;
```
Sergei-Lebedev, line 32: function returns uint32_t and -1 looks confusing here. I would better use some constant

alex--m (approved):

manjugv (commented):
Why can’t we use this semantic to implement this optimization? When creating resources either during MPI_Init or at some point before the user (MPI) creates the MPI communicator.

Do (* a) Create UCC context without OOB + (* b) UCC team world internally
instead of 
(*) Context create with OOB. 

Then, when the user (MPI) creates the team (communicator), split the UCC (internal) world team using the UCC team split interface based on the input provided by the user. In this approach, I’m assuming that creating a context with OOB (on UCP TL) would be similar to creating a team as we scale up. 

The main advantage of this approach is that it is simpler. We don’t have to introduce any new concepts such as context eps and mapping of context eps and team eps. 

Another useful interface could be create_team_from_eps() that we had discussed earlier but decided not to include for the first version. With this, do (a) and (b) instead of context create with OOB (and create storage for eps). Then when the user creates the team (new MPI communicator), use the create_team_from_eps interface. The main advantage would be if the transport (TL) doesn't require additional coordination, it would be a local operation.  

Thoughts?

manjugv:
@vspetrov Since we have other PR, should we close this?

vspetrov:
Currently EP storage is optimized using hashing. Closing this.

PR #80, opened 2021-02-16
CORE: ucc_coll_args changes

vspetrov:
- Explicitly differentiate between "scalar" and "Vector" collective coll args via union of ucc_coll_buffer_info_t and     ucc_coll_buffer_info_v_t
- Each src and dst buffer can be either scalar or vector or both (depending on the coll_type, see tabel below)
- Rename ucc_coll_op_args_t to ucc_coll_args_t (following UCC_COLL_ARGS_FIELD naming)
- Add UCC_COLL_ARGS_FIELD_BUFFER_INFO_FLAGS and UCC_COLL_ARGS_FIELD_BUFFER_INFO_MEM_TYPE
- Add ucc_memory_type_t to ucc_coll_args_t

Mask only required to specified "optional" input for collective. Info about buffers is necessary and UCC spec defines what user has to pass. E.g., the table like below could be part of spec.

[ucc_args_spec.xlsx](https://github.com/openucx/ucc/files/5986789/ucc_args_spec.xlsx)


src/ucc/api/ucc.h
```diff
@@ -1422,7 +1443,7 @@ typedef struct ucc_coll_op_args {
     ucc_coll_id_t                   tag; /*!< Used for ordering collectives */
     uint64_t                        root; /*!< Root endpoint for rooted
                                              collectives */
-} ucc_coll_op_args_t;
+} ucc_coll_args_t;
```
bureddy, line 1446: seems ucc_memory_type_t not added as per PR description? BTW, should mem type should be aprt of coll_args (or)  buffer info?

src/ucc/api/ucc.h
```diff
@@ -1384,32 +1396,41 @@ enum ucc_coll_op_args_field {
...
     uint64_t                        mask;
-    ucc_coll_type_t                 coll_type; /*!< Type of collective operation
-                                                */
-    ucc_coll_buffer_info_t          buffer_info; /*!< Buffer info for the
-                                                   collective */
+    ucc_coll_type_t                 coll_type; /*!< Type of collective operation */
+    union {
+        ucc_coll_buffer_info_t      info;   /*!< Buffer info for the collective */
+        ucc_coll_buffer_info_v_t    info_v; /*!< Buffer info for the collective */
+    } src_buffer;
```
bureddy, line 1429: is it better src_buffer/src/ ?

vspetrov: done

src/ucc/api/ucc.h
vspetrov: done

manjugv:
Do you want to document (that you posted as excel) the usage of coll args?

manjugv:
> Do you want to document (that you posted as excel) the usage of coll args?

@vspetrov Between, if you end up using the table similar to excel, you might want to collapse the three columns (buffer, count, and datatype) into one. There is no additional value in having the separate columns.

vspetrov:
> > Do you want to document (that you posted as excel) the usage of coll args?
> 
> @vspetrov Between, if you end up using the table similar to excel, you might want to collapse the three columns (buffer, count, and datatype) into one. There is no additional value in having the separate columns.

@manjugv do you mean adding doxygen doc? src/docs/doxygen/collectives.md ? how to add table to doxygen, do you have a link or smth ?

src/ucc/api/ucc.h
```diff
@@ -1333,25 +1333,38 @@ ucc_status_t ucc_team_get_all_eps(ucc_team_h team, uint64_t **ep,
...
  */
 typedef enum {
-    UCC_COLL_BUFF_FLAG_IN_PLACE             = UCC_BIT(0),
-    UCC_COLL_BUFF_FLAG_PERSISTENT           = UCC_BIT(1),
-    UCC_COLL_BUFF_FLAG_COUNT_64BIT          = UCC_BIT(2),
-    UCC_COLL_BUFF_FLAG_DISPLACEMENTS_64BIT  = UCC_BIT(3)
-} ucc_coll_buffer_flags_t;
+    UCC_COLL_ARGS_FLAG_IN_PLACE             = UCC_BIT(0),
+    UCC_COLL_ARGS_FLAG_PERSISTENT           = UCC_BIT(1),
+    UCC_COLL_ARGS_FLAG_COUNT_64BIT          = UCC_BIT(2),
```
bureddy, line 1338: what is the use of this flag?

src/ucc/api/ucc.h
```diff
@@ -1333,25 +1333,38 @@ ucc_status_t ucc_team_get_all_eps(ucc_team_h team, uint64_t **ep,
...
 
 /**
  *  @ingroup UCC_COLLECTIVES_DT
  */
+typedef enum ucc_memory_type {
+    UCC_MEMORY_TYPE_HOST,         /**< Default system memory */
+    UCC_MEMORY_TYPE_CUDA,         /**< NVIDIA CUDA memory */
+    UCC_MEMORY_TYPE_CUDA_MANAGED, /**< NVIDIA CUDA managed memory */
+    UCC_MEMORY_TYPE_ROCM,         /**< AMD ROCM memory */
+    UCC_MEMORY_TYPE_ROCM_MANAGED, /**< AMD ROCM managed system memory */
```
bureddy, line 1350: should we change the comments text format  to "` /*!<` " to match with other places in this header file

manjugv:
> @manjugv do you mean adding doxygen tex or src/docs/doxygen/collectives.md? I don't know how to add tables there .. is there an example ?

There are two ways that I can think about how to do this: 1) create a separate markdown file like library.md/teams.md and add to the documentation 2) ASCII art/verbatim : https://stackoverflow.com/questions/16804691/doxygen-table-drawing

src/ucc/api/ucc.h
manjugv: Yes. This is required for generating description for the fields in the doxygen.

src/ucc/api/ucc.h
vspetrov: This flag specifies that the displacements and counts arrays in the invo_v are 8 bytes values in contrast to 4b if flag is not set. @manjugv correct me plz if i'm wrong

src/ucc/api/ucc.h
vspetrov: done

vspetrov:
> > > Do you want to document (that you posted as excel) the usage of coll args?
> > 
> > 
> > @vspetrov Between, if you end up using the table similar to excel, you might want to collapse the three columns (buffer, count, and datatype) into one. There is no additional value in having the separate columns.
> 
> @manjugv do you mean adding doxygen tex or src/docs/doxygen/collectives.md? I don't know how to add tables there .. is there an example ?

Added tables into the docs, plz have a look: 
[ucc.pdf](https://github.com/openucx/ucc/files/6002881/ucc.pdf) see the "collectives" section

src/ucc/api/ucc.h
manjugv: Correct. It is to support Big Count in MPI.

vspetrov:
> > > > > Do you want to document (that you posted as excel) the usage of coll args?
> > > > 
> > > > 
> > > > @vspetrov Between, if you end up using the table similar to excel, you might want to collapse the three columns (buffer, count, and datatype) into one. There is no additional value in having the separate columns.
> > > 
> > > 
> > > @manjugv do you mean adding doxygen tex or src/docs/doxygen/collectives.md? I don't know how to add tables there .. is there an example ?
> > 
> > 
> > Added tables into the docs, plz have a look:
> > [ucc.pdf](https://github.com/openucx/ucc/files/6002881/ucc.pdf) see the "collectives" section
> 
> Can you also add the source of table pdf ? Also, wondering if we can collapse the rows of buffer, count, and datatype.

The source is the excel table - i just exported it into pdf - will that do? I tried to collapse but it is not possible for all collectives. Example : reduce_scatterv. THere you need to set src.info.buffer and dst.info_v.buffer+counts+datatype. The amount of send data is defined as the sum of all recv_counts, ie dst.info_v.counts. So for some collectives we don't have to set all the fields in the "info". In the table i tried to summarize "exactly" which fields are to be set for each coll. I followed MPI spec basically because it is imho covers almost everything with respect to collectives.

src/ucc/api/ucc.h
bureddy: why can't ucc always say it support 64 bit.

manjugv:
> The source is the excel table - i just exported it into pdf - will that do? I tried to collapse but it is not possible for all collectives. Example : reduce_scatterv. THere you need to set src.info.buffer and dst.info_v.buffer+counts+datatype. The amount of send data is defined as the sum of all recv_counts, ie dst.info_v.counts. So for some collectives we don't have to set all the fields in the "info". In the table i tried to summarize "exactly" which fields are to be set for each coll. I followed MPI spec basically because it is imho covers almost everything with respect to collectives.

I thought you had a markup. If it is excel and pdf image, probably, we can ignore it. On collapsing the columns - I see your point.

src/ucc/api/ucc.h
manjugv: Remember this impacts both counts and displacements. For vector collectives, given the common case is 32 bit count and displacement, we might have more bytes than necessary (32 bits * number of ranks). We can have a union to combine 32 and 64 bit, but then it is cumbersome.

src/ucc/api/ucc.h
```diff
@@ -1384,32 +1400,41 @@ enum ucc_coll_op_args_field {
...
  *  @n @n
- *  @ref ucc_coll_op_args_t defines the parameters that can be used to customize
+ *  @ref ucc_coll_args_t defines the parameters that can be used to customize
  *  the collective operation. The "mask" bit array fields are defined by @ref
- *  ucc_coll_op_args_field. The bits in "mask" bit array is defined by @ref
- *  ucc_coll_op_args_field, which correspond to fields in structure @ref
- *  ucc_coll_op_args_t. The valid fields of the structure are specified by
- *  setting the corresponding bit to "1" in the bit-array "mask". When bits
- *  corresponding to the fields are not set, the fields are not defined.
+ *  ucc_coll_args_field. The bits in "mask" bit array is defined by @ref
```
manjugv, line 1405: ucc_coll_args_field -> ucc_coll_args_field

src/ucc/api/ucc.h
```diff
@@ -1384,32 +1400,41 @@ enum ucc_coll_op_args_field {
...
+ *  ucc_coll_args_field, which correspond to fields in structure @ref
```
manjugv, line 1406: ucc_coll_args_field -> ucc_coll_args_field

manjugv (approved):

manjugv:
@vspetrov It looks good. Please address minor comments (to make doxygen happy) and please rebase.

src/ucc/api/ucc.h
vspetrov: @bureddy main reason, we don't want to force user to re-allocate uint64_t array and re-pack (potentially large) arrays of displs/counts. So we have this flag for the user to say what is the type of storage.

src/ucc/api/ucc.h
vspetrov: We have struct: ucc_coll_args_t, but for some reason field is called "ucc_coll_arg_field_t". I think proper fix to call the field "ucc_coll_args_field_t". @manjugv ?

vspetrov:
@bureddy @Sergei-Lebedev any more comments?

bureddy (approved):

src/ucc/api/ucc.h
manjugv: > We have struct: ucc_coll_args_t, but for some reason field is called "ucc_coll_arg_field_t". I think proper fix to call the field "ucc_coll_args_field_t". @manjugv ?

Agree.

src/ucc/api/ucc.h
```diff
@@ -151,16 +151,22 @@ typedef enum {
...
-    UCC_COLL_TYPE_ALLREDUCE          = UCC_BIT(2),
-    UCC_COLL_TYPE_REDUCE             = UCC_BIT(3),
-    UCC_COLL_TYPE_ALLTOALL           = UCC_BIT(4),
-    UCC_COLL_TYPE_ALLGATHER          = UCC_BIT(5),
-    UCC_COLL_TYPE_GATHER             = UCC_BIT(6),
-    UCC_COLL_TYPE_SCATTER            = UCC_BIT(7),
-    UCC_COLL_TYPE_FANIN              = UCC_BIT(8),
-    UCC_COLL_TYPE_FANOUT             = UCC_BIT(9)
+    UCC_COLL_TYPE_ALLGATHER          = UCC_BIT(0),
+    UCC_COLL_TYPE_ALLGATHERV         = UCC_BIT(0),
```
Sergei-Lebedev, line 155: Why Allgather and Allgatherv have the same id?

Sergei-Lebedev (approved):

src/ucc/api/ucc.h
vspetrov: 👍

PR #81, opened 2021-02-17
Tl/UCP: recursive knomial allreduce

vspetrov:
## What
Adds recursive knomial algorithm for allreduce in TL/UCP
Adds implementation of ucc_test_mpi correctness test and framework. 
mpirun  -np 16  ./ucc_test_mpi -t world,half -c barrier,allreduce -M 8:65536 -d int64,uint8,float32 -o max,sum -m host

Data from 8 nodes, ppn=28. Coll/tuned + pml/ucx vs UCC + TL/UCP:
(coll/tuned is forced to use recursive doubling alg, and TL/UCP allreduce kn radix is set to 2 to have the same algos in comparison)

msgsize | ucc | tuned
-- | -- | --
4 | 8.71 | 8.37
8 | 9.22 | 8.94
16 | 11.45 | 9.02
32 | 9.83 | 9.08
64 | 9.69 | 9.53
128 | 10.65 | 10.93
256 | 13.46 | 13.42
512 | 14.63 | 14.42
1024 | 18.92 | 17.15
2048 | 20.82 | 22.08


src/core/ucc_mc.h
```diff
@@ -23,4 +23,41 @@ ucc_status_t ucc_mc_free(void *ptr, ucc_memory_type_t mem_type);
 
 ucc_status_t ucc_mc_finalize();
 
+ucc_status_t ucc_mc_reduce(const void *src1, const void *src2, void *dst,
+                           size_t count, ucc_datatype_t dt,
+                           ucc_reduction_op_t op, ucc_memory_type_t mem_type);
+ucc_status_t ucc_mc_reduce_multi(void *sbuf1, void *sbuf2, void *rbuf,
```
Sergei-Lebedev, line 29: rename sbuf/src and rbuf/dst to be consistent

test/mpi/buffer.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <iostream>
+#include <mpi.h>
+#include <ucc/api/ucc.h>
+BEGIN_C_DECLS
+#include "core/ucc_mc.h"
+#include "utils/ucc_math.h"
+END_C_DECLS
+
+template<typename T>
+void init_buffer_host(void *buf, size_t count, int _value)
```
Sergei-Lebedev, line 15: T *buf instead of void *buf

test/mpi/buffer.cc
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+void init_buffer(void *buf, size_t count, ucc_datatype_t dt,
+                 ucc_memory_type_t mt, int value)
+{
+    if (mt != UCC_MEMORY_TYPE_HOST) {
+        std::cerr << "Unsupported mt\n";
+        MPI_Abort(MPI_COMM_WORLD, -1);
+    } else {
```
Sergei-Lebedev, line 30: no need for else

test/mpi/buffer.cc
```diff
@@ -0,0 +1,99 @@
...
+            MPI_Abort(MPI_COMM_WORLD, -1);
+            break;
+        }
+    }
+}
+
+template<typename T>
+ucc_status_t compare_buffers_fp(T *b1, T *b2, size_t count) {
+    for (int i = 0; i < count; i++) {
+        if (fabs((b1[i]-b2[i])/b1[i]) > 1e-5) {
```
Sergei-Lebedev, line 73: what if b1[i] is zero

src/components/tl/ucp/allreduce/allreduce_knomial.c
```diff
@@ -0,0 +1,197 @@
...
+    size_t             data_size = count * ucc_dt_size(dt);
+    ucc_status_t       status;
+    task->allreduce_kn.phase          = PHASE_INIT;
+    task->allreduce_kn.iteration      = 0;
+    task->allreduce_kn.radix_mask_pow = 1;
+    task->allreduce_kn.radix =
+        ucc_min(UCC_TL_UCP_TEAM_CTX(team)->cfg.allreduce_kn_radix, team->size);
+    ucc_mc_alloc(&task->allreduce_kn.scratch,
+                 (task->allreduce_kn.radix - 1) * data_size,
+                 task->args.src.info.mem_type);
```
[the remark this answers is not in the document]
vspetrov: Right, just to clarify on our convention - we don't treat UNKNOWN as host, correct? It is responsibility of CORE to setup mem_type correctly unless user provided it. Just one more question: what will our mc type detection return for "host buffer, when libcuda is there but buffer was allocated w/o cuda api"? - will it return UNKNOWN for us? then it all does not work...

test/mpi/buffer.cc
vspetrov: then i will have to do the type case every time i call this init_buffer function because in the test class the buffer is void*.
Right now i just call init_buffer_host(ptr, ...), if wi declare buf argument as typed will i need to call it : init_buffer_host((float*), ...) ?or typecast from "void*" to T *buf will be made by compiler w/o warnings?

test/mpi/buffer.cc
vspetrov: it is floating point value, it can't be exactly 0, it might be very small of course, but test buffers are initialized to avoid that. how else i can test PF buffers?

src/components/tl/ucp/allreduce/allreduce_knomial.c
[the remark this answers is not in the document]
Sergei-Lebedev: > Right, just to clarify on our convention - we don't treat UNKNOWN as host, correct? It is responsibility of CORE to setup mem_type correctly unless user provided it.
I think so, if user doesn't provide memory type we will detect it in core, but we don't have this yet.

> Just one more question: what will our mc type detection return for "host buffer, when libcuda is there but buffer was allocated w/o cuda api"? - will it return UNKNOWN for us? then it all does not work...
In that case cuda mc will try to detect memory type but will fail and ucc_mc_type will return host memory. In general ucc_mc_type never returns UNKNOWN.

test/mpi/buffer.cc
Sergei-Lebedev: ok, in worst case we will get nan or inf. There are a lot of ways how to do that, but maybe for our purposes this one is enough.

test/mpi/buffer.cc
Sergei-Lebedev: I think it will cast it implicitly

vspetrov: i checked - it won't. even gcc 4.8.5 complains on type mismatch.

test/mpi/buffer.cc
vspetrov: if you know better way, plz share. quick googling didn't help me - i found essentially the same approach..

Sergei-Lebedev: One approach is to move denominator to right hand side, but it's slightly different inequality
see boost test description https://www.boost.org/doc/libs/1_32_0/libs/test/doc/components/test_tools/floating_point_comparison.html and corresponding implementation https://www.boost.org/doc/libs/1_32_0/boost/test/floating_point_comparison.hpp

Another suggestion is to compute max of b1 and b2 and use that max in denominator, that will allow to get consistent result regardless of b1 and b2 permutations
e.g. from Knuth book
```
bool approximatelyEqual(float a, float b, float epsilon)
{
    return fabs(a - b) <= ( (fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
}

bool essentiallyEqual(float a, float b, float epsilon)
{
    return fabs(a - b) <= ( (fabs(a) > fabs(b) ? fabs(b) : fabs(a)) * epsilon);
}
```

And float test from googletest is a good reference as well

src/core/ucc_mc.h
vspetrov: 👍

test/mpi/buffer.cc
vspetrov: 👍

test/mpi/buffer.cc
vspetrov: 👍 took approximatelyEqual

src/components/tl/ucp/allreduce/allreduce_knomial.c
```diff
@@ -0,0 +1,197 @@
...
+    if (KN_NODE_PROXY == node_type || KN_NODE_EXTRA == node_type) {
+        if (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
+            SAVE_STATE(PHASE_EXTRA);
+            return task->super.super.status;
+        }
+        if (KN_NODE_EXTRA == node_type) {
+            goto completion;
+        } else {
+            ucc_dt_reduce(sbuf, scratch, rbuf,
+                          count, dt, UCC_MEMORY_TYPE_HOST, &task->args);
```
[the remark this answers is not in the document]
vspetrov: 👍

src/components/tl/ucp/allreduce/allreduce.c
```diff
@@ -0,0 +1,29 @@
...
+ucc_status_t ucc_tl_ucp_allreduce_knomial_progress(ucc_coll_task_t *task);
+
+ucc_status_t ucc_tl_ucp_allreduce_init(ucc_tl_ucp_task_t *task)
+{
+    if (task->args.mask & UCC_COLL_ARGS_FIELD_USERDEFINED_REDUCTIONS) {
+        tl_error(UCC_TL_TEAM_LIB(task->team),
+                 "userdefined reductions are not supported yet");
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+    if (!UCC_IS_INPLACE(task->args) && (task->args.src.info.mem_type !=
```
manjugv, line 20: With this PR, are you also supporting allreduce when the buffers are GPU buffers? If not, you probably have check the memory type, right?

manjugv (commented):
We have discussed but not concluded about reproducibility. Recently, some of the users (weather code) asked question about this. They require library level reproducibility. Should we make a statement about it? Thoughts?

src/components/tl/ucp/allreduce/allreduce.c
vspetrov: I support them with this PR (through mc component and ucp gpu support)

vspetrov:
> We have discussed but not concluded about reproducibility. Recently, some of the users (weather code) asked question about this. They require library level reproducibility. Should we make a statement about it? Thoughts?

plz define "reproducibility": (1) "same" result buffer for the same src on the same team for the 2 identical coll_args ? (2) same as (1) + the same JOB (ie, within single parallel proc env), (3) do you imply "binary" bit-by-bit exact buffers.

I think this is much broader discussion, out of the scope of this PR

Sergei-Lebedev (approved):

manjugv:
> > We have discussed but not concluded about reproducibility. Recently, some of the users (weather code) asked question about this. They require library level reproducibility. Should we make a statement about it? Thoughts?
> 
> plz define "reproducibility": (1) "same" result buffer for the same src on the same team for the 2 identical coll_args ? (2) same as (1) + the same JOB (ie, within single parallel proc env), (3) do you imply "binary" bit-by-bit exact buffers.
> 
> I think this is much broader discussion, out of the scope of this PR

What climate users were asking was more like (3). Though related (given allreduce need to address this), I agree this is outside the scope of this PR.

manjugv (approved):

manjugv:
@alex--m Any input for @vspetrov

PR #82, opened 2021-02-18
CORE: fix tl lib finalize

Sergei-Lebedev:
## What
Add missing tl lib finalize


alex--m (approved):

vspetrov (approved):

src/core/ucc_lib.c
```diff
@@ -393,6 +393,9 @@ ucc_status_t ucc_finalize(ucc_lib_info_t *lib)
     /* If some CL components fails in finalize we will return
```
bureddy, line 393: seems this comment is incorrect as we always returning UCC_OK

Sergei-Lebedev: yes, comment is incorrect. Finalize function doesn't return any value because its type is ucs_class_cleanup_func_t.
Removed the comment.

manjugv:
@Sergei-Lebedev Can you please rebase. It looks good.

manjugv (approved):

bureddy (approved):

PR #84, opened 2021-02-19
TL/NCCL: add control path and alltoall

Sergei-Lebedev:
## What
Adding TL NCCL

## Why ?
TL NCCL can be used for GPU collectives

## How ?
New TL has been added, it's available only if CUDA and NCCL libs are there. NCCL comm is mapped to UCC team, so each new team creates new NCCL comm. CL Basis tries to initialize NCCL TL but if it can't it simply ignores it and proceed.

### DIFF
--- a/src/components/cl/basic/cl_basic_coll.c
+++ b/src/components/cl/basic/cl_basic_coll.c
@@ -6,6 +6,11 @@
 
 #include "cl_basic.h"
 
+static inline int ucc_is_inplace(ucc_coll_args_t *args) {
+    return (args->mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
+           (args->flags & UCC_COLL_ARGS_FLAG_IN_PLACE);
+}
+
 ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
                                     ucc_cl_basic_team_t *cl_basic_team,
                                     ucc_tl_team_t **tl_team)
@@ -16,8 +21,7 @@ ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
         switch (args->args.coll_type) {
         case UCC_COLL_TYPE_ALLTOALL:
             if ((args->args.src.info.mem_type == UCC_MEMORY_TYPE_CUDA) &&
-                (((args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
-                  (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
+                (ucc_is_inplace(&args->args) ||
                  (args->args.dst.info.mem_type == UCC_MEMORY_TYPE_CUDA))) {
                 *tl_team = cl_basic_team->tl_nccl_team;
                 return UCC_OK;
@@ -25,8 +29,7 @@ ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
             break;
         case UCC_COLL_TYPE_ALLTOALLV:
             if ((args->args.src.info_v.mem_type == UCC_MEMORY_TYPE_CUDA) &&
-                (((args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
-                  (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
+                (ucc_is_inplace(&args->args) ||
                  (args->args.dst.info_v.mem_type == UCC_MEMORY_TYPE_CUDA))) {
                 *tl_team = cl_basic_team->tl_nccl_team;
                 return UCC_OK;
--- a/src/components/cl/basic/cl_basic_coll.c
+++ b/src/components/cl/basic/cl_basic_coll.c
@@ -5,11 +5,7 @@
  */
 
 #include "cl_basic.h"
-
-static inline int ucc_is_inplace(ucc_coll_args_t *args) {
-    return (args->mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
-           (args->flags & UCC_COLL_ARGS_FLAG_IN_PLACE);
-}
+#include "utils/ucc_coll_utils.h"
 
 ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
                                     ucc_cl_basic_team_t *cl_basic_team,
@@ -20,17 +16,17 @@ ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
     if (cl_basic_team->tl_nccl_team != NULL) {
         switch (args->args.coll_type) {
         case UCC_COLL_TYPE_ALLTOALL:
-            if ((args->args.src.info.mem_type == UCC_MEMORY_TYPE_CUDA) &&
-                (ucc_is_inplace(&args->args) ||
-                 (args->args.dst.info.mem_type == UCC_MEMORY_TYPE_CUDA))) {
+            if ((args->args.dst.info.mem_type == UCC_MEMORY_TYPE_CUDA) &&
+                (UCC_IS_INPLACE(args->args) ||
+                 (args->args.src.info.mem_type == UCC_MEMORY_TYPE_CUDA))) {
                 *tl_team = cl_basic_team->tl_nccl_team;
                 return UCC_OK;
             }
             break;
         case UCC_COLL_TYPE_ALLTOALLV:
-            if ((args->args.src.info_v.mem_type == UCC_MEMORY_TYPE_CUDA) &&
-                (ucc_is_inplace(&args->args) ||
-                 (args->args.dst.info_v.mem_type == UCC_MEMORY_TYPE_CUDA))) {
+            if ((args->args.dst.info_v.mem_type == UCC_MEMORY_TYPE_CUDA) &&
+                (UCC_IS_INPLACE(args->args) ||
+                 (args->args.src.info_v.mem_type == UCC_MEMORY_TYPE_CUDA))) {
                 *tl_team = cl_basic_team->tl_nccl_team;
                 return UCC_OK;
             }
--- a/src/components/cl/basic/cl_basic_coll.c
+++ b/src/components/cl/basic/cl_basic_coll.c
@@ -13,13 +13,12 @@ ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
     /* If NCCL TL is available and both src and dst buffers are CUDA then
        choose NCCL TL for alltoall and alltoallv otherwise use UCP TL*/
     if (cl_basic_team->tl_nccl_team != NULL) {
-        switch (args->args.coll_type)
-        {
+        switch (args->args.coll_type) {
         case UCC_COLL_TYPE_ALLTOALL:
             if ((args->args.src.info.mem_type == UCC_MEMORY_TYPE_CUDA) &&
                 (((args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
-                (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
-                (args->args.dst.info.mem_type == UCC_MEMORY_TYPE_CUDA))) {
+                  (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
+                 (args->args.dst.info.mem_type == UCC_MEMORY_TYPE_CUDA))) {
                 *tl_team = cl_basic_team->tl_nccl_team;
                 return UCC_OK;
             }
@@ -27,8 +26,8 @@ ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
         case UCC_COLL_TYPE_ALLTOALLV:
             if ((args->args.src.info_v.mem_type == UCC_MEMORY_TYPE_CUDA) &&
                 (((args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
-                (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
-                (args->args.dst.info_v.mem_type == UCC_MEMORY_TYPE_CUDA))) {
+                  (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
+                 (args->args.dst.info_v.mem_type == UCC_MEMORY_TYPE_CUDA))) {
                 *tl_team = cl_basic_team->tl_nccl_team;
                 return UCC_OK;
             }
@@ -50,6 +49,6 @@ ucc_status_t ucc_cl_basic_coll_init(ucc_base_coll_args_t *coll_args,
     ucc_tl_team_t *tl_team;
 
     ucc_cl_basic_select_tl(coll_args, cl_team, &tl_team);
-    return UCC_TL_TEAM_IFACE(tl_team)
-        ->coll.init(coll_args, &tl_team->super, task);
+    return UCC_TL_TEAM_IFACE(tl_team)->coll.init(coll_args, &tl_team->super,
+                                                 task);
 }
--- a/src/components/cl/basic/cl_basic_coll.c
+++ b/src/components/cl/basic/cl_basic_coll.c
@@ -6,11 +6,50 @@
 
 #include "cl_basic.h"
 
+ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
+                                    ucc_cl_basic_team_t *cl_basic_team,
+                                    ucc_tl_team_t **tl_team)
+{
+    /* If NCCL TL is available and both src and dst buffers are CUDA then
+       choose NCCL TL for alltoall and alltoallv otherwise use UCP TL*/
+    if (cl_basic_team->tl_nccl_team != NULL) {
+        switch (args->args.coll_type)
+        {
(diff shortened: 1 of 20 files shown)


src/components/cl/basic/cl_basic_team.c
```diff
@@ -61,8 +71,37 @@ ucc_status_t ucc_cl_basic_team_create_test(ucc_base_team_t *cl_team)
     ucc_cl_basic_team_t    *team = ucc_derived_of(cl_team, ucc_cl_basic_team_t);
     ucc_cl_basic_context_t *ctx  = UCC_CL_BASIC_TEAM_CTX(team);
     ucc_status_t            status;
+    ucc_base_team_t        *b_team;
+
     status = UCC_TL_CTX_IFACE(ctx->tl_ucp_ctx)
                  ->team.create_test(&team->tl_ucp_team->super);
+    if (UCC_OK == status) {
```
vspetrov, line 78: This could be some common functionality for CL. Ie, some CLs may want to create multiple TL teams. E.g., hierarchical CL will create multiple different TL teams at differet hierarchy levels.
 It would be beneficial to make some helper routine at CL level which is used during cl team_create_post/test. For example:
we define a struct like {tl_team_ptr **tl_team, params, tl_ctx}. Then during team_create_post CL allocates an array of such structs that will describe ALL the TL teams it needs. Next CL calls smth like: cl_create_tl_teams_post(array_of_descriptors, &request). And from the test fn it calls the corresponding cl_create_tl_teams_test(request). 

In your example the array of descriptiors would have 2 positions: ucp and nccl. @Sergei-Lebedev wdyt ?

src/components/tl/nccl/Makefile.am
```diff
@@ -0,0 +1,20 @@
...
+# Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+#
+
+sources =             \
+	tl_nccl.h         \
+	tl_nccl.c         \
+	tl_nccl_lib.c     \
+	tl_nccl_context.c \
+	tl_nccl_team.c    \
+    tl_nccl_coll.h    \
```
vspetrov, line 11: alignment

src/components/tl/nccl/tl_nccl.h
```diff
@@ -0,0 +1,66 @@
...
+    ucc_mpool_t                  req_mp;
+} ucc_tl_nccl_context_t;
+UCC_CLASS_DECLARE(ucc_tl_nccl_context_t, const ucc_base_context_params_t *,
+                  const ucc_base_config_t *);
+
+typedef struct ucc_tl_nccl_team {
+    ucc_tl_team_t        super;
+    ncclUniqueId        *unique_id;
+    void                *oob_req;
+    ucc_team_oob_coll_t  oob;
```
vspetrov, line 49: i guess oob might be needed in multiple places as well. Does it makes sense to save it on ucc_base_team_t ?
Alternatively, we already have ucc_team_params_t (which has oob) on core ucc_team_t. Maybe we should have a pointer to core team on the ucc_base_team. Then we store oob there on params and all tls/cls have access to it.

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -0,0 +1,61 @@
...
+
+ucc_status_t ucc_nccl_collective_progress(ucc_coll_task_t *coll_task)
+{
+    cudaError_t cuda_st;
+    ucc_tl_nccl_task_t *task = ucc_derived_of(coll_task, ucc_tl_nccl_task_t);
+
+    cuda_st = cudaEventQuery(task->completed);
+    switch(cuda_st) {
+    case cudaSuccess:
+        return UCC_OK;
```
vspetrov, line 18: don't you need to set task->super.super.status = UCC_OK here?

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -0,0 +1,61 @@
...
+
+    // }
+    // NCCLCHECK(ncclGroupEnd());
+
+    return UCC_ERR_NOT_IMPLEMENTED;
+}
+
+ucc_status_t ucc_tl_nccl_alltoall_init(ucc_tl_nccl_task_t *task)
+{
+    task->super.post = ucc_nccl_alltoall_start;
```
vspetrov, line 58: task->super.status = UCC_INPROGRESS.

src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -0,0 +1,174 @@
...
+    if (status != UCC_OK) {
+        tl_error(team->super.super.context->lib, "oob req free failed");
+        goto free_unique_id;
+    }
+    CUDACHECK_GOTO(cudaStreamCreateWithFlags(&team->stream,
+                   cudaStreamNonBlocking), free_unique_id, status,
+                   team->super.super.context->lib);
+    nccl_status = ncclCommInitRank(&team->nccl_comm,team->size,
+                                   team->unique_id[0], team->rank);
+    if (nccl_status != ncclSuccess) {
```
vspetrov, line 110: NCCLCHECK_GOTO ?

test/mpi/test_mpi.c
```diff
@@ -81,6 +85,13 @@ int ucc_mpi_test_init(int argc, char **argv, uint64_t coll_types,
     MPI_Comm_rank(MPI_COMM_WORLD, &rank);
     MPI_Comm_size(MPI_COMM_WORLD, &size);
 
+#ifdef UCC_TEST_WITH_CUDA
```
vspetrov, line 88: the test_mpi is completely reworked in kn_allreduce PR, will need to add this change there. just to keep in mind.

manjugv:
@Sergei-Lebedev Can you motivate the PR better, what exactly we are supporting here? Does it support when data is in GPU buffers? stream-based collectives ? kernel-based collectives?

src/components/tl/nccl/tl_nccl_team.c
Sergei-Lebedev: I don't use NCCLCHECK here because it will print error. Instead in case of failure we print info message and continue.
We can consider additional paramer that tells TL if context create or team create is fatal error or not

src/components/tl/nccl/tl_nccl_coll.c
Sergei-Lebedev: yes, I just didn't start implementing "real" data path

src/components/cl/basic/cl_basic_team.c
Sergei-Lebedev: Agree, it would be helpful

Sergei-Lebedev:
@manjugv it's targeting GPU buffers. I've added simple selection logic to basic CL, if NCCL is available and both src and dst buffers are cuda then use NCCL TL otherwise use UCP TL.

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -0,0 +1,187 @@
...
+    ptrdiff_t           rbuf   = (ptrdiff_t)task->args.dst.info.buffer;
+    size_t data_size;
+    ucc_rank_t peer;
+
+    task->super.super.status = UCC_INPROGRESS;
+    data_size = (size_t)task->args.src.info.count *
+                ucc_dt_size(task->args.src.info.datatype);
+    NCCLCHECK_GOTO(ncclGroupStart(), exit_coll, status, UCC_TL_TEAM_LIB(team));
+    for (peer = 0; peer < gsize; peer++) {
+        NCCLCHECK_GOTO(ncclSend((void *)(sbuf + peer * data_size), data_size,
```
bureddy, line 81: shoudl we check 0 byte NCCL send/recv (https://github.com/pytorch/pytorch/blob/master/torch/csrc/cuda/nccl.cpp#L690)

src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+ucc_status_t ucc_tl_nccl_team_create_test(ucc_base_team_t *tl_team)
+{
+    ucc_tl_nccl_team_t *team = ucc_derived_of(tl_team, ucc_tl_nccl_team_t);
+    ucc_status_t status;
+    ncclResult_t nccl_status;
+
+    status = team->oob.req_test(team->oob_req);
+    if (status < 0) {
```
bureddy, line 72: != UCC_OK

src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -0,0 +1,157 @@
...
+    ucc_free(team->unique_id);
+    return status;
+}
+
+static ucc_status_t ucc_tl_nccl_coll_finalize(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_nccl_task_t *task = ucc_derived_of(coll_task, ucc_tl_nccl_task_t);
+    tl_info(task->team->super.super.context->lib, "finalizing coll task %p",
+            task);
+    cudaEventDestroy(task->completed);
```
bureddy, line 112: cuda return check

src/components/cl/basic/cl_basic_coll.c
```diff
@@ -6,11 +6,49 @@
 
 #include "cl_basic.h"
 
+ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
+                                    ucc_cl_basic_team_t *cl_basic_team,
+                                    ucc_tl_team_t **tl_team)
+{
+    /* If NCCL TL is available and both src and dst buffers are CUDA then
+       choose NCCL TL for alltoall and alltoallv otherwise use UCP TL*/
```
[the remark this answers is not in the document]
Sergei-Lebedev: It's just a placeholder for selection. In general we need to think about more flexible selection algorithm that for instance queries tl for supported collectives and memory types, but it's probably topic for another PR

src/components/cl/basic/cl_basic_coll.c
```diff
@@ -6,11 +6,49 @@
...
+                 (args->args.dst.info.mem_type == UCC_MEMORY_TYPE_CUDA))) {
+                *tl_team = cl_basic_team->tl_nccl_team;
+                return UCC_OK;
+            }
+            break;
+        case UCC_COLL_TYPE_ALLTOALLV:
+            if ((args->args.src.info_v.mem_type == UCC_MEMORY_TYPE_CUDA) &&
+                (((args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) &&
+                  (args->args.flags & UCC_COLL_ARGS_FLAG_IN_PLACE)) ||
+                 (args->args.dst.info_v.mem_type == UCC_MEMORY_TYPE_CUDA))) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: Added check for inplace, but I can't check mem_type for alltoall and alltoallv using common function because memory type field is part of info/info_v

manjugv:
Discussion with Sergey, Deva, Val: Creating multiple teams with NCCL TL can result in deadlocks (known issue). It is the user's responsibility not to create multiple teams and using them at the same time when NCCL TL is enabled. @Sergei-Lebedev will create and issue. We will capture in readme and usage notes.

manjugv (approved):

src/components/cl/basic/cl_basic_coll.c
```diff
@@ -6,11 +6,52 @@
...
+ucc_status_t ucc_cl_basic_select_tl(ucc_base_coll_args_t *args,
+                                    ucc_cl_basic_team_t *cl_basic_team,
+                                    ucc_tl_team_t **tl_team)
+{
+    /* If NCCL TL is available and both src and dst buffers are CUDA then
+       choose NCCL TL for alltoall and alltoallv otherwise use UCP TL*/
+    if (cl_basic_team->tl_nccl_team != NULL) {
+        switch (args->args.coll_type) {
+        case UCC_COLL_TYPE_ALLTOALL:
+            if ((args->args.src.info.mem_type == UCC_MEMORY_TYPE_CUDA) &&
```
vspetrov, line 23: inplace means src is ignored according to spec args table, right? is this condition correct?
suppose: inplace_is set, dst mem type is HOST, and src mem_type has garbage (since it is not required for inplace), then condition depends on the uninitialized src.info.mem_type.

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -0,0 +1,193 @@
...
+        ucc_progress_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+    } else if (status < 0) {
+        task->super.super.status = status;
+    }
+    return status;
+}
+
+ucc_status_t ucc_tl_nccl_alltoall_init(ucc_tl_nccl_task_t *task)
+{
+    if ((task->args.src.info.datatype == UCC_DT_USERDEFINED) ||
```
vspetrov, line 105: inplace check?

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -0,0 +1,193 @@
...
+        tl_error(UCC_TL_TEAM_LIB(task->team),
+                 "user defined datatype is not supported");
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+    task->super.post     = ucc_nccl_alltoall_start;
+    task->super.progress = ucc_nccl_collective_progress;
+    return UCC_OK;
+}
+
+static inline size_t get_count(ucc_count_t *counts, ucc_rank_t idx,
```
vspetrov, line 116: we already have those get_count/disp in the coll_utils

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -0,0 +1,193 @@
...
+        ucc_progress_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+    } else if (status < 0) {
+        task->super.super.status = status;
+    }
+    return status;
+}
+
+ucc_status_t ucc_tl_nccl_alltoallv_init(ucc_tl_nccl_task_t *task)
+{
+    if ((task->args.src.info_v.datatype == UCC_DT_USERDEFINED) ||
```
vspetrov, line 184: inplace check

src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -0,0 +1,163 @@
...
+    self->unique_id = ucc_malloc(sizeof(ncclUniqueId) * (self->size + 1),
+                                 "tl_nccl_unique_id");
+    if (!self->unique_id) {
+        tl_error(ctx->super.super.lib,
+                 "failed to allocate %zd bytes for unique_id array",
+                 sizeof(ncclUniqueId) * (self->size + 1));
+        return UCC_ERR_NO_MEMORY;
+    }
+    if (self->rank == 0) {
+        NCCLCHECK_GOTO(ncclGetUniqueId(&self->unique_id[self->size]),
```
vspetrov, line 30: this will produce a hang, all ranks except for 0 will hang in the allgather. do we want to use a special value to indicate the error and check after the allgather?

src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -0,0 +1,163 @@
...
+                                                      ucc_tl_nccl_context_t);
+    ucc_tl_nccl_task_t    *task;
+    ucc_status_t status;
+
+    task = ucc_mpool_get(&nccl_ctx->req_mp);
+    ucc_coll_task_init(&task->super);
+    memcpy(&task->args, &coll_args->args, sizeof(ucc_coll_args_t));
+    task->team = nccl_team;
+    task->super.finalize = ucc_tl_nccl_coll_finalize;
+    CUDACHECK_GOTO(cudaEventCreateWithFlags(&task->completed,
```
vspetrov, line 135: is cudaEventCreateWithFlags costly? Do you want another mpool for that? (i think Devendar used a pool for cuda events in hcoll's nccl bcol, but i don't know if it is needed)

src/components/tl/ucc_tl.h
```diff
@@ -84,11 +85,31 @@ ucc_status_t ucc_tl_context_get(ucc_context_t *ctx, ucc_tl_type_t type,
                                 ucc_tl_context_t **tl_context);
 ucc_status_t ucc_tl_context_put(ucc_tl_context_t *tl_context);
 
+typedef struct ucc_team_create_multiple_req {
```
vspetrov, line 88: I would do that little bit differently - all exactly the same logic but looks another way. Tell me what you think.
```
struct team_desc {
    ucc_tl_context_t *ctx;
    ucc_tl_team_t     *team;
    ucc_status_t       status;
    ucc_base_team_params_t p; //value instead of ptr, it is not a critical path and memcpy is not important. but don't 
                                                    need to do extra mallocs
};

struct ucc_team_create_multiple_req {
    int n_teams;
    int last_created;
    team_desc_t descs[1];
};
```
then in you req_alloc
```
req = malloc(sizeof(multiple_req_t) + (n_teams-1)*sizeof(team_desc));
```
and cleanup will be just 1 free.

Everything else will be the same.

Wdyt?

vspetrov (approved):

src/components/cl/basic/cl_basic_coll.c
Sergei-Lebedev: you are right, need to swap src and dst

src/components/tl/ucc_tl.h
Sergei-Lebedev: looks much better!

bureddy (approved):

PR #86, opened 2021-02-25
DOCS: update coll_args tables

vspetrov:
## What
Update doc coll_args tables: adding mem_type field.

Result: 
[ucc.pdf](https://github.com/openucx/ucc/files/6043510/ucc.pdf)


manjugv:
@vspetrov Have we defined the Fan-in and Fan-out routines? Without that, I'm not sure how to interpret that part of the table.

vspetrov:
> @vspetrov Have we defined the Fan-in and Fan-out routines? Without that, I'm not sure how to interpret that part of the table.

Fanin - is 0 byte reduce to root, ie, all ranks notify the root their presence.
Fanout is inverse of that - 0 byte bcast from root.

MPI does not have those. We use them internally in multiple cases to build more complex algorithms. If we think we don't need to expose them to user we can remove them.

manjugv:
Got it, I think this would be useful for users building custom/hierarchies on top of UCC as well. This shows the need for documenting collective semantics. We need to make it explicit somewhere. I'm not sure what is the correct place. Spec or header ?

manjugv:
Overall, the table looks fine to me.

manjugv (approved):

manjugv:
@vspetrov Did you check the documentation in html format? I did require png format files for html documentation. I was wondering, if you had the same issue.

vspetrov:
> Got it, I think this would be useful for users building custom/hierarchies on top of UCC as well. This shows the need for documenting collective semantics. We need to make it explicit somewhere. I'm not sure what is the correct place. Spec or header ?

i would say "spec". too long descriptions in the header make it look too complicated imho.

vspetrov:
> @vspetrov Did you check the documentation in html format? I did require png format files for html documentation. I was wondering, if you had the same issue.

:+1 added "png" tables for html

bureddy:
Don't we need allreduce entry in the table? (basically saying root will be ignored)

vspetrov:
> Don't we need allreduce entry in the table? (basically saying root will be ignored)

@bureddy i don't undertand. allreduce IS there and the "root" flag is not set in the table. plz recheck and clarify what you meant

bureddy:
> > Don't we need allreduce entry in the table? (basically saying root will be ignored)
> 
> @bureddy i don't undertand. allreduce IS there and the "root" flag is not set in the table. plz recheck and clarify what you meant

my bad. looking the wrong table.

bureddy (approved):

PR #88, opened 2021-02-25
Topic/context estimated eps ppn

vspetrov:
## What
Adds ucc context configuration for estimated_num_eps/ppn.

## Why ?
Estimated num eps/ppn are optimization hints provided by the user. estimated_num_eps: estimated number of endpoints that will be connected for the given context; estimated_num_ppn  - the same but for eps on a single node. Both values exactly match the same configuration parameters for UCP. Internally they are passed to TL/UCP where they will affect transport selection (and hence better TL/UCP perf).

## How ?
Context config used to have only configs from CLs. Now we added 2 more fields on the ucc_context_config structure itself. ucc_context_config_modify is changed. ITs 3rd parameter "cls" (comma separated string of CLs to modify config) now can be NULL. In that case the main core ucc_context_config parameters are modified. ucc_context_config_print is also modified accordingly. gtest is expanded.

### DIFF
--- a/src/core/ucc_context.c
+++ b/src/core/ucc_context.c
@@ -11,6 +11,20 @@
 #include "utils/ucc_log.h"
 #include "utils/ucc_list.h"
 #include "ucc_progress_queue.h"
+
+static ucc_config_field_t ucc_context_config_table[] = {
+    {"ESTIMATED_NUM_EPS", "0",
+     "An optimization hint of how many endpoints will be created on this context",
+     ucc_offsetof(ucc_context_config_t, estimated_num_eps), UCC_CONFIG_TYPE_UINT},
+
+    {"ESTIMATED_NUM_PPN", "0",
+     "An optimization hint of how many endpoints created on this context reside"
+     " on the same node",
+     ucc_offsetof(ucc_context_config_t, estimated_num_ppn), UCC_CONFIG_TYPE_UINT},
+
+    {NULL}
+};
+
 ucc_status_t ucc_context_config_read(ucc_lib_info_t *lib, const char *filename,
                                      ucc_context_config_t **config_p)
 {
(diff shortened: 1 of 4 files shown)


Sergei-Lebedev (approved):
LGTM, the only question i have is why default value for NUM_EPS and NUM_PPN is 0? Is it some special value in UCP?

No. Core/ucc context will pass 0 for ppn/eps down to tl_ucp_context_create. There we don't pass those values to UCP if they are 0. So, basically 0 - means unset

vspetrov:
@manjugv @bureddy kind reminder, plz review

bureddy:
@vspetrovis is these hits only UCP TL in UCC?   if this the case, why can't the user can set UCX options directly?

vspetrov:
> @vspetrovis is these hits only UCP TL in UCC? if this the case, why can't the user can set UCX options directly?

@bureddy 2 reasons:  1. right now it is only for UCP/TL but potentially other tls could use it. 2  - we wanted to have a way to set it through the ucc_context_config_modify call, e.g. ompi/mca/coll/ucc would do that inside ompi code (w/o user setting env var)

bureddy (approved):

manjugv:
@vspetrov What is a typical usage for this from collectives perspective? What kind of values do you expect the users to set? For UCX, the motivation is easier. From collectives perspective, is this primarily for neighborhood collectives or a case, where team is not equal or much much lesser than "Communicator World"

vspetrov:
> @vspetrov What is a typical usage for this from collectives perspective? What kind of values do you expect the users to set? For UCX, the motivation is easier. From collectives perspective, is this primarily for neighborhood collectives or a case, where team is not equal or much much lesser than "Communicator World"

It is exactly same logic as in UCP: the value is the "hint" from user for the ucc_context. User says that it is expected to have that many "total eps" and "local node eps" on that context during the job. In other words, UCC can expect to have a team created with the context that will have "esimated_num_eps" total processes and "ppn" of them will belong to the same local node. Again - it is a hint, not a strikt obligation.

PR #90, opened 2021-03-02
CORE: ctx progress queue multi thread

lappazos:
## What
Implements MT ucc_context_progress

## Why ?
core api

## How ?
In a multi-threaded environment, where we have multiple threads inserting and progressing collective requests to the progress_queue, we need a thread safe data structure.
Here we use double array structure where all new tasks inserted to the primary array and old tasks to the secondary. tasks are being extracted only from the primary array. when primary is empty, they switch roles. All insertions & deletions are performed with atomic operations.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -62,6 +62,7 @@ libucc_la_SOURCES =                  \
 	core/ucc_coll.c                  \
 	core/ucc_progress_queue.c        \
 	core/ucc_progress_queue_st.c     \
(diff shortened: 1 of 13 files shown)


lappazos:
Still need to add mpi test, will add it tomorrow

manjugv:
Blocker for v1.0 ?

lappazos:
replacment in https://github.com/openucx/ucc/pull/125

PR #95, opened 2021-03-05
TEST: test reduction operations

Sergei-Lebedev:
## What
Add reduction tests to gtest

### DIFF
--- a/test/gtest/core/test_mc_reduce.cc
+++ b/test/gtest/core/test_mc_reduce.cc
@@ -1,0 +1,257 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+extern "C" {
+#include <core/ucc_mc.h>
+}
+#include <common/test.h>
+
+template<ucc_datatype_t, template <typename P> class op>
+struct ReductionTest;
+
+template<template <typename P> class op>
+struct ReductionTest<UCC_DT_INT16, op>
+{
+    using type = int16_t;
+    const static ucc_datatype_t dt = UCC_DT_INT16;
+    const static ucc_reduction_op_t redop = op<type>::redop;
+    static void assert_equal(type arg1, type arg2) {
+        ASSERT_EQ(arg1, arg2);
+    }
+    static type do_op(type arg1, type arg2) {
+        op<type> _op;
+        return _op(arg1, arg2);
+    }
+};
+
+template<template <typename P> class op>
+struct ReductionTest<UCC_DT_INT32, op>
+{
+    using type = int32_t;
+    const static ucc_datatype_t dt = UCC_DT_INT32;
+    const static ucc_reduction_op_t redop = op<type>::redop;
+    static void assert_equal(type arg1, type arg2) {
+        ASSERT_EQ(arg1, arg2);
+    }
+    static type do_op(type arg1, type arg2) {
+        op<type> _op;
+        return _op(arg1, arg2);
+    }
+};
+
+template<template <typename P> class op>
+struct ReductionTest<UCC_DT_INT64, op>
+{
+    using type = int64_t;
+    const static ucc_datatype_t dt = UCC_DT_INT64;
+    const static ucc_reduction_op_t redop = op<type>::redop;
+    static void assert_equal(type arg1, type arg2) {
+        ASSERT_EQ(arg1, arg2);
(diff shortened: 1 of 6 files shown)


test/gtest/core/test_mc_reduce.cc
```diff
@@ -0,0 +1,257 @@
...
+                                           ReductionTest<UCC_DT_FLOAT64, max>,
+                                           ReductionTest<UCC_DT_FLOAT32, min>,
+                                           ReductionTest<UCC_DT_FLOAT64, min>,
+                                           ReductionTest<UCC_DT_FLOAT32, sum>,
+                                           ReductionTest<UCC_DT_FLOAT64, sum>,
+                                           ReductionTest<UCC_DT_FLOAT32, prod>,
+                                           ReductionTest<UCC_DT_FLOAT64, prod>>;
+
+TYPED_TEST_CASE(test_mc_reduce, ReductionTypesOps);
+TYPED_TEST(test_mc_reduce, ucc_op_sum) {
```
vspetrov, line 249: why "ucc_op_sum" ? it covers all combinations right?

test/gtest/core/test_mc_reduce.cc
```diff
@@ -0,0 +1,257 @@
...
+        ASSERT_EQ(arg1, arg2);
+    }
+    static type do_op(type arg1, type arg2) {
+        op<type> _op;
+        return _op(arg1, arg2);
+    }
+};
+
+template<template <typename P> class op>
+struct ReductionTest<UCC_DT_INT64, op>
```
vspetrov, line 45: Maybe make a macro to declare this template class: the code is identical almost entirely:
#define REDUCTION_TEST_DECLARE(_type, _ucc_type, _assert_eq) do {
using type = _type; \
const status ucc_datatype_t dt = _ucc_type; \
ucc_reduction_op redop = op<type>::redop;\
static void assert_equal(...) { \
_assert_eq(arg1, arg2); \
} \
...
} while (0)

and then 
REDUCTION_TEST_DECLARE(int64, UCC_DT_INT64, ASSERT_EQ);
REDUCTION_TEST_DECLARE(float, UCC_DT_FLOAT32, ASSERT_EQ_FLOAT);

etc

test/gtest/core/test_mc_reduce.cc
```diff
@@ -0,0 +1,257 @@
...
+                                           ReductionTest<UCC_DT_FLOAT32, sum>,
+                                           ReductionTest<UCC_DT_FLOAT64, sum>,
+                                           ReductionTest<UCC_DT_FLOAT32, prod>,
+                                           ReductionTest<UCC_DT_FLOAT64, prod>>;
+
+TYPED_TEST_CASE(test_mc_reduce, ReductionTypesOps);
+TYPED_TEST(test_mc_reduce, ucc_op_sum) {
+    ucc_mc_reduce(this->buf1, this->buf2, this->res,
+                  this->COUNT, TypeParam::dt,
+                  TypeParam::redop, UCC_MEMORY_TYPE_HOST);
```
vspetrov, line 252: what about MEMORY_TYPE_CUDA?

vspetrov: What about reduce_multi?

vspetrov (approved):

test/gtest/core/test_mc_reduce.cc
Sergei-Lebedev: Sure, it's just not easy to debug when test is wrapped into macro. But I want first extend test with memory type and add reduce multi.

test/gtest/core/test_mc_reduce.cc
Sergei-Lebedev: Yes, if overall approach looks good I will add cuda and reduce multi

test/gtest/core/test_mc_reduce.cc
vspetrov: we can keep it as is if you think it makes inconvenient debug

Sergei-Lebedev: i mean debug of test itself, once all tests parameters are there (reduce, reduce multi, mem types) we can wrap class declaration in macro

Sergei-Lebedev:
@vspetrov added cuda and reduce multi, pls check

vspetrov:
@manjugv @bureddy plz have a look. It's been ready for a while

bureddy (approved):

PR #97, opened 2021-03-10
TL/UCP: collective buffer pre registration optimization

bureddy:


src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
```diff
@@ -43,6 +43,25 @@ static inline size_t get_displacement(ucc_aint_t *displacements, ucc_rank_t idx,
     return ((uint32_t *)displacements)[idx];
 }
 
+static inline size_t get_total_count(ucc_count_t *counts,  ucc_tl_ucp_task_t *task)
```
vspetrov, line 46: in allgatherv PR @Sergei-Lebedev adds generic ucc_coll_args_get_count/dipls routines to the common place. Should we add this get_total_count to a similar common location?

src/components/tl/ucp/tl_ucp.c
```diff
@@ -93,3 +101,29 @@ ucs_memory_type_t ucc_memtype_to_ucs[UCC_MEMORY_TYPE_LAST+1] = {
...
+    ucc_status_t status;
+
+    mem_attr.field_mask = UCC_MEM_ATTR_FIELD_BASE_ADDRESS |
+                          UCC_MEM_ATTR_FIELD_ALLOC_LENGTH;
+    status = ucc_mc_query(addr, length, &mem_attr);
+    if (status == UCC_OK) {
+        base_address = mem_attr.base_address;
+        alloc_length = mem_attr.alloc_length;
+    } else {
+       tl_warn(UCC_TL_TEAM_LIB(team), "failed to query base addr and len");
```
vspetrov, line 120: when the failure can happen? do we need this warning? If this happens on each collective call wouldn't it clutter the output entirely?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -137,6 +137,34 @@ UCC_CLASS_CLEANUP_FUNC(ucc_tl_ucp_context_t)
...
+                              UCP_MEM_MAP_PARAM_FIELD_LENGTH  |
+                              UCP_MEM_MAP_PARAM_FIELD_MEMORY_TYPE;
+    mmap_params.address     = addr;
+    mmap_params.length      = length;
+    mmap_params.memory_type = mem_type;
+
+    /* do map and umap to populate the cache */
+    status = ucp_mem_map(ctx->ucp_context, &mmap_params, &mh);
+    if (status != UCS_OK) {
+        return UCC_ERR_NO_MEMORY;
```
vspetrov, line 157: we have ucs_status_to_ucc_status helper

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -137,6 +137,34 @@ UCC_CLASS_CLEANUP_FUNC(ucc_tl_ucp_context_t)
 
 UCC_CLASS_DEFINE(ucc_tl_ucp_context_t, ucc_tl_context_t);
 
+ucc_status_t ucc_tl_ucp_mem_map(void *addr, size_t length, ucs_memory_type_t mem_type,
+                                ucc_tl_ucp_context_t *ctx)
+{
+    ucp_mem_map_params_t mmap_params;
+    ucp_mem_h mh;
```
vspetrov, line 144: vars are not aligned (you can run clang on the commit diff with: git-clang-format-11  HEAD~1 --style file)

src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
```diff
@@ -102,6 +121,19 @@ ucc_status_t ucc_tl_ucp_alltoallv_pairwise_start(ucc_coll_task_t *coll_task)
 
     task->super.super.status = UCC_INPROGRESS;
     task->n_polls            = ucc_min(1, task->n_polls);
+
+    if (UCC_TL_UCP_TEAM_CTX(team)->cfg.pre_reg_mem) {
+        ucc_tl_ucp_pre_register_mem(team, task->args.src.info_v.buffer,
```
Sergei-Lebedev, line 126: Is it safe to register the whole range here? In alltoall we know that buffer is contig and it's ok to register it, but in alltoallv there are might be holes somewhere in the middle with unused and unallocated memory, how ucp_mem_map will handle this? Maybe instead we should check all displacements to make sure that all offsets are part of one large allocation or register only base address in alltoallv if it's possible?

vspetrov: worth discussion in WG. What we define in spec - do we allow arbitrary offsets from address 0, or all displs should be in the single buffer

Sergei-Lebedev: i already use offsets from 0 in the application, it's very convenient because it allows to save copy in and copy out if data is not contig. Maybe consider add coll arg flag e.g. UCC_COLL_ARGS_FLAG_SRC_CONTIG, UCC_COLL_ARGS_DST_CONTIG?

bureddy: > Is it safe to register the whole range here? In alltoall we know that buffer is contig and it's ok to register it, but in alltoallv there are might be holes somewhere in the middle with unused and unallocated memory, how ucp_mem_map will handle this? Maybe instead we should check all displacements to make sure that all offsets are part of one large allocation or register only base address in alltoallv if it's possible?

if that( each buffer is from a different allocation) the case, in cuda case it will just register the  first buffer region. it won't fail but may not affect.

src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
bureddy: fixed

src/components/tl/ucp/tl_ucp.c
bureddy: it possible that cuda API can return error from cuda mem component if something is really wrong.  I don't want to silently ignore this. it is readlly hard to say if it can happen each time.

src/components/tl/ucp/tl_ucp_context.c
bureddy: fixed

src/components/tl/ucp/tl_ucp_context.c
bureddy: fixed

src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
bureddy: added flags to indicate src/dst contig or not?

Sergei-Lebedev (approved):

src/components/tl/ucp/tl_ucp.c
```diff
@@ -98,3 +106,29 @@ ucs_memory_type_t ucc_memtype_to_ucs[UCC_MEMORY_TYPE_LAST+1] = {
...
+    }
+
+    status = ucc_tl_ucp_mem_map(base_address, alloc_length,
+                                ucc_memtype_to_ucs[mem_type],
+                                UCC_TL_UCP_TEAM_CTX(team));
+    if (status == UCC_OK) {
```
vspetrov, line 131: Should this be status != UCC_OK ?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -137,6 +137,34 @@ UCC_CLASS_CLEANUP_FUNC(ucc_tl_ucp_context_t)
 
 UCC_CLASS_DEFINE(ucc_tl_ucp_context_t, ucc_tl_context_t);
 
+ucc_status_t ucc_tl_ucp_mem_map(void *addr, size_t length, ucs_memory_type_t mem_type,
```
vspetrov, line 140: maybe call the nd ucc_tl_ucp_populate_rcache ? or smth like that. Otherwise, it really looks weird that "mem_map" fn does map+unmap. If rache is not available (for some reason), then the effect is that memory is de-registered.

vspetrov (changes requested):

src/components/tl/ucp/tl_ucp_context.c
bureddy: changed the func name.
yes, this is not optimization when rcache is not there.

vspetrov (approved):

PR #98, opened 2021-03-11
CORE: fixes CL libs filtering

vspetrov:
## What
Makes proper CL libs filtering based on user thread mode requirement: params.thread_mode

## Why ?
Currently filtering is not correct and uses uninitialized cl_iface.attr.thread_mode. Works only because all our tests use THREAD_SINGLE.

## How ?
cl_iface.attr is not initialized/used - removed it.
thread_mode is queried from cl lib objects.
The highest and lowest supported thread_mode is detected among all the available CL libs. Next, we first try to satisfy user requirement: ie, if at least some CL supports required thread_mode (highest_tm >= params.thread_mode) then we keep it as active thread_mode and filter out those CLs that might have lower supported tm (lowest_tm < highest_tm). If NO CL libs can support user requirement (highest_tm < params.thread_mode) then we set active thread_mode of ucc_lib to lowest_tm and keep all the opened CLs - it is still user responsibility  to query provided thread_mode; lastly if all CLs have tm >=params.thread_mode then we just keep them all.

### DIFF
--- a/src/core/ucc_lib.c
+++ b/src/core/ucc_lib.c
@@ -67,9 +67,11 @@ static ucc_status_t ucc_cl_lib_init(const ucc_lib_params_t *user_params,
 {
     int                   n_cls = ucc_global_config.cl_framework.n_components;
     uint64_t              supported_coll_types = 0;
-    ucc_thread_mode_t     supported_tm         = UCC_THREAD_MULTIPLE;
+    ucc_thread_mode_t     highest_tm           = UCC_THREAD_SINGLE;
+    ucc_thread_mode_t     lowest_tm            = UCC_THREAD_MULTIPLE;
     ucc_lib_params_t      params               = *user_params;
     ucc_cl_lib_config_t  *cl_config            = NULL;
+    ucc_cl_lib_attr_t     attrs[UCC_CL_LAST];
     ucc_cl_iface_t       *cl_iface;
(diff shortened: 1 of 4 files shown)


src/core/ucc_lib.c
```diff
@@ -67,9 +67,11 @@ static ucc_status_t ucc_cl_lib_init(const ucc_lib_params_t *user_params,
 {
     int                   n_cls = ucc_global_config.cl_framework.n_components;
     uint64_t              supported_coll_types = 0;
-    ucc_thread_mode_t     supported_tm         = UCC_THREAD_MULTIPLE;
+    ucc_thread_mode_t     highest_tm           = UCC_THREAD_SINGLE;
+    ucc_thread_mode_t     lowest_tm            = UCC_THREAD_MULTIPLE;
```
manjugv, line 71: This should be the opposite right? UCC_THREAD_SINGLE < UCC_THREAD_FUNNELED < UCC_THREAD_MULTIPLE ?

vspetrov: yes, this is exactly how it is now. These are initialized values "to search for highest and lowest" modes among CLs. See the code below. I initialize highest to SINGLE and if anything lowest to MULTIPLE in order to search max/min values.

manjugv (approved):

Sergei-Lebedev (approved):

PR #99, opened 2021-03-12
EE/EVENT Implementation

bureddy:
example API  flow:

```

    ee_params.ee_type = UCC_EE_CUDA_STREAM;
    CUDACHECK(cudaStreamCreateWithFlags(&cudaStrm,
                cudaStreamNonBlocking));
    ee_params.ee_context = (void *) cudaStrm;
    ee_params.ee_context_size = sizeof(cudaStream_t);

    UCC_CHECK(ucc_ee_create(team, &ee_params, &ee));

    UCC_CHECK(ucc_mc_alloc(&sbuf_cuda, msgsize, UCC_MEMORY_TYPE_CUDA));

    root = 0;
    args.coll_type            = UCC_COLL_TYPE_BCAST;
    args.src.info.buffer      = sbuf_cuda;
    args.src.info.count       = msgsize/ sizeof(uint32_t);
    args.src.info.datatype    = UCC_DT_INT32;
    args.src.info.mem_type    = UCC_MEMORY_TYPE_CUDA;
    args.root                 = root;
    UCC_CHECK(ucc_collective_init(&args, &req, team));


    event.ev_type = UCC_EVENT_COMPUTE_COMPLETE;
     event.ev_context = NULL;
    event.ev_context_size = 0;
    event.req = req;

    UCC_CHECK(ucc_collective_triggered_post(ee, &event));

    compute_event.ev_type = UCC_EVENT_COMPUTE_COMPLETE;
    compute_event.ev_context = NULL;
    compute_event.ev_context_size = 0;
    compute_event.req = req;

    UCC_CHECK(ucc_ee_set_event(ee, &compute_event));

    while (UCC_OK != ucc_ee_get_event(ee, &post_event)) {
           ucc_context_progress(ctx_h);
    }

    UCC_CHECK(ucc_ee_ack_event(ee, post_event));


    while (UCC_OK != ucc_collective_test(req)) {
        ucc_context_progress(ctx_h);
    }

    UCC_CHECK(ucc_collective_finalize(req));
    UCC_CHECK(ucc_mc_free(sbuf_cuda, UCC_MEMORY_TYPE_CUDA));
    UCC_CHECK(ucc_ee_destroy(ee));
```


src/components/mc/cuda/mc_cuda.h
```diff
@@ -9,19 +9,51 @@
...
 } ucc_mc_cuda_config_t;
 
 typedef struct ucc_mc_cuda {
-    ucc_mc_base_t super;
-    cudaStream_t  stream;
+    ucc_mc_base_t                  super;
+    cudaStream_t                   stream;
+    ucs_mpool_t                    events;
+    ucs_mpool_t                    strm_reqs;
+    ucc_mc_cuda_strm_task_mode_t   strm_task_mode;
```
Sergei-Lebedev, line 42: why do you need this field? it's already part of super config

src/components/mc/cuda/mc_cuda.c
```diff
@@ -183,9 +337,89 @@ static ucc_status_t ucc_mc_cuda_mem_query(const void *ptr,
...
+free_req:
+    ucs_mpool_put(req);
+    return status;
+}
+
+ucc_status_t ucc_ee_cuda_task_query(void *ee_req)
+{
+    ucc_mc_cuda_stream_request_t *req = ee_req;
+
+    if (req->status != UCC_INPROGRESS) {
```
Sergei-Lebedev, line 390: looks like condition is not correct

src/components/mc/cuda/mc_cuda.c
```diff
@@ -183,9 +337,89 @@ static ucc_status_t ucc_mc_cuda_mem_query(const void *ptr,
...
+    ucc_mc_cuda_event_t *cuda_event = (ucc_mc_cuda_event_t *)req->event;
+    cudaError_t cu_err;
+
+    req->status = UCC_OK;
+
+    if(cuda_event) {
+#if 1
+    do {
+        cu_err = cudaEventQuery(cuda_event->event);
+    } while (cu_err == cudaSuccess);
```
Sergei-Lebedev, line 409: why do we need to wait for event here?

src/core/ucc_mc.h
```diff
@@ -64,5 +70,6 @@ static inline ucc_status_t ucc_dt_reduce_multi(void *src1, void *src2, void *dst
     }
 }
 
+ucs_status_t ucc_mc_check_event_auto_triggered(ucc_ee_type_t ee_type, void *ee_context, void *ev_context);
```
Sergei-Lebedev, line 73: unused?

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,105 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+            return UCC_OK;
+        }
+    }
+
+    if (task->super.ee_task == NULL) {
+        status = ucc_mc_ee_task_post(task->super.ee->ee_context,
+                                     task->super.ee->ee_type, &task->super.ee_task);
+        if (status != UCC_OK) {
+            tl_error(task->team->super.super.context->lib, "error in ee task post");
+            return UCC_ERR_NOT_SUPPORTED;
```
Sergei-Lebedev, line 107: return status

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,105 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+        }
+
+        post_event->ev_type = UCC_EVENT_COLLECTIVE_POST;
+        post_event->ev_context_size = 0;
+        ucc_ee_set_event_internal(coll_task->ee, post_event, &coll_task->ee->event_out_queue);
+        task->super.ee->ee_task = task->super.ee_task;
+        task->super.super.status = UCC_OK;
+    }
+
+    return UCC_OK;
```
Sergei-Lebedev, line 130: perhaps it should return task->super.super.status

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,105 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+}
+
+static ucc_status_t ucc_tl_ucp_event_trigger_complete(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+
+    tl_info(task->team->super.super.context->lib, "event triggered. task:%p", coll_task);
+
+    coll_task->post(coll_task);
+    ucc_assert(coll_task->super.status = UCC_INPROGRESS);
```
Sergei-Lebedev, line 72: sometimes collective might complete immediately

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,114 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+                                                      ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_status_t status;
+
+    tl_info(task->team->super.super.context->lib, "event triggered. task:%p", coll_task);
+
+    coll_task->ee_task = parent_task->ee_task;
+    status = coll_task->post(coll_task);
+    if (status == UCC_OK) {
```
vspetrov, line 76: post return code does not mean completion. it only reports if error happend. if you need to check completion, you should check task status.

src/core/ucc_team.c
```diff
@@ -229,3 +237,110 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+{
+    ucc_event_desc_t *event_desc;
+    ucc_queue_elem_t *elem;
+
+    if (ucc_queue_is_empty(queue)) {
+        return UCC_ERR_NOT_FOUND;
+    }
+
+    ucc_spin_lock(&ee->lock);
+    elem = ucc_queue_pull(queue);
```
vspetrov, line 287: do you intend to support multithread there? if yes, then isn't this a race:
threads A and B check if ucc_queue_is_empty: there is 1 entry both threads see it, and proceed. Then thread A grabs the lock, takes the element from queue and releases the lock. Now queue is empty: thread B pulls from empty queue.

src/core/ucc_team.c
```diff
@@ -229,3 +237,110 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+}
+
+ucc_status_t ucc_ee_set_event(ucc_ee_h ee, ucc_ev_t *ev)
+{
+    return ucc_ee_set_event_internal(ee, ev, &ee->event_in_queue);
+}
+
+ucc_status_t ucc_ee_wait(ucc_ee_h ee, ucc_ev_t *ev)
+{
+    while(UCC_OK != ucc_ee_get_event(ee, &ev));
```
vspetrov, line 343: do you need any progress there?

src/core/ucc_team.c
bureddy: you are right. queue empty check also under spinlock

src/components/tl/ucp/tl_ucp_coll.c
bureddy: fixed

src/core/ucc_team.c
bureddy: fixed

src/core/ucc_team.c
```diff
@@ -229,3 +237,114 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+
+ucc_status_t ucc_ee_destroy(ucc_ee_h ee)
+{
+    ucc_spinlock_destroy(&ee->lock);
+    ucc_free(ee);
+
+    ucc_info("ucc_ee_create: ee is destroyed ");
+
+    return UCC_OK;
+}
```
manjugv, line 275: You might have to also check and destroy during team destroy.

manjugv (commented):
On component organization - did you try to handle the whole “ee” as a different component? Any obvious pros or cons? We are in the process of implementing/using this for other offload engines.

bureddy:
> On component organization - did you try to handle the whole “ee” as a different component? Any obvious pros or cons? We are in the process of implementing/using this for other offload engines.

I dont see a reason for to go for new component at least now. it can be simply extracted and can make different components if needed.  Maybe if we need to treat DPU also different memory domain, then may have DPU mc component as well.

src/components/mc/cuda/mc_cuda.c
```diff
@@ -36,7 +143,53 @@ static ucc_status_t ucc_mc_cuda_init()
             cfg->reduce_num_blocks = prop.maxGridSize[0];
         }
     }
-    CUDACHECK(cudaStreamCreate(&ucc_mc_cuda.stream));
+
+    CUDACHECK(cudaStreamCreateWithFlags(&ucc_mc_cuda.stream, cudaStreamNonBlocking));
```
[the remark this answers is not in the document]
bureddy: yes. library own stream.  This is still stream-based collective. but this is to where is the collective is executed (user stream (or) lib stream). if still create a dependency between the user stream and lib stream if the collective is issued on the lib stream

src/core/ucc_team.c
bureddy: you mean if user creates it but not destroyed it?

src/core/ucc_team.c
```diff
@@ -229,3 +237,114 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+    return ucc_ee_get_event_internal(ee, ev, &ee->event_out_queue);
+}
+
+ucc_status_t ucc_ee_ack_event(ucc_ee_h ee, ucc_ev_t *ev)
+{
+    ucc_event_desc_t *event_desc;
+
+    event_desc = ucc_container_of(ev, ucc_event_desc_t, ev);
+    /* TODO destroy event context */
+    ucc_free(event_desc);
```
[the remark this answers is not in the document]
bureddy: this event is get to only one thread who called ucc_ee_get_event. assuming only that thread will call ack_event

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,113 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+    ev_task->super.ev = NULL;
+    ev_task->super.flags = UCC_COLL_TASK_FLAG_INTERNAL;
+    ev_task->super.finalize = ucc_tl_ucp_coll_finalize;
+    ev_task->super.super.status = UCC_INPROGRESS;
+
+    ev_task->super.progress = ucc_tl_ucp_ee_wait_for_event_trigger;
+    ucc_event_manager_init(&ev_task->super.em);
+    coll_task->handlers[UCC_EVENT_COMPLETED] = ucc_tl_ucp_event_trigger_complete;
+    ucc_event_manager_subscribe(&ev_task->super.em, UCC_EVENT_COMPLETED, coll_task);
+    ucc_progress_enqueue(UCC_TL_UCP_TEAM_CORE_CTX(ev_task->team)->pq, &ev_task->super);
```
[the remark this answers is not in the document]
bureddy: this is task that waits for the event trigger

src/core/ucc_team.c
```diff
@@ -229,3 +237,114 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+        ucc_error("failed to allocate %zd bytes for ucc execution engine",
+                  sizeof(ucc_ee_t));
+        return UCC_ERR_NO_MEMORY;
+    }
+
+    ee->team = team;
+    ee->ee_type = params->ee_type;
+    ee->ee_context_size = params->ee_context_size;
+    ee->ee_context = params->ee_context;
+    ucc_spinlock_init(&ee->lock, 0);
```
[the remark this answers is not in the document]
bureddy: need a if condition is to check thread mode in that case, spinlock is mostly an atomic operation. it should not be costly if there no threads.  what approach do you think better?

src/core/ucc_team.c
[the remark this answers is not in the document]
manjugv: Currently, we did not impose that restriction. I'm wondering, if it really required to place such restriction on the user.

src/core/ucc_team.c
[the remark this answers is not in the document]
manjugv: I agree on the performance. Probably, we can enable this only on multithread modes (UCC_THREAD_MULTIPLE).

src/core/ucc_team.c
manjugv: Either library or user has to do the book keeping of allocated ee's for a given team. 

We can address this two ways: 
1) Place responsibility on the user to destroy the ee before the team destroy.
2)Library destroying the ee connected to a team, before team destroy operation.

src/components/mc/cuda/mc_cuda.h
```diff
@@ -9,19 +9,55 @@
...
+    UCC_MC_CUDA_TASK_MEM_OPS,
+    UCC_MC_CUDA_TASK_AUTO,
+} ucc_mc_cuda_strm_task_mode_t;
+
+typedef enum ucc_mc_cuda_task_stream_type {
+    UCC_MC_CUDA_USER_STREAM,
+    UCC_MC_CUDA_INTERNAL_STREAM
+} ucc_mc_cuda_task_stream_type_t;
+
+enum ucc_mcc_task_status {
```
vspetrov, line 26: typedef and use ucc_mc_task_status ?

src/components/mc/cuda/mc_cuda.c
```diff
@@ -183,9 +338,120 @@ static ucc_status_t ucc_mc_cuda_mem_query(const void *ptr,
...
+    return UCC_OK;
+
+free_event:
+    ucs_mpool_put(cuda_event);
+free_req:
+    ucs_mpool_put(req);
+    return status;
+}
+
+ucc_status_t ucc_ee_cuda_task_query(void *ee_req)
```
vspetrov, line 386: why all interfaces with "void" req? Usually we would define ucc_ee_req_t "super" request (i guess some ucc_ee_status could be there) and components would extend it and use "ucc_derived_of". I know it is only a visual difference but it would be more consistent with the rest of the code.

src/components/mc/cuda/mc_cuda.c
```diff
@@ -183,9 +338,120 @@ static ucc_status_t ucc_mc_cuda_mem_query(const void *ptr,
...
+    CUDACHECK(cudaEventRecord((cudaEvent_t)event, stream));
+    return UCC_OK;
+}
+
+ucc_status_t ucc_ee_cuda_event_test(void *event)
+{
+    cudaError_t cu_err;
+
+    cu_err = cudaEventQuery((cudaEvent_t)event);
+    switch(cu_err) {
```
vspetrov, line 440: maybe add cuda_error_to_ucc_status helper inline (either using this kind of switch or array map) ?

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,117 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+            task->super.ev = (ucc_ev_t *) 0xFFFF; /* dummy event */
+            task->super.ee_task = NULL;
+        } else if (UCC_OK == ucc_ee_get_event_internal(task->super.ee, &ev,
+                                                &task->super.ee->event_in_queue)) {
+            tl_info(task->team->super.super.context->lib,
+                    "triggered event arrivied. task:%p", coll_task);
+            task->super.ev = ev;
+            task->super.ee_task = NULL;
+        } else {
+            return UCC_OK;
```
vspetrov, line 111: is this return reachable?

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,117 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+
+static ucc_status_t ucc_tl_ucp_event_trigger_complete(ucc_coll_task_t *parent_task,
+                                                      ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+
+    tl_info(task->team->super.super.context->lib, "event triggered. task:%p", coll_task);
+
+    coll_task->ee_task = parent_task->ee_task;
+    coll_task->post(coll_task);
```
vspetrov, line 74: ccheck return of post - if not UCC_OK - then it failed

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,117 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+    if (coll_task->super.status == UCC_OK) {
+        return ucc_tl_ucp_triggered_coll_complete(coll_task, coll_task);
+    } else {
+
```
vspetrov, line 78: empty lines

src/core/ucc_team.c
vspetrov: But then: do we also need to do the same from collective request finalize? It is also user responsibility to call finalize. At the same time we might do a bookeeping of requests allocated/started per team. We can at least keep counting (initialized, posted, finalize) and upon team destruction through a warn if initialized != finalized.

src/core/ucc_team.c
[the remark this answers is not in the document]
vspetrov: Where is the restriction? When user gets the event - we have a locking. Then user gets the even and "some" thread (only 1 thread) will call event ack (currently just ucc_free). We don't access any global struct here. Looks like nothing to protect with lock.

src/core/ucc_team.c
vspetrov: maybe even better with "double evaluation": first check queue is epmty w/o the lock. then go to your lines 282-286, ie. check again with the lock. This way you can amortize locking when queue is empty. Ideally we should think of some better lock-free queue or smth: this looks very similar to our progress queue, we might have a contention there. But this is for later ..

src/schedule/ucc_schedule.h
```diff
@@ -17,23 +17,34 @@ typedef enum {
...
 
+enum {
+    UCC_COLL_TASK_FLAG_INTERNAL = UCC_BIT(0)
+};
+
 typedef struct ucc_coll_task {
     ucc_coll_req_t             super;
+    uint32_t                   flags;
     ucc_coll_post_fn_t         post;
+    ucc_coll_triggered_post_fn_t triggered_post;
```
vspetrov, line 39: align

src/core/ucc_team.c
```diff
@@ -229,3 +237,114 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+
+    event_desc = ucc_container_of(elem, ucc_event_desc_t, queue);
+    *ev = &event_desc->ev;
+
+    ucc_info("EE Event Get. ee:%p, queue:%p ev_type:%s ",
+                ee, queue, ucc_ee_ev_names[event_desc->ev.ev_type]);
+    return UCC_OK;
+}
+
+ucc_status_t ucc_ee_get_event(ucc_ee_h ee, ucc_ev_t **ev)
```
vspetrov, line 301: i think all these functions should go to core/ucc_ee.c, not into team.c

src/components/mc/cuda/mc_cuda.c
```diff
@@ -9,6 +9,17 @@
 #include <cuda_runtime.h>
 #include <cuda.h>
 
+static const char *strm_task_modes[] = {
```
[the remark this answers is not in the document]
bureddy: these are task modes instead of stream.  I will change it to "stream_task_modes"

src/components/mc/cuda/mc_cuda.c
```diff
@@ -183,9 +338,120 @@ static ucc_status_t ucc_mc_cuda_mem_query(const void *ptr,
...
+    req->status = UCC_MC_CUDA_TASK_POSTED;
+    req->stream = (cudaStream_t)ee_stream;
+
+    if (ucc_mc_cuda.task_strm_type == UCC_MC_CUDA_USER_STREAM) {
+        status = ucc_mc_cuda.post_strm_task(req->dev_status, req->stream);
+        if (status != UCC_OK) {
+            goto free_req;
+        }
+    } else {
+        cuda_event = ucs_mpool_get(&ucc_mc_cuda.events);
```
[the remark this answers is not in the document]
bureddy: not really in performance wise. but it might help in some debug cases to see only this stream in profiles.

src/components/tl/ucp/tl_ucp_coll.c
bureddy: yes if ucc_ee_get_event_internal fails because event not posted by user in above if condition

src/core/ucc_team.c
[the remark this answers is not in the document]
manjugv: In multithreaded case, there can be race between get and ack event, if called from two different threads. Either we should impose a restriction saying that get and ack should be called from the same thread or protect it.

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,123 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+        } else {
+            return UCC_OK;
+        }
+    }
+
+    if (task->super.ee_task == NULL) {
+        status = ucc_mc_ee_task_post(task->super.ee->ee_context,
+                                     task->super.ee->ee_type, &task->super.ee_task);
+        if (status != UCC_OK) {
+            tl_error(task->team->super.super.context->lib, "error in ee task post");
```
Sergei-Lebedev, line 119: also need to set task status
task->super.super.status = status

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -51,6 +53,123 @@ static ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+
+        /* TODO: mpool */
+        post_event = ucc_malloc(sizeof(ucc_ev_t), "event");
+        if (post_event == NULL) {
+            tl_error(task->team->super.super.context->lib,
+                     "failed to allocate memory for event");
+            return UCC_ERR_NO_MEMORY;
+        }
+
+        post_event->ev_type = UCC_EVENT_COLLECTIVE_POST;
```
Sergei-Lebedev, line 131: set post_event->req

bureddy: fixed

src/components/tl/ucp/tl_ucp_coll.c
bureddy: fixed

src/core/ucc_team.c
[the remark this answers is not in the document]
bureddy: not touching any global structs in ack event.

src/core/ucc_team.c
bureddy: fixed

Sergei-Lebedev (approved):

vspetrov (approved):

manjugv (approved):

PR #96, opened 2021-03-09
TEST: added mpi alltoall[v] test

karasevb:

PR #105, opened 2021-03-16
Test mpi fixes

vspetrov:
## What
TEST/MPI:
- Fixes random root generation (use local random engine instead of rand())
- Fixes hang due to lack of MPI progress (need to move MPI runtime when waiting for ucc collective)
- Adds broadcast to the test list
- Enables INPLACE + NO_INPLACE


vspetrov:
Closing, since we these changes are ported into #96

PR #110, opened 2021-03-18
UCC Team ID allocation

vspetrov:
## What
 - Service Allreduce TL interface
 - TL/UCP service Allreduce
 - Service tl/ucp team creation in core
 - Internal collective team id allocation algorithm using service allreduce
 - Option to get team id from ucc_team_params_t when provided by user

## Why ?
Team id is necessary for tag multiplexing in tl/ucp (and potentially in other TLs)

## How ?
*** Team Id allocation algorithm
Each ucc_context contains id pool: array of uint64_t values. This size of the array is controlled with the env var parameter. It defines the max id that can be allocated (hence it defines max number of coexisting teams). Each bit in the id pool represents the ID : from 1 to pool_size*64. 
The algorithm: the whole pool is initialized to 1 in the beginning. Ranks of the team perform allreduce with BAND op over pool bits and then search for first bit set -> this will be the id. The bit is then cleared, so that on next allreduce the rank will report the ID is occupied.

*** Service coll iface
Right now TL only defines allreduce. It is different from the default coll interface. It has additional parameter: subset. It defines a subset of processes of the TL team  that want execute a collective. This is not used right now for the team id allocation (all ranks participate there) but it will be used in hierarchical CL to allocate subgroup teams.


src/utils/ucc_math.h
```diff
@@ -22,4 +22,26 @@ static inline size_t ucc_dt_size(ucc_datatype_t dt)
...
+        break;
+    case UCC_EP_MAP_CB:
+        r = (ucc_rank_t)map.cb.cb(rank, map.cb.cb_ctx);
+        break;
+    default:
+        r = -1;
+    }
+    return r;
+}
+
```
bureddy, line 46: seems not appropriate for generic core EP hepler func in utils_math header

src/core/ucc_team.c
```diff
@@ -121,8 +125,14 @@ ucc_status_t ucc_team_create_post(ucc_context_h *contexts, uint32_t num_contexts
         status = UCC_ERR_NO_MEMORY;
         goto err_ctx_alloc;
     }
+
     memcpy(team->contexts, contexts, sizeof(ucc_context_t *) * num_contexts);
     ucc_copy_team_params(&team->params, params);
+    /* check if user provides team id and if it is not too large */
+    if ((params->mask & UCC_TEAM_PARAM_FIELD_ID) &&
+        (params->id <= UCC_TEAM_ID_MAX)) {
```
bureddy, line 133: why is param->id is 64 bit and internal ID is 16 bit? what happens if user creates two teams with MSB bit sets (LSB 15 bits 0)?

vspetrov: Currently we have a limitation on internal team id: 16 bits, due to UCP tag space limitation. In the API we decided to keep it uint64_t since maybe we will be able to support larger IDs in the future. Here we check if the value provided by the user can be actually used. If MSB bits are set then params->id > UCC_TEAM_ID_MAX. We will not use it: we will allocate team id internally. Only when params->id fits into 15 bits, we will take it (e.g., in hpcx in ompi this is the case for many many workloads).

src/core/ucc_team.c
```diff
@@ -229,3 +297,105 @@ ucc_status_t ucc_team_destroy(ucc_team_h team)
...
+}
+
+static void ucc_team_relase_id(ucc_team_t *team)
+{
+    ucc_context_t *ctx = team->contexts[0];
+    /* release the id pool bit if it was not provided by user */
+    if (!UCC_TEAM_ID_IS_EXTERNAL(team)) {
+        set_id_bit(ctx->ids.pool, team->id);
+    }
+}
```
manjugv, line 401: Val, I have been thinking about this algorithm. From what I see, this algorithm can cause a live-lock in a multithreaded case. In UCC’s case, we avoid this because we don’t support multiple team posts. Is my current statement correct? If so, should we warn somewhere. Is there an edge case, where we can still cause livelock?

manjugv (approved):

src/utils/ucc_math.h
vspetrov: @bureddy will move it to a separate header under under utils + will remove definitions of ep mappers from ucc.h.

bureddy (approved):

Issue #108, opened 2021-03-17
UCC team create/destroy in a loop leaks memory due to UCP


vspetrov:
UCC code like below:
```
ucc_init
ucc_context_create->context
for i in [1..100]:
    ucc_team_create(context)->team
    ucc_alltoall(team)
    ucc_team_destroy(team)
```
leaks memory when executed using TL/UCP. The reason for that is that each time we create a team with TL/UCP we allocate UPC endpoints. The call to alltoall forces ALL eps to connect (N^2 total connections). The call to ucc_team_destroy will result in ucp_ep_close(FLUSH) being executed on each local endpoint. HOWEVER: for some transports (rc is one of them) only local part of the p2p ep (qp) will be cleaned up by UCP - ie. the corresponding part of the connection stored on the remote ucp worker is left and consumes memory -> leak. Attached is a stand-alone UCP reproduced for the issue.
[ucp_test.txt](https://github.com/openucx/ucc/files/6159016/ucp_test.txt)

Looks like it is known limitation and we can't expect it to be resolved in UCP soon.

What can we do in UCC about it?
1. First: short term solution.
Use hashing and store UCP endpoints in the global hashtable. Requires uniq participant ID generation. I will prepare a prototype PR.
2. Long term:
For some runtimes (mostly i'm talking about MPI) we want to optimize. We have a notion of "WORLD" there, so we can store global endpoints array and re-map team ranks there. The first approach was made in PR #77. This implies global (collective) ucc context AND context rank notion, which we would like to avoid. Another approach could be: use comm_world ucc team and alternative interface for team creation.

PR #111, opened 2021-03-19
TL/UCP: endpoints hash

vspetrov:
## What
Changes the way ucp ep are stored in TL. Now we use a hash table located on ucc_tl_ucp_contet_t. The hash implementation is a "khash" - opensource free license MIT implementation also used by UCX. 

## Why ?
Storing eps on team results in huge memory leak (#108) and team create overhead doe to connect-disconnect

## How ?
The key to the hash table is ucc_context_id_t structure which consists of a triplet: host_id, pid, and ucc context sequence number.
TL/UCP exchanges ucp_worker addresses + the keys during team creation (exchange during ctx create or PMIx-like iface is not yet implemented). The ep is then created only once and put into hash based on the key.

OVerheads:
1. i've measured the cost of the hash itself (implemented a simple benchmark):
    1.1 for the size of the hash ~1000 the overhead to find the ep in the has is ~0.02 us (<2% compared to network p2p cost)
     1.2 for the size of the hash ~10000 it is ~0.03 us.
     1.3 for the size of the hash ~1000000 it is ~0.08 us.
2. Eval on the Collectives (see xls attached 
[ep_hash_eval.xlsx](https://github.com/openucx/ucc/files/6171030/ep_hash_eval.xlsx). Shortly, for the allreduce, the difference ~5% is observed on 1 node, ~3% on 2 nodes, and nearly 0 difference on 8 nodes. And of course it can only be observed on small messages.

All this makes the approach feasible imho. THe overheads are minimal. Moreover, we can optimize smallest colls (barrier, and small msg ar) by pre-allocating the alg data structure with EPs (doing hash access once during team create) - this will be made in another PR.

### DIFF
--- a/src/components/tl/ucp/tl_ucp.h
+++ b/src/components/tl/ucp/tl_ucp.h
@@ -9,7 +9,7 @@
 #include "components/tl/ucc_tl.h"
 #include "components/tl/ucc_tl_log.h"
 #include "utils/ucc_mpool.h"
-
+#include "tl_ucp_ep_hash.h"
 #include <ucp/api/ucp.h>
 #include <ucs/memory/memory_type.h>
 
@@ -57,10 +57,11 @@ typedef struct ucc_tl_ucp_context {
     ucp_worker_h                ucp_worker;
     size_t                      ucp_addrlen;
     ucp_address_t              *worker_address;
+    ucp_ep_h                   *eps;
     ucc_tl_ucp_ep_close_state_t ep_close_state;
     ucc_mpool_t                 req_mp;
-    ucp_ep_h                   *eps;
     ucc_tl_ucp_addr_storage_t  *addr_storage;
+    tl_ucp_ep_hash_t           *ep_hash;
 } ucc_tl_ucp_context_t;
 UCC_CLASS_DECLARE(ucc_tl_ucp_context_t, const ucc_base_context_params_t *,
                   const ucc_base_config_t *);
@@ -69,15 +70,10 @@ typedef struct ucc_tl_ucp_task ucc_tl_ucp_task_t;
 typedef struct ucc_tl_ucp_team {
     ucc_tl_team_t              super;
     ucc_status_t               status;
-    int                        context_ep_storage; /*< The flag
-              indicates whether ucp endpoints are stored on the
-              ucc_tl_ucp_context or are they created per-team.
-              This optimization is only possible when user provides
-              the necessary rank mappings team_rank->context_rank. */
-    ucp_ep_h                  *eps;
     ucc_rank_t                 size;
     ucc_rank_t                 rank;
     ucc_tl_ucp_addr_storage_t *addr_storage;
+    int                        ctx_addr_exchange;
     uint32_t                   id;
     uint32_t                   scope;
     uint32_t                   scope_id;
--- a/src/components/tl/ucp/tl_ucp.h
+++ b/src/components/tl/ucp/tl_ucp.h
@@ -46,7 +46,6 @@ UCC_CLASS_DECLARE(ucc_tl_ucp_lib_t, const ucc_base_lib_params_t *,
 typedef struct ucc_tl_ucp_addr_storage ucc_tl_ucp_addr_storage_t;
 
 typedef struct ucc_tl_ucp_ep_close_state {
-    ucc_rank_t ep;
     void      *close_req;
 } ucc_tl_ucp_ep_close_state_t;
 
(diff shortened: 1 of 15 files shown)


manjugv (approved):

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -126,7 +129,12 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
 
 UCC_CLASS_CLEANUP_FUNC(ucc_tl_ucp_context_t)
 {
+    ucc_status_t status;
     tl_info(self->super.super.lib, "finalizing tl context: %p", self);
+    while (UCC_OK != (status = ucc_tl_ucp_close_eps(self))) {
+        ; //TODO can we hang the runtime this way ?
+    }
+	kh_destroy(tl_ucp_ep_hash, self->ep_hash);
```
Sergei-Lebedev, line 137: alignment

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -29,47 +30,54 @@ static inline ucc_status_t ucc_tl_ucp_connect_ep(ucc_tl_ucp_context_t *ctx,
...
+        ep = tl_ucp_hash_pop(ctx->ep_hash);
+        if (NULL == ep) {
+            break;
         }
         state->close_req =
-            ucp_ep_close_nb(eps[state->ep], UCP_EP_CLOSE_MODE_FLUSH);
+            ucp_ep_close_nb(ep, UCP_EP_CLOSE_MODE_FLUSH);
         if (UCS_PTR_IS_ERR(state->close_req)) {
             tl_error(ctx->super.super.lib, "failed to start ep close, ep %p",
-                     eps[state->ep]);
```
Sergei-Lebedev, line 72: looks like ep field is not needed in ucc_tl_ucp_ep_close_state_t  anymore, should we delete it?

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -29,47 +30,54 @@ static inline ucc_status_t ucc_tl_ucp_connect_ep(ucc_tl_ucp_context_t *ctx,
...
             return UCC_INPROGRESS;
         }
         ucp_request_free(state->close_req);
-        eps[state->ep++] = NULL;
     }
-
-    for (; state->ep < n_eps; state->ep++) {
-        if (!eps[state->ep]) {
-            continue;
+    while (1) {
```
Sergei-Lebedev, line 71: minor, I would better write it without break, something like
```C
ep = tl_ucp_hash_pop(ctx->ep_hash);
while(ep) {
...
  ep = tl_ucp_hash_pop(ctx->ep_hash);
}
```

src/components/tl/ucp/tl_ucp.h
```diff
@@ -69,15 +70,10 @@ typedef struct ucc_tl_ucp_task ucc_tl_ucp_task_t;
...
-    int                        context_ep_storage; /*< The flag
-              indicates whether ucp endpoints are stored on the
-              ucc_tl_ucp_context or are they created per-team.
-              This optimization is only possible when user provides
-              the necessary rank mappings team_rank->context_rank. */
-    ucp_ep_h                  *eps;
     ucc_rank_t                 size;
     ucc_rank_t                 rank;
     ucc_tl_ucp_addr_storage_t *addr_storage;
+    int                        ctx_addr_exchange;
```
Sergei-Lebedev, line 76: what is the meaning of ctx_addr_exchange? is it used?

src/components/tl/ucp/tl_ucp_ep.h
```diff
@@ -10,34 +10,45 @@
...
 ucc_status_t ucc_tl_ucp_connect_ctx_ep(ucc_tl_ucp_context_t *ctx, ucc_rank_t ctx_rank);
 
-ucc_status_t ucc_tl_ucp_close_eps(ucc_tl_ucp_context_t *ctx, ucp_ep_h *eps,
-                                  ucc_rank_t n_eps);
+ucc_status_t ucc_tl_ucp_close_eps(ucc_tl_ucp_context_t *ctx);
+
+static inline ucc_context_id_t
+ucc_tl_ucp_get_rank_key(ucc_tl_ucp_team_t *team, ucc_rank_t rank)
+{
+    size_t max_addrlen = team->addr_storage->max_addrlen;
```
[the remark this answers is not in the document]
vspetrov: :+1 fixed

src/components/tl/ucp/tl_ucp.h
vspetrov: It was old flow (potentially eps could be stored on CTX) - removed that now. Maybe we will still implement exchange of addresses during ctx create (in global mode) but later.

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -126,7 +129,12 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
 
 UCC_CLASS_CLEANUP_FUNC(ucc_tl_ucp_context_t)
 {
+    ucc_status_t status;
     tl_info(self->super.super.lib, "finalizing tl context: %p", self);
+    while (UCC_OK != (status = ucc_tl_ucp_close_eps(self))) {
```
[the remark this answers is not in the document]
vspetrov: 👍

src/components/tl/ucp/tl_ucp_ep.c
vspetrov: 👍

Sergei-Lebedev (approved):

PR #112, opened 2021-03-22
Topic/mpi test fixes

vspetrov:
TL/UCP: kn bcast del incorrect assert
        - Create teams after parsing args
        - Properly set TestCase::test_skip after global allreduce
        - Bcast skip should not depend on comm size
        - Alltoallv proper 64 bit dsiplacements cleanup in case of skip


test/mpi/main.cc
```diff
@@ -391,8 +391,8 @@ int main(int argc, char *argv[])
     int rank;
     int failed = 0;
 
-    UccTestMpi test(argc, argv, UCC_THREAD_SINGLE, teams, cls);
     ProcessArgs(argc, argv);
+    UccTestMpi test(argc, argv, UCC_THREAD_SINGLE, teams, cls);
```
Sergei-Lebedev, line 395: In that order mtype_str_to_type will not work for cuda buffers, because it uses ucc_mc_available function

vspetrov: fixed by creating teams explicitly after parsing args

bureddy (approved):

Sergei-Lebedev (approved):

PR #121, opened 2021-03-29
TEST: mpi: added set CUDA device

karasevb:

### DIFF
--- a/test/mpi/main.cc
+++ b/test/mpi/main.cc
@@ -303,7 +303,7 @@ void PrintInfo()
               << std::endl;
 }
 
-void ProcessArgs(int argc, char** argv)
+int ProcessArgs(int argc, char** argv)
 {
     const char *const short_opts  = "c:t:m:d:o:M:I:r:s:C:D:i:Z:Th";
     const option      long_opts[] = {
@@ -379,10 +379,10 @@ void ProcessArgs(int argc, char** argv)
         case 'h': // -h or --help
         case '?': // Unrecognized option
         default:
-            PrintHelp();
-            break;
+            return 1;
         }
     }
+    return 0;
 }
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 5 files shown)


test/mpi/main.cc
```diff
@@ -383,34 +397,103 @@ void ProcessArgs(int argc, char** argv)
     }
 }
 
+#ifdef HAVE_CUDA
+void cuda_set_device()
+{
+    MPI_Comm local_comm;
+    int cuda_dev_count;
+    int local_rank;
+    MPI_Comm_split_type(MPI_COMM_WORLD, OMPI_COMM_TYPE_NODE, 0, MPI_INFO_NULL,
```
vspetrov, line 406: this is OMPI specific. Is there an MPI-spec way to do it? (so that it works e.g. with MPICH). We use MPI only to bootstrap ranks here, so any should work imho.

test/mpi/main.cc
```diff
@@ -383,34 +397,103 @@ void ProcessArgs(int argc, char** argv)
...
+                        &local_comm);
+    MPI_Comm_rank(local_comm, &local_rank);
+    CUDA_CHECK(cudaGetDeviceCount(&cuda_dev_count));
+    if(local_rank >= cuda_dev_count) {
```
vspetrov, line 410: Sometimes maybe it worth setting device = (local_rank % cuda_dev_count). Maybe make --set-device option with paramter: 0 - don't set, 1 - set device=local_rank and error if local_rank >= cuda_dev_count and 2 - device = local_rank % cuda_dev_count.

test/mpi/main.cc
```diff
@@ -383,34 +397,103 @@ void ProcessArgs(int argc, char** argv)
...
+        /* .coll_types = coll_types, */
+    };
+
+    /* Init ucc context for a specified UCC_TEST_TLS */
+    ucc_context_params_t ctx_params = {
+        .mask     = UCC_CONTEXT_PARAM_FIELD_TYPE,
+        .ctx_type = UCC_CONTEXT_EXCLUSIVE,
+    };
+    UCC_CHECK(ucc_lib_config_read(NULL, NULL, &lib_config));
+    UCC_CHECK(ucc_init(&lib_params, lib_config, &lib));
```
vspetrov, line 448: Need to rebase carefully: CONTEXT_EXCLUSIVE is not used any more (for multithreaded mode); 
Why you extracted all these UCC setup out of test constructor?
In master ParseArgs currently is made BEFORE ucc_test, so that all parameters are available for constructor.

test/mpi/main.cc
karasevb: Should be OK with MPICH as well: https://www.mpich.org/static/docs/v3.2/www3/MPI_Comm_split_type.html

test/mpi/main.cc
karasevb: fixed

test/mpi/main.cc
karasevb: fixed

vspetrov (approved):

test/mpi/main.cc
```diff
@@ -376,13 +383,18 @@ void ProcessArgs(int argc, char** argv)
         case 'i':
             iterations = std::stoi(optarg);
             break;
+#ifdef HAVE_CUDA
+        case 'S':
+            test_cuda_set_device = (test_set_cuda_device_t)std::stoi(optarg);
+            break;
+#endif
```
Sergei-Lebedev, line 390: minor, add #else for HAVE_CUDA and print something like "test wasn't compiled with cuda support"

Sergei-Lebedev (approved):

test/mpi/main.cc
karasevb: This option will be disabled If test compiled without CUDA, so the user cannot specify this. Do you want the option should be visible anyway?

test/mpi/test_mpi.cc
```diff
@@ -39,25 +39,9 @@ static ucc_status_t oob_allgather_free(void *req)
 }
 
 UccTestMpi::UccTestMpi(int argc, char *argv[], ucc_thread_mode_t _tm, int is_local) {
-    int required = (_tm == UCC_THREAD_SINGLE) ? MPI_THREAD_SINGLE
-        : MPI_THREAD_MULTIPLE;
-    int size, provided;
```
vspetrov, line 44: @karasevb have tries to compile after rebase? Looks like you removed "size" which is used below in ctx params?

karasevb: fixed

test/mpi/main.cc
Sergei-Lebedev: ok

test/mpi/test_mpi.cc
```diff
@@ -247,6 +233,38 @@ void UccTestMpi::set_displ_vsizes(std::vector<ucc_test_vsize_flag_t> &_displs_vs
...
 }
 
+#ifdef HAVE_CUDA
+void UccTestMpi::set_cuda_device(test_set_cuda_device_t set_device)
+{
+    MPI_Comm local_comm;
+    int cuda_dev_count;
+    int local_rank;
+    int device_id;
+    MPI_Comm_split_type(MPI_COMM_WORLD, OMPI_COMM_TYPE_NODE, 0, MPI_INFO_NULL,
```
vspetrov, line 243: plz change OMPI_COMM_TYPE_NODE to MPI_COMM_TYPE_SHARED, and 0 -> comm_world rank. THis way it will be portable across MPIs and not depend on OMPI (UCC is MPI agnostic)

vspetrov (changes requested):

vspetrov (approved):

vspetrov:
@Sergei-Lebedev Plz have a look again at cuda reduce Kernel return code check

test/mpi/test_mpi.cc
karasevb: Fixed

Sergei-Lebedev:
> @Sergei-Lebedev Plz have a look again at cuda reduce Kernel return code check

I would better move it from macro to actual function, just before stream synchronize.

vspetrov:
> > @Sergei-Lebedev Plz have a look again at cuda reduce Kernel return code check
> 
> 
> 
> I would better move it from macro to actual function, just before stream synchronize.

@karasevb plz apply

vspetrov (approved):

vspetrov:
@karasevb plz cherry-pick to v0.1.x

PR #124, opened 2021-03-30
Topic/tl ucp ep close sync

vspetrov:
## What
If OOB is provided for UCC context then we should launch a barrier after endpoints have been closed. Otherwise, we have to create ucp ep endpoints with ERR_HANDLER (to avoid segv during ep_flush). This may results in worse perf do to suboptimal ucp ep lanes selection.

## Why ?
W/o the fix segv is likely to happen in ucc_tl_ucp_context_destroy

### DIFF
--- a/src/components/tl/ucp/tl_ucp_ep.c
+++ b/src/components/tl/ucp/tl_ucp_ep.c
@@ -2,6 +2,12 @@
 #include "tl_ucp_ep.h"
 #include "tl_ucp_addr.h"
 
+static void ucc_tl_ucp_err_handler(void *arg, ucp_ep_h ep, ucs_status_t status)
+{
+    /* Dummy fn - we don't expect errors in disconnect flow */
+    ;
+}
+
 static inline ucc_status_t ucc_tl_ucp_connect_ep(ucc_tl_ucp_context_t *ctx,
                                                  ucp_ep_h *ep, char *addr_array,
                                                  size_t max_addrlen, ucc_rank_t rank)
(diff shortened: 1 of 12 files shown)


vspetrov:
@Sergei-Lebedev @manjugv @bureddy plz review. made according to our today's discussion.

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -17,6 +23,13 @@ static inline ucc_status_t ucc_tl_ucp_connect_ep(ucc_tl_ucp_context_t *ctx,
     ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
     ep_params.address    = (ucp_address_t*)address->addr;
 
+    if (!UCC_TL_CTX_HAS_OOB(ctx)) {
+        ep_params.err_mode        = UCP_ERR_HANDLING_MODE_PEER;
+        ep_params.err_handler.cb  = ucc_tl_ucp_err_handler;
+        ep_params.err_handler.arg = NULL;
+        ep_params.field_mask     |= UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE |
+                                    UCP_EP_PARAM_FIELD_ERR_HANDLER;
+    }
```
manjugv, line 32: @vspetrov Should we warn the user about non-optimal behavior?

manjugv (approved):
Minor comment. Otherwise, it looks fine.

Sergei-Lebedev (approved):

src/components/tl/ucp/tl_ucp_ep.c
vspetrov: @manjugv i think not, we might add some note in the manual or readme

PR #136, opened 2021-04-02
MC/CUDA: Option for non-blocking stream task.

bureddy:


Sergei-Lebedev (approved):

vspetrov (approved):
need to be properly merged with #137

PR #140, opened 2021-04-05
Fix linter warnings

vspetrov:
Actual issues are fixed, unused vars are suppressed with NOLINT

### DIFF
--- a/src/coll_score/ucc_coll_score.c
+++ b/src/coll_score/ucc_coll_score.c
@@ -414,7 +414,7 @@ out:
 
 static ucc_status_t ucc_coll_score_parse_str(const char       *str,
                                              ucc_coll_score_t *score,
-                                             ucc_rank_t        team_size)
+                                             ucc_rank_t        team_size)//NOLINT
 {
     ucc_status_t       status  = UCC_OK;
     ucc_coll_type_t   *ct      = NULL;
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -136,7 +136,7 @@ err_cfg:
 static void ucc_tl_ucp_context_barrier(ucc_tl_ucp_context_t *ctx,
                                        ucc_context_oob_coll_t *oob)
 {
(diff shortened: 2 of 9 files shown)


src/coll_score/ucc_coll_score.c
```diff
@@ -414,7 +414,7 @@ static ucc_status_t str_to_msgranges(const char *str, size_t **ranges,
 
 static ucc_status_t ucc_coll_score_parse_str(const char       *str,
                                              ucc_coll_score_t *score,
-                                             ucc_rank_t        team_size)
+                                             ucc_rank_t        team_size)//NOLINT
```
manjugv, line 417: @vspetrov Adding NOLINT directive to the code seems like distraction? Is there a different way that is less distractive to add the directives?

vspetrov: I dont know another way. Those are "unused vars" in the functions that are not fully implemented. We either need to suppress these false positives, OR remove check for unused vars from Linter. @manjugv @Sergei-Lebedev thoughts?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -136,7 +136,7 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
 static void ucc_tl_ucp_context_barrier(ucc_tl_ucp_context_t *ctx,
                                        ucc_context_oob_coll_t *oob)
 {
-    char        *rbuf = ucc_malloc(sizeof(char*) * oob->participants,
+    char        *rbuf = ucc_malloc(sizeof(char) * oob->participants,
```
Sergei-Lebedev, line 139: also fix tl_error sizeof(char*) * oob->participants if failed to allocate rbuf

Sergei-Lebedev (approved):

src/coll_score/ucc_coll_score.c
Sergei-Lebedev: I think NOLINT here is  just another reminder that something is not implemented, so it looks fine to me

vspetrov:
@manjugv @bureddy plz have a look

bureddy (approved):

PR #142, opened 2021-04-05
CORE: change COLL_SCORE separator

vspetrov:
Changes separator character in UCC_*_COLL_SCORE parsing from ";"
to "#". The reason is that some runtimes (MPIs, e.g. OpenMPI)
incorrectly parse ";" token. W/o that change one has to always use
quote marks to guard the string.

This is just minor usability improvement.

### DIFF
--- a/src/coll_score/ucc_coll_score.c
+++ b/src/coll_score/ucc_coll_score.c
@@ -496,7 +496,7 @@ ucc_status_t ucc_coll_score_alloc_from_str(const char *str,
     if (UCC_OK != status) {
         return status;
     }
-    tokens = ucc_str_split(str, ";");
+    tokens = ucc_str_split(str, "#");
     if (!tokens) {
         status = UCC_ERR_INVALID_PARAM;
         goto error;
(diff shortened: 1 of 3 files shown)


manjugv (approved):

vspetrov:
After short discussion with Sergey we think we want more input from UCC people. Should we keep it ";" and then user will need to guard the string with "" in some runtimes (e.g. OMPI). Or should we change , like in this PR. This is a very minor readability question. My preference to change to # and don't think about quotes. what does other people think?

bureddy:
will  `","` work?

vspetrov:
> will `","` work?

no, "," is already in use. Example cmd line: UCC_TL_UCP_COLL_SCORE=allreduce,barrier:cuda,host:0-1M:inf;bcast,alltoall:0
comma is used a separator in multiple qualifiers

vspetrov:
btw, since we are discussing this now: currently the parameter is called UCC_CL/TL_*_COLL_SCORE. Is "COLL" redundant" here? maybe just UCC_TL_UCP_SCORE?

alex--m (approved):

manjugv:
> btw, since we are discussing this now: currently the parameter is called UCC_CL/TL_*_COLL_SCORE. Is "COLL" redundant" here? maybe just UCC_TL_UCP_SCORE?

I think UCC_TL_UCP_SCORE makes sense.

PR #147, opened 2021-04-06
CORE: Fix bug in MT progress queue

lappazos:
## What
Fix bug in MT progress queue

## Why ?
Hang while using mt progress queue after PR 125

## How ?
Create locked queue per pool instead of global. Also, organize the structs in a more accurate way.
In addition, remove counters (seems like only hure performance) - add unnecessary actions per call. When empty, dequeue func simply finish the iteration.

### DIFF
--- a/src/core/ucc_progress_queue_mt.c
+++ b/src/core/ucc_progress_queue_mt.c
@@ -8,18 +8,12 @@
 #include "utils/ucc_malloc.h"
 #include "utils/ucc_log.h"
 #include "utils/ucc_spinlock.h"
-#include "utils/ucc_atomic.h"
-
-// Number of tasks in a single lock free pool - could be changed, but in tests performed great
-#define LINE_SIZE 8
-#define NUM_POOLS 2 // We need exactly two pools
+#include "utils/ucc_list.h"
+#include "utils/ucc_lock_free_queue.h"
 
 typedef struct ucc_pq_mt {
     ucc_progress_queue_t super;
-    ucc_spinlock_t       locked_queue_lock[NUM_POOLS];
-    ucc_coll_task_t *    tasks[NUM_POOLS][LINE_SIZE];
-    uint8_t              which_pool;
-    ucc_list_link_t      locked_queue[NUM_POOLS];
+    ucc_lf_queue_t       lf_queue;
 } ucc_pq_mt_t; // TODO the struct isn't a queue because not maintaining order, maybe another name
 
 typedef struct ucc_pq_mt_locked {
@@ -41,23 +35,11 @@ static void ucc_pq_locked_mt_enqueue(ucc_progress_queue_t *pq,
 static void ucc_pq_mt_enqueue(ucc_progress_queue_t *pq, ucc_coll_task_t *task)
 {
     ucc_pq_mt_t *pq_mt      = ucc_derived_of(pq, ucc_pq_mt_t);
-    int          which_pool = task->was_progressed ^ (pq_mt->which_pool & 1);
-    int          i;
-    for (i = 0; i < LINE_SIZE; i++) {
-        if (ucc_atomic_bool_cswap64((uint64_t *)&(pq_mt->tasks[which_pool][i]),
-                                    0, (uint64_t)task)) {
-            return;
-        }
-    }
-
-    ucc_spin_lock(&pq_mt->locked_queue_lock[which_pool]);
-    ucc_list_add_tail(&pq_mt->locked_queue[which_pool], &task->list_elem);
-    ucc_spin_unlock(&pq_mt->locked_queue_lock[which_pool]);
+    ucc_lf_queue_enqueue(&pq_mt->lf_queue, &task->lf_elem);
 }
 
 static void ucc_pq_locked_mt_dequeue(ucc_progress_queue_t *pq,
-                                     ucc_coll_task_t **    popped_task,
-                                     int                   is_first_call) //NOLINT 
+                                     ucc_coll_task_t **    popped_task)
 {
     ucc_pq_mt_locked_t *pq_mt = ucc_derived_of(pq, ucc_pq_mt_locked_t);
     *popped_task              = NULL;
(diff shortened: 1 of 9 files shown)


lappazos:
@alex--m regarding moving the structure outside of core - one of the structure features is the "was_progressed" flag - I think it is a feature that special for tasks objects, and not for any object. what do you think?

alex--m:
@lappazos I think even if we consider it a "task queue" and not a generic locked list it may both still be valueable to other components and be better placed in a separate section outside the "core" folder. This also makes it easier to test it separately and evaluate that threshold Manju asked about, without making it too progress-logic-specific. Regarding the "was-progressed" - yes, it makes sense to me to keep it even if the entire implementation moves to a common area.

vspetrov:
> @lappazos I think even if we consider it a "task queue" and not a generic locked list it may both still be valueable to other components and be better placed in a separate section outside the "core" folder. This also makes it easier to test it separately and evaluate that threshold Manju asked about, without making it too progress-logic-specific. Regarding the "was-progressed" - yes, it makes sense to me to keep it even if the entire implementation moves to a common area.

We could do a generic implementation of LF queue. Need to define a simple struct  - queue element:
```
struct ucc_lf_queue_elem {
    uint8_t was_progressed;
}
```
Then in the task struct instead of was_progressed you will have ucc_lf_queue_elem. And then you can perform do
ucc_lf_enqueue(ctx->pq, &task->lf_elem)
and
lf_elem = ucc_lf_dequeue(ctx->pq)
task = container_of(lf_elem, ...)

This way it will be generic and re-usable if needed.

lappazos:
@alex--m @vspetrov - Done

src/utils/ucc_lock_free_queue.h
```diff
@@ -0,0 +1,106 @@
...
+#include <string.h>
+
+/* This data structure is thread safe */
+
+// Number of elements in a single lock free pool - could be changed, but in tests performed great
+#define LINE_SIZE 8
+#define NUM_POOLS 2
+
+typedef struct ucc_lf_queue_elem {
+    uint8_t         was_progressed;
```
vspetrov, line 20: name "was_progresses" does not fit. It is generic lock-free queue implementation and it has no notion of "progress" - plz rename. Maybe "was_requeued" or smth?

src/utils/ucc_lock_free_queue.h
```diff
@@ -0,0 +1,106 @@
...
+            if (ucc_atomic_bool_cswap64(
+                    (uint64_t *)&(queue->elements[which_pool][i]),
+                    (uint64_t)elem, 0)) {
+                elem->was_progressed = 1;
+                return elem;
+            }
+        }
+    }
+    elem = NULL;
+    ucc_spin_lock(&queue->locked_queue_lock[which_pool]);
```
vspetrov, line 71: So, an element from the locked list can only be extracted when LINE_SIZE elements are placed in the alternative pool. Looks like there is a big imbalance in favor of pools there. this might lead to the stalls in the progress.. i don't know if we can improve that - and make the dequeue extracting elements with the same probability rate ?

vspetrov (commented):
In general i guess it should fix the hang that we saw in torch. However, need the blessing from @Sergei-Lebedev  - need to check it works (we had slightly different fix).
Secondly, @lappazos since we are adding a generic data structure - plz add a gtest and test it for correctness. need to design stressful test cases: 1. 1 producer, 1 cosumer; 2. 1 producer multiple consumers, 3. multiple producers multiple consumers; need to stress -test the fallback to the list specifically.

lappazos:
> In general i guess it should fix the hang that we saw in torch. However, need the blessing from @Sergei-Lebedev - need to check it works (we had slightly different fix).
> Secondly, @lappazos since we are adding a generic data structure - plz add a gtest and test it for correctness. need to design stressful test cases: 1. 1 producer, 1 cosumer; 2. 1 producer multiple consumers, 3. multiple producers multiple consumers; need to stress -test the fallback to the list specifically.

ill add the tests.

src/utils/ucc_lock_free_queue.h
lappazos: I'm not sure what you meant. the locked queue is a direct continuation of the LINE_SIZE pool. when inserting, you search for an empty spot from the beginning of LINE_SIZE till the end of locked_queue. when you pop, you search for elem from the beginning of LINE_SIZE to the end of locked queue. 
Of course, as we know, it will not maintain order, meaning - new elems that were inserted after some other elem, could be popped before it, doesn't matter if they are on LINE_SIZE or locked queue. But, we have a guarantee that no "old" elem will be popped before some new elem. meaning, if some 2 objects A & B are in pool and B was popped and reinserted, we can assure that B will never be popped again before A will.

src/utils/ucc_lock_free_queue.h
```diff
@@ -0,0 +1,106 @@
...
+#define UCC_LOCKFREE_QUEUE_H_
+
+#include "utils/ucc_spinlock.h"
+#include "utils/ucc_atomic.h"
+#include <string.h>
+
+/* This data structure is thread safe */
+
+// Number of elements in a single lock free pool - could be changed, but in tests performed great
+#define LINE_SIZE 8
```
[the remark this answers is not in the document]
lappazos: @Sergei-Lebedev ?

src/utils/ucc_lock_free_queue.h
lappazos: Fixed

lappazos:
@manjugv please add ready for review label

vspetrov:
@Sergei-Lebedev when you have time could you plz check that this MT implementation works with Pytorch (where we saw hang last time) - this will be last check.

Sergei-Lebedev:
@lappazos test container build failed with this PR with following error
/usr/bin/ld: ucc_info-ucc_info.o: undefined reference to symbol 'pthread_spin_init@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:493: ucc_info] Error 1

lappazos:
> @lappazos test container build failed with this PR with following error
> /usr/bin/ld: ucc_info-ucc_info.o: undefined reference to symbol 'pthread_spin_init@@GLIBC_2.2.5'
> /usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
> collect2: error: ld returned 1 exit status
> make[2]: *** [Makefile:493: ucc_info] Error 1

Fixed

src/core/ucc_progress_queue_mt.c
```diff
@@ -8,105 +8,65 @@
...
 
 typedef struct ucc_pq_mt {
     ucc_progress_queue_t super;
-    ucc_spinlock_t       locked_queue_lock;
-    ucc_coll_task_t *    tasks[NUM_POOLS][LINE_SIZE];
-    uint8_t              which_pool;
-    ucc_list_link_t      locked_queue;
-    uint32_t             tasks_counters[NUM_POOLS];
+    ucc_lf_queue_t       lf_queue;
 } ucc_pq_mt_t; // TODO the struct isn't a queue because not maintaining order, maybe another name
```
vspetrov, line 17: propose better name

src/core/ucc_progress_queue_mt.c
```diff
@@ -126,37 +86,49 @@ static int ucc_pq_mt_progress(ucc_progress_queue_t *pq)
...
-    ucc_list_head_init(&pq_mt->locked_queue);
-    pq_mt->which_pool        = 0;
-    pq_mt->tasks_counters[0] = 0;
-    pq_mt->tasks_counters[1] = 0;
-
     if (lock_free_progress_q) {
-        pq_mt->super.enqueue    = ucc_pq_mt_enqueue_opt;
-        pq_mt->super.dequeue    = ucc_pq_mt_dequeue_opt;
-    } else {
+    	// TODO: should we use ucc_mc_alloc?
```
vspetrov, line 107: no

src/utils/ucc_lock_free_queue.h
```diff
@@ -0,0 +1,104 @@
...
+    uint8_t         was_queued;
+    ucc_list_link_t locked_list_elem;
+} ucc_lf_queue_elem_t;
+
+typedef struct ucc_lf_queue {
+    ucc_spinlock_t       locked_queue_lock[NUM_POOLS];
+    ucc_lf_queue_elem_t *elements[NUM_POOLS][LINE_SIZE];
+    uint8_t              which_pool;
+    ucc_list_link_t      locked_queue[NUM_POOLS];
+} ucc_lf_queue_t; // TODO the struct isn't a queue because not maintaining order, maybe another name
```
vspetrov, line 30: propose better name

test/gtest/utils/test_lock_free_queue.cc
```diff
@@ -0,0 +1,113 @@
...
+    uint32_t       elems_num;
+    uint32_t       active_producers_threads;
+    uint32_t       memory_err;
+} ucc_test_queue_t;
+
+void *producer_thread(void *arg)
+{
+    ucc_test_queue_t *test = (ucc_test_queue_t *)arg;
+    for (int j = 0; j < 5000000; j++) {
+        // TODO: should we use ucc_mc_alloc instead of ucc_malloc?
```
vspetrov, line 29: no
add #defeine NUM_ITERS 500000

Sergei-Lebedev:
> In general i guess it should fix the hang that we saw in torch. However, need the blessing from @Sergei-Lebedev - need to check it works (we had slightly different fix).
Param comms alltoall test passed

vspetrov:
> > In general i guess it should fix the hang that we saw in torch. However, need the blessing from @Sergei-Lebedev - need to check it works (we had slightly different fix).
> 
> Param comms alltoall test passed

@Sergei-Lebedev  Performance wise: if we are to make the lock-free implementation default, which tests we need to run to make sure there is no perf degradation?

Sergei-Lebedev:
> > > In general i guess it should fix the hang that we saw in torch. However, need the blessing from @Sergei-Lebedev - need to check it works (we had slightly different fix).
> > 
> > 
> > Param comms alltoall test passed
> 
> @Sergei-Lebedev Performance wise: if we are to make the lock-free implementation default, which tests we need to run to make sure there is no perf degradation?

I usually run nonblocking mode in pytorch param benchmark, it stresses progress queue more. Do we have other use cases?

lappazos:
Replace with #210

PR #149, opened 2021-04-06
CODESTYLE: Change .clang-format file to match UCX

lappazos:
## What
Change clang format according to UCX's format

## Why ?
Clang format needs to improve

## How ?
Copy UCX .clang-format. 5 fields that appear in UCX clang were removed, since they are special clang expansions performed by UCX.
they use a custom build of clang-format, which can be found & compiled by:
`RUN export BUILD_ROOT=/tmp/llvm-project && \
    git clone https://github.com/openucx/llvm-project.git --depth=1 -b ucx-clang-format --single-branch ${BUILD_ROOT} && \
    mkdir -p ${BUILD_ROOT}/build && cd ${BUILD_ROOT}/build && \
    cmake -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS=clang -G "Unix Makefiles" \
    ../llvm && \
    make -j$(nproc) && make install && rm -rf ${BUILD_ROOT}`
    Using it in a CI pipeline requires replacing the generic image which the pipeline uses, by a custom built image which includes the above-mentioned custom build of LLVM.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,126 +1,119 @@
----
-Language:        Cpp
-# BasedOnStyle:  LLVM
-AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
+# C
+AlignEscapedNewlines: Right
 AlignConsecutiveAssignments: true
 AlignConsecutiveDeclarations: true
-AlignEscapedNewlines: Right
-AlignOperands:   true
-AlignTrailingComments: true
-AllowAllArgumentsOnNextLine: true
-AllowAllConstructorInitializersOnNextLine: true
-AllowAllParametersOfDeclarationOnNextLine: true
+AlignConsecutiveStructMembers: true
+AlignConsecutiveMacros: true
+AlignAfterOpenBracket: true
+AlignOperands: true
+SpaceAfterLogicalNot: false
+ReflowComments: false
+PenaltyBreakString: 250
+PenaltyBreakComment: 150
+IndentWrappedFunctionNames: false
+PointerAlignment: Right
+DerivePointerAlignment: false
+AlignTrailingComments: false
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: false
-AllowShortLambdasOnASingleLine: All
 AllowShortIfStatementsOnASingleLine: Never
 AllowShortLoopsOnASingleLine: false
-AlwaysBreakAfterDefinitionReturnType: None
+AllowShortEnumsOnASingleLine: false
+AllowDesignatedInitializersOnASingleLine: false
 AlwaysBreakAfterReturnType: None
+PenaltyReturnTypeOnItsOwnLine: 20
+PenaltyBreakAssignment: 100
+PenaltyExcessCharacter: 100
+PenaltyBreakBeforeFirstCallParameter: 100
+PenaltyBreakMemberAccess: 250
+PenaltyBreakLastMemberAccess: 300
+PenaltyIndentedWhitespace: 0
+ColumnLimit: 80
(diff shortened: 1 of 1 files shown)


lappazos:
@manjugv please review the removed fields in the commit, and try to find fields that UCX doesn't use but we might want to keep.

lappazos:
Please look at explanation in 'How ?' field

vspetrov:
I downloaded the PR and applied it on src/components/tl/ucp to see how it looks. The change commit is there: https://github.com/vspetrov/ucc/commit/f44aec4b7c5fcd3b0f200b32832bf10d017d0a49 for you just to look.

What i see is different from ours:
1. Fields inside structs/enums are no longer aligned: https://github.com/vspetrov/ucc/blob/f44aec4b7c5fcd3b0f200b32832bf10d017d0a49/src/components/tl/ucp/tl_ucp.h#L57
Does not look nice
2. Structs/enum initializations now have 8 spaces offset (used to be 4): https://github.com/vspetrov/ucc/blob/f44aec4b7c5fcd3b0f200b32832bf10d017d0a49/src/components/tl/ucp/tl_ucp.c#L17
Looks fine to me but this diverges from the "4 spaces alignment" that we agreed on originally
3. This style file solves the issue of:
```
if (aaa) {
    ...
}
else {
```
This is good imho.

So, the question is: what do we want to achieve with that PR? Is this a proposal for new code-stype or are we trying to fix issues with current clang format file, to make it produce what we agreed originally? For the latter case need at least to solve "struct fields alignment".

lappazos:
Spoke with @artemry-nv, he will push a PR adding the build of UCX clang to our CI pipeline. once that will be done, ill fix the current file to be exactly like UCX file, and then we can review

lappazos:
> I downloaded the PR and applied it on src/components/tl/ucp to see how it looks. The change commit is there: [vspetrov@f44aec4](https://github.com/vspetrov/ucc/commit/f44aec4b7c5fcd3b0f200b32832bf10d017d0a49) for you just to look.
> 
> What i see is different from ours:
> 
> 1. Fields inside structs/enums are no longer aligned: https://github.com/vspetrov/ucc/blob/f44aec4b7c5fcd3b0f200b32832bf10d017d0a49/src/components/tl/ucp/tl_ucp.h#L57
>    Does not look nice
> 2. Structs/enum initializations now have 8 spaces offset (used to be 4): https://github.com/vspetrov/ucc/blob/f44aec4b7c5fcd3b0f200b32832bf10d017d0a49/src/components/tl/ucp/tl_ucp.c#L17
>    Looks fine to me but this diverges from the "4 spaces alignment" that we agreed on originally
> 3. This style file solves the issue of:
> 
> ```
> if (aaa) {
>     ...
> }
> else {
> ```
> 
> This is good imho.
> 
> So, the question is: what do we want to achieve with that PR? Is this a proposal for new code-stype or are we trying to fix issues with current clang format file, to make it produce what we agreed originally? For the latter case need at least to solve "struct fields alignment".

As I wrote above, ill change it soon to be **extactly** like UCX clang, and then I believe the issues you mentioned will be resolved.
The point of this PR is to improve our automatic clang checks, and since we inspire to be as close as possible to UCX code style according to @manjugv, copy their clang should be the right approach.

alex--m (approved):

lappazos:
@vspetrov @manjugv please approve

manjugv:
@lappazos Probably you should check this new update: https://github.com/openucx/ucx/wiki/Code-style-checking

lappazos:
@manjugv, @vspetrov and I went over the configurations one by one. Both of us agree on these right now.
Also, I consulted with Yossi Itgin, and UCX left their code unchanged when they pushed new clang format - it was only applied for new PRs.
Bottom line - this PR is ready to be reviewed and merged.

lappazos:
@vspetrov @manjugv please approve

vspetrov:
closing. Necessary .clang_format fixes have been applied as part of be42d818 commit

PR #151, opened 2021-04-07
TEST: MPI: fixes tests skipping

karasevb:
## Why ?
If the UCC returns that something is not supported, these tests are flagged as skipped. In this case, updating tests won't be required if UCC starts supporting something.

### DIFF
--- a/test/mpi/test_allgatherv.cc
+++ b/test/mpi/test_allgatherv.cc
@@ -23,10 +23,8 @@ TestAllgatherv::TestAllgatherv(size_t _msgsize, ucc_test_mpi_inplace_t _inplace,
     MPI_Comm_size(team.comm, &size);
     args.coll_type = UCC_COLL_TYPE_ALLGATHERV;
 
-    if (test_max_size < (_msgsize*size)) {
-        test_skip = TEST_SKIP_MEM_LIMIT;
-    }
-    if (TEST_SKIP_NONE != skip_reduce(test_skip, team.comm)) {
+    if (TEST_SKIP_NONE != skip_reduce(test_max_size < (_msgsize*size),
+                                      TEST_SKIP_MEM_LIMIT, team.comm)) {
         return;
     }
 
@@ -64,7 +62,7 @@ TestAllgatherv::TestAllgatherv(size_t _msgsize, ucc_test_mpi_inplace_t _inplace,
     args.dst.info_v.displacements = (ucc_aint_t*)displacements;
     args.dst.info_v.datatype      = TEST_DT;
     args.dst.info_v.mem_type      = _mt;
-    UCC_CHECK(ucc_collective_init(&args, &req, team.team));
+    UCC_CHECK_SKIP(ucc_collective_init(&args, &req, team.team), test_skip);
 }
 
 TestAllgatherv::~TestAllgatherv() {
--- a/test/mpi/test_allreduce.cc
+++ b/test/mpi/test_allreduce.cc
@@ -8,19 +8,21 @@
 #include "mpi_util.h"
 TestAllreduce::TestAllreduce(size_t _msgsize, ucc_test_mpi_inplace_t _inplace,
                              ucc_datatype_t _dt, ucc_reduction_op_t _op,
-                             ucc_memory_type_t _mt, ucc_test_team_t &_team) :
-    TestCase(_team, _mt, _msgsize, _inplace)
+                             ucc_memory_type_t _mt, ucc_test_team_t &_team,
+                             size_t _max_size) :
+    TestCase(_team, _mt, _msgsize, _inplace, _max_size)
 {
     size_t dt_size = ucc_dt_size(_dt);
     size_t count = _msgsize/dt_size;
     int rank;
+
     MPI_Comm_rank(team.comm, &rank);
     op = _op;
     dt = _dt;
-
     args.coll_type = UCC_COLL_TYPE_ALLREDUCE;
 
-    if (TEST_SKIP_NONE != skip_reduce(test_skip, team.comm)) {
+    if (skip_reduce(test_max_size < _msgsize, TEST_SKIP_MEM_LIMIT,
+                    team.comm)) {
         return;
     }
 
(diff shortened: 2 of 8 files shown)


test/mpi/test_case.cc
```diff
@@ -114,21 +114,15 @@ ucc_status_t TestCase::exec()
     return status;
 }
 
-test_skip_cause_t TestCase::skip_reduce(int skip_cond, test_skip_cause_t cause,
-                                        MPI_Comm comm)
+test_skip_cause_t TestCase::skip_reduce(test_skip_cause_t cause, MPI_Comm comm)
```
vspetrov, line 117: "cause" is unused?

test/mpi/test_allgatherv.cc
```diff
@@ -64,7 +61,11 @@ TestAllgatherv::TestAllgatherv(size_t _msgsize, ucc_test_mpi_inplace_t _inplace,
     args.dst.info_v.displacements = (ucc_aint_t*)displacements;
     args.dst.info_v.datatype      = TEST_DT;
     args.dst.info_v.mem_type      = _mt;
-    UCC_CHECK(ucc_collective_init(&args, &req, team.team));
+
+    UCC_CHECK_SKIP(ucc_collective_init(&args, &req, team.team), test_skip);
```
vspetrov, line 65: Looks like now the test will init the collective if max mssg size is exceeded - so it won't be skipped, because you do collective_init before the skip_reduce. If we exceeding the msg size we don't want to spend time allocating/initializing/freeing buffers.

karasevb: Fixed, but it adds extra mpi_allreduce for the skip_cause

test/mpi/test_case.cc
karasevb: fix

test/mpi/test_mpi.h
```diff
@@ -23,10 +23,22 @@ END_C_DECLS
...
+        std::cerr << "*** UCC TEST FAIL: " << STR(_call) << "\n";       \
+        MPI_Abort(MPI_COMM_WORLD, -1);                                  \
+    }
+
+#define UCC_CHECK_SKIP(_call, _skip_cause)                              \
+    {                                                                   \
+        ucc_status_t status;                                            \
+        status = (_call);                                               \
+        if(UCC_ERR_NOT_SUPPORTED == status)  {                          \
+            _skip_cause = TEST_SKIP_NOT_IMPL_INPLACE;                   \
```
vspetrov, line 37: lets call it TEST_SKIP_NOT_SUPPORTED (there can be another reasons when UCC returns not supported, e.g. - not supported datatype)

test/mpi/test_bcast.cc
```diff
@@ -22,14 +22,6 @@ TestBcast::TestBcast(size_t _msgsize, ucc_test_mpi_inplace_t _inplace,
     root = _root;
     args.coll_type = UCC_COLL_TYPE_BCAST;
 
-    if (TEST_INPLACE == inplace && !ucc_coll_inplace_supported(args.coll_type)) {
-        test_skip = TEST_SKIP_NOT_IMPL_INPLACE;
-    }
-    if (skip_reduce(test_max_size < _msgsize, TEST_SKIP_MEM_LIMIT,
```
vspetrov, line 28: where is the MEM_LIMIT skip check fo bcast now?

test/mpi/test_allreduce.cc
```diff
@@ -14,16 +14,12 @@ TestAllreduce::TestAllreduce(size_t _msgsize, ucc_test_mpi_inplace_t _inplace,
...
     size_t count = _msgsize/dt_size;
     int rank;
+
     MPI_Comm_rank(team.comm, &rank);
     op = _op;
     dt = _dt;
-
     args.coll_type = UCC_COLL_TYPE_ALLREDUCE;
 
-    if (TEST_SKIP_NONE != skip_reduce(test_skip, team.comm)) {
```
vspetrov, line 23: MEM_LIMIT check?

test/mpi/test_allgatherv.cc
karasevb: @vspetrov removed mpi_allreduce for after `ucc_collective_init`

test/mpi/test_allreduce.cc
karasevb: fixed

test/mpi/test_bcast.cc
karasevb: fixed

test/mpi/test_mpi.h
karasevb: renamed

test/mpi/test_mpi.h
```diff
@@ -94,7 +106,7 @@ static inline const char* skip_str(test_skip_cause_t s) {
     switch(s) {
     case TEST_SKIP_MEM_LIMIT:
         return "maximum buffer size reached";
-    case TEST_SKIP_NOT_IMPL_INPLACE:
+    case TEST_SKIP_NOT_SUPPORTED:
```
Sergei-Lebedev, line 109: I have general question. Should we differentiate UCC_ERR_NOT_SUPPROTED and UCC_ERR_NOT_IMPLEMENTED? For example custom reductions in NCCL TL are not supported and likely will not be supported in future that's why return code might be UCC_ERR_NOT_SUPPORTED, in contrast in UCP TL custom reduction will be supported but so far not implemented thus return code is UCC_ERR_NOT_IMPLEMENTED.

vspetrov: I agree this is a difference. But from the test perspective: if ucc returns either NOT SUPPORTED or NOT_IMPLEMENTED it is not a test failure imho - there is nothing to fix. This is the expected flow that user has to process properly, right? On the other hand i remember @manjugv was saying that at some point UCC must SUPPORT all the spec. This makes sense as well... Meaning - even if some TL can't support we must still fallback to other path (cl_basic->tl_ucp) which supports everything. But this won't happen soon. I think so far we can do the test that way: lets check both NOT_SUPPORTED and NOT_IMPLEMENTED and consider them a skip reason.

Sergei-Lebedev: @karasevb pls fix error message accordingly

karasevb: Fixed with force-pushed commit

vspetrov (approved):

Sergei-Lebedev (approved):

PR #160, opened 2021-04-19
TL/UCP: knomial allreduce move memory allocation from post to init

Sergei-Lebedev:
## What
Split knomial allreduce post into init and post. Init function allocates memory.

## Why ?
1. In case of triggered post cuda malloc might hang if wait kernel is running, we can do allocation earlier in init and free it in finalize to avoid deadlock
2. For persistent collectives we don't need to allocate/free scratch space on each collective post, instead do it just once in init/finalize.

### DIFF
--- a/src/components/tl/ucp/tl_ucp_service_coll.c
+++ b/src/components/tl/ucp/tl_ucp_service_coll.c
@@ -17,6 +17,7 @@ ucc_status_t ucc_tl_ucp_service_allreduce(ucc_base_team_t *team, void *sbuf,
 {
     ucc_tl_ucp_team_t *tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
     ucc_tl_ucp_task_t *task    = ucc_tl_ucp_get_task(tl_team);
+    ucc_status_t status;
     ucc_coll_args_t args = {
         .coll_type            = UCC_COLL_TYPE_ALLREDUCE,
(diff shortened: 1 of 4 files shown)


src/components/tl/ucp/tl_ucp_service_coll.c
```diff
@@ -34,15 +35,32 @@ ucc_status_t ucc_tl_ucp_service_allreduce(ucc_base_team_t *team, void *sbuf,
...
+    }
     task->subset = subset;
     task->team = tl_team;
     task->tag  = UCC_TL_UCP_SERVICE_TAG;
     task->n_polls = 10; // TODO need a var ?
     task->super.progress = ucc_tl_ucp_allreduce_knomial_progress;
     memcpy(&task->args, &args, sizeof(ucc_coll_args_t));
     *task_p = &task->super;
-    return ucc_tl_ucp_allreduce_knomial_start(&task->super);
+    status = ucc_tl_ucp_allreduce_init(task);
```
vspetrov, line 49: this looks strange. we use knomial start below but for init we are now using old wrapper (ucc_tl_ucp_allreduce_init) which is not necessarily connected to knomial alg. why can't we call ucc_tl_ucp_allreduce_knomial_init_common here?

Sergei-Lebedev: sure, can use ucc_tl_ucp_allreduce_knomial_init_common

vspetrov (approved):

bureddy (approved):

PR #161, opened 2021-04-19
TL/UCP: SRA knomial allreduce

vspetrov:
## What
Adds knomial scatter-reduce-allgather (with arbitrary radix) algorithm.

## Why ?
This is one of the algs targeting large msg sizes

## How ?
Implemented as a schedule constructed of reduce-scatter and allgather tasks

(depends on #159) 
Some perf data: 
[ucc_sra.xlsx](https://github.com/openucx/ucc/files/6351836/ucc_sra.xlsx)

### DIFF
--- a/src/coll_patterns/recursive_knomial.h
+++ b/src/coll_patterns/recursive_knomial.h
@@ -53,9 +53,11 @@ typedef struct ucc_knomial_pattern {
  *  @param [out] p     ucc_knomial_pattern
  */
 
-static inline void ucc_knomial_pattern_init_impl(ucc_rank_t size, ucc_rank_t rank,
-                                            ucc_kn_radix_t radix,
-                                                 ucc_knomial_pattern_t *p, int backward)
+static inline void ucc_knomial_pattern_init_impl(ucc_rank_t             size,
+                                                 ucc_rank_t             rank,
+                                                 ucc_kn_radix_t         radix,
+                                                 ucc_knomial_pattern_t *p,
+                                                 int backward)
 {
     ucc_rank_t full_pow_size, n_full_subtrees;
     CALC_POW_RADIX_SUP(size, radix, p->pow_radix_sup, full_pow_size);
@@ -63,23 +65,25 @@ static inline void ucc_knomial_pattern_init_impl(ucc_rank_t size, ucc_rank_t ran
     p->n_extra       = size - n_full_subtrees * full_pow_size;
     p->radix         = radix;
     p->iteration     = backward ? p->pow_radix_sup - 1 : 0;
-    p->radix_pow = backward ? ((full_pow_size == size) ?
-                               full_pow_size / radix : full_pow_size): 1;
+    p->radix_pow = backward ? ((full_pow_size == size) ? full_pow_size / radix
+                                                       : full_pow_size)
+                            : 1;
     p->node_type     = KN_NODE_BASE;
     if (rank < p->n_extra * 2) {
         p->node_type = (rank % 2) ? KN_NODE_EXTRA : KN_NODE_PROXY;
     }
 }
 
-static inline void ucc_knomial_pattern_init_backward(ucc_rank_t size, ucc_rank_t rank,
-                                            ucc_kn_radix_t radix,
-                                            ucc_knomial_pattern_t *p)
+static inline void ucc_knomial_pattern_init_backward(ucc_rank_t     size,
+                                                     ucc_rank_t     rank,
+                                                     ucc_kn_radix_t radix,
+                                                     ucc_knomial_pattern_t *p)
 {
     ucc_knomial_pattern_init_impl(size, rank, radix, p, 1);
 }
 
 static inline void ucc_knomial_pattern_init(ucc_rank_t size, ucc_rank_t rank,
-                                            ucc_kn_radix_t radix,
+                                            ucc_kn_radix_t         radix,
                                             ucc_knomial_pattern_t *p)
 {
     ucc_knomial_pattern_init_impl(size, rank, radix, p, 0);
@@ -103,7 +107,8 @@ static inline int ucc_knomial_pattern_loop_done(ucc_knomial_pattern_t *p)
     return p->iteration == p->pow_radix_sup;
 }
 
-static inline int ucc_knomial_pattern_loop_done_backward(ucc_knomial_pattern_t *p)
+static inline int
+ucc_knomial_pattern_loop_done_backward(ucc_knomial_pattern_t *p)
 {
     return p->iteration == UINT8_MAX;
 }
@@ -111,7 +116,7 @@ static inline int ucc_knomial_pattern_loop_done_backward(ucc_knomial_pattern_t *
 static inline ucc_rank_t ucc_knomial_pattern_loop_rank(ucc_knomial_pattern_t *p,
                                                        ucc_rank_t rank)
 {
-    return (rank < p->n_extra * 2) ? rank/2 : rank - p->n_extra;
+    return (rank < p->n_extra * 2) ? rank / 2 : rank - p->n_extra;
 }
 
 static inline ucc_rank_t ucc_knomial_pattern_get_loop_peer(ucc_knomial_pattern_t *p,
@@ -139,7 +144,8 @@ static inline void ucc_knomial_pattern_next_iteration(ucc_knomial_pattern_t *p)
     ucc_assert(p->iteration <= p->pow_radix_sup);
 }
 
-static inline void ucc_knomial_pattern_next_iteration_backward(ucc_knomial_pattern_t *p)
+static inline void
+ucc_knomial_pattern_next_iteration_backward(ucc_knomial_pattern_t *p)
 {
     p->iteration--;
     p->radix_pow /= p->radix;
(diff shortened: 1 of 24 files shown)


manjugv (commented):
Thanks for the PR, Val. At a high level, the implementation is good. As discussed, please add this information before the function signature? 

- Algorithm Description 
- Reference 
- Variations to the original algorithm
- When to use this algorithm? 
- Known Issues

vspetrov:
> Thanks for the PR, Val. At a high level, the implementation is good. As discussed, please add this information before the function signature?
> 
> * Algorithm Description
> * Reference
> * Variations to the original algorithm
> * When to use this algorithm?
> * Known Issues

Done

src/coll_patterns/sra_knomial.h
```diff
@@ -0,0 +1,121 @@
...
+{
+    ucc_rank_t step_radix = 0;
+    ucc_rank_t k, peer;
+
+    for(k = 1; k < p->radix; k++) {
+        peer = ucc_knomial_pattern_get_loop_peer(p, rank, size, k);
+        if (peer == UCC_KN_PEER_NULL) continue;
+        step_radix += 1;
+    }
+    step_radix += 1;
```
Sergei-Lebedev, line 23: init strep_radix to 1 instead of this line

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
```diff
@@ -0,0 +1,270 @@
...
+        local_seg_count  = ucc_sra_kn_compute_seg_size(block_count, step_radix, local_seg_index);
+
+        rbuf = task->reduce_scatter_kn.scratch;
+        if (p->iteration != 0) {
+            rbuf = PTR_OFFSET(rbuf, block_count*dt_size);
+        }
+        for (loop_step = 1; loop_step < radix; loop_step++) {
+            peer = ucc_knomial_pattern_get_loop_peer(p, rank, size, loop_step);
+            if (peer == UCC_KN_PEER_NULL)
+                continue;
```
Sergei-Lebedev, line 135: can we do break instead?

src/components/tl/ucp/allreduce/allreduce.h
```diff
@@ -8,9 +8,61 @@
...
+#define CHECK_USERDEFINED_OP(_args, _team) do {                         \
+        if (_args.mask & UCC_COLL_ARGS_FIELD_USERDEFINED_REDUCTIONS) {  \
+            tl_error(UCC_TL_TEAM_LIB(_team),                            \
+                     "userdefined reductions are not supported yet");   \
+            status = UCC_ERR_NOT_SUPPORTED;                             \
+            goto out;                                                   \
+        }                                                               \
+    }while(0)
+
+#define CHECK_INPLACE(_args, _team) do {                                \
```
Sergei-Lebedev, line 26: I think CHECK_SAME_MEMTYPE is a better name

src/components/base/ucc_base_iface.h
```diff
@@ -98,9 +98,16 @@ typedef struct ucc_base_team_iface {
...
+
+enum {
+    UCC_BASE_COLL_ARGS_RADIX
+};
+
 typedef struct ucc_base_coll_args {
+    uint64_t        mask;
     ucc_coll_args_t args;
     ucc_team_t     *team;
+    uint32_t        radix;
```
Sergei-Lebedev, line 110: I don't think radix is generic enough to add it to base args. Why not extend tl ucp collective init so that it can take additional argument?

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -0,0 +1,210 @@
...
+            break;                                                             \
+        };                                                                     \
+    } while (0)
+
+
+#define SAVE_STATE(_phase)                      \
+    do {                                        \
+        task->reduce_scatter_kn.phase = _phase;      \
+    } while (0)
+
```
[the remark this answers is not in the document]
vspetrov: GOTO_PHASE can't be generic since not always we need the same number of phases. E.g. there is no "PROXY" phase in reduce_scatter. If you put into common place then you will get "unused labels". Regarding moving into pattern: i don't think this is right since all those PHASEs stuff is purely tl_ucp implementation specific thing. I will try to collapse some of them in tl_ucp

src/components/tl/ucp/tl_ucp.c
```diff
@@ -38,6 +38,21 @@ static ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
...
+    {"REDUCE_SCATTER_KN_RADIX", "4",
+     "Radix of the knomial reduce-scatter algorithm",
+     ucc_offsetof(ucc_tl_ucp_lib_config_t, reduce_scatter_kn_radix),
+     UCC_CONFIG_TYPE_UINT},
+
+    {"ALLGATHER_KN_RADIX", "4",
+     "Radix of the knomial allgather algorithm",
+     ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
+     UCC_CONFIG_TYPE_UINT},
+
```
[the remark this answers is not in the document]
vspetrov: What do you mean - single var for all the radixes? We usually use different radixes for small/large msg (small ar, and sra e.g.). I think. Do you mean adding separate table for "radix" variables which will inherit from the 1 common table? So that we could set them all?

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
vspetrov: no, the peer can be non-null at some further loop_step

src/components/tl/ucp/allreduce/allreduce.h
vspetrov: agree

src/components/base/ucc_base_iface.h
vspetrov: I want to use our common interface for tl_ucp_reduce_scatter_knomial and allgather_knomial so that i could use "schedule/task" mechanisms (and also all this should be available for custom schedules in hierarchical CL). If you break the fn signature - it won't work. 

SRA schedule allocates 2 tasks (2 different base_args) and needs to set the radix somehow for those tasks. I also felt it is not ideal place, but couldn't find another way.

src/coll_patterns/recursive_knomial.h
```diff
@@ -53,23 +53,44 @@ typedef struct ucc_knomial_pattern {
...
 {
     ucc_rank_t full_pow_size, n_full_subtrees;
     CALC_POW_RADIX_SUP(size, radix, p->pow_radix_sup, full_pow_size);
     n_full_subtrees  = size / full_pow_size;
     p->n_extra       = size - n_full_subtrees * full_pow_size;
     p->radix         = radix;
-    p->iteration     = 0;
-    p->radix_pow     = 1;
+    p->iteration     = backward ? p->pow_radix_sup - 1 : 0;
+    /* p->radix_pow     = backward ? full_pow_size*radix  : 1; */
```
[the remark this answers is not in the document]
vspetrov: )) turned out to be :
`p->radix_pow     = backward ?  ((full_pow_size == size) ? full_pow_size / radix : full_pow_size): 1;`

src/components/tl/ucp/tl_ucp.c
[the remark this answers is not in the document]
Sergei-Lebedev: Not a separate table, but something like UCC_TL_UCP_KNOMIAL_RADIX with default value auto. If user sets some value for this variable then we force apply that value for all knomial collectives.

src/components/tl/ucp/allgather/allgather_knomial.c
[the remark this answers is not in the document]
Sergei-Lebedev: if some phases are not needed you can just create dummy label with assert so that label is not unused

src/components/base/ucc_base_iface.h
Sergei-Lebedev: ok, why not to have two init functions?
```c
ucc_tl_ucp_allgather_init_internal(task, radix)
{

}
ucc_tl_ucp_allgather_init(task)
{
    return ucc_tl_ucp_allgather_init_internal(task, radix_from_config);
}
```

src/components/tl/ucp/allgather/allgather_knomial.c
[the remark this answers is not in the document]
vspetrov: moved to common recursive_knomial.h

src/components/base/ucc_base_iface.h
vspetrov: yes this is a possibility i considered. The only issue  i had is: if TL/hier would like to use reduce_scatter + allgather (on 1st level of hierarchy for example) with the radix value that TL/hier wants to set - how would this work? looks like it will be option of tl/hier and it will have to use generic TL interface (that has only base args, and task).

src/components/tl/ucp/tl_ucp.c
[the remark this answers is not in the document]
vspetrov: added a convenience variable, plz have a look if that is what you meant.

src/components/base/ucc_base_iface.h
vspetrov: added internal init function. removed radix from bargs.

Sergei-Lebedev (approved):

vspetrov:
@bureddy @manjugv any comments?

manjugv:
@vspetrov It looks fine to me.

manjugv (approved):

Issue #163, opened 2021-04-21
UCC compiled with cuda print ERROR if GPU is not available


vspetrov:
1. UCC compiled with CUDA
2. cudart is available in runtime
3. GPU is not available (or not enabled)
[~/workspace/ompi/build_rel]
```
[valentinp@hpchead]> mpirun -np 2 osu_allreduce -x 0 -i 1 -m 4:4
[1619030637.735735] [jazz20:166840:0]         mc_cuda.c:145  �   ERROR cuda failed with ret:100(no CUDA-capable device is detected)
[1619030637.735735] [jazz20:166841:0]         mc_cuda.c:145  |   ERROR cuda failed with ret:100(no CUDA-capable device is detected)
```
@Sergei-Lebedev should we handle this w/o error maybe?

vspetrov:
also note "?" character - log_component->name might be missing on mc (need to check)

PR #164, opened 2021-04-22
MC/CUDA: check device count

Sergei-Lebedev:
## What
Don't initialize cuda memory component if there are no cuda devices.
Set component name for logger
fixing #163


vspetrov (approved):

bureddy (approved):

vspetrov:
@Sergei-Lebedev plz port to v0.1.x. @manjugv i think this should be ported since it is sort of bug fix

PR #171, opened 2021-04-29
Cleanup build flags

Sergei-Lebedev:
## What

- Use compiler and preprocessor flags configured from compiler.m4
- Unify flags across different components. Use BASE_CPFFLAGS, BASE_CFLAGS, BASE_CXXFLAGS as common options
- Print configure summary table
- Fix incorrect statuses (ucs_status messed up with ucc_status)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -157,6 +157,8 @@ AS_IF([test "x$with_docs_only" = xyes],
      m4_include([config/m4/mpi.m4])
      m4_include([test/gtest/configure.m4])
 
+     mc_modules=":cpu"
+     tl_modules=""
      AC_MSG_RESULT([MPI perftest: ${mpi_enable}])
      AC_ARG_ENABLE([debug],
               AS_HELP_STRING([--enable-debug], [Enable extra debugging code (default is NO).]),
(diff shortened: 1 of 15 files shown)


configure.ac
```diff
@@ -169,29 +171,33 @@ AS_IF([test "x$with_docs_only" = xyes],
...
+     fi
     ]) # Docs only
 
-
-
-
-includes="-I${UCC_TOP_SRCDIR}/src -I${UCC_TOP_BUILDDIR}/src"
 CFLAGS="$CFLAGS -std=gnu11"
-CPPFLAGS="$UCS_CPPFLAGS $CPPFLAGS $includes -Wall -Werror"
+CPPFLAGS="$CPPFLAGS $UCS_CPPFLAGS $includes"
```
vspetrov, line 200: Where is "includes" variable? I cant find it. Does it compile out of src tree?

Sergei-Lebedev: yes, it was duplication, same set of includes was defined in BASE_CPPFLAGS
BASE_CPPFLAGS="$BASE_CPPFLAGS -I\${abs_top_srcdir}/src"
BASE_CPPFLAGS="$BASE_CPPFLAGS -I\${abs_top_builddir}"
BASE_CPPFLAGS="$BASE_CPPFLAGS -I\${abs_top_builddir}/src"

vspetrov (approved):

manjugv (approved):

PR #173, opened 2021-04-29
TOOLS: adding more options to perftest

Sergei-Lebedev:
## What
in ucc_perftest added support for:
- allgather, alltoall, broadcast, barrier
- different memory types
- different datatypes
- set esimated num eps and ppn

### DIFF
--- a/tools/perf/ucc_pt_bootstrap.h
+++ b/tools/perf/ucc_pt_bootstrap.h
@@ -19,16 +19,19 @@ protected:
     ucc_context_oob_coll_t context_oob;
     ucc_team_oob_coll_t team_oob;
     int ppn;
-    int find_ppn()
+    int local_rank;
+    void find_ppn()
     {
         int     comm_size = get_size();
+        int     comm_rank = get_rank();
         size_t *hashes    = new size_t[comm_size];
-        int     ppn       = 0;
         ucc_status_t st;
         void *req;
 
+        ppn = 0;
+        local_rank = 0;
         st = team_oob.allgather(&node_hash, hashes, sizeof(node_hash),
-                                                    team_oob.coll_info, &req);
+                                team_oob.coll_info, &req);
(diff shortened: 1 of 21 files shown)


tools/perf/ucc_pt_bootstrap.h
```diff
@@ -8,14 +8,66 @@
...
+        std::cerr <<"failed to find ppn" <<std::endl;
+        delete[] hashes;
+        return ppn;
+    }
 public:
+    ucc_pt_bootstrap()
+    {
+        char hostname[256];
+        gethostname(hostname, sizeof(hostname));
+        node_hash = std::hash<std::string>{}(std::string(hostname));
```
vspetrov, line 59: I've got a question regarding node_hash in general. I know we used that in several places, but what about the collision ? It can happen right, and there is no way even to detect it. The  code will not work and it is even hard to understand why. What should we do (asking for future, for UCC hierarchy discovery) ?
In this case btw, you can make get_ppn virtual and in MPI case you can use OMPI/MPICH env vars or fallback to base class if no env var found.

vspetrov (approved):

manjugv:
@Sergei-Lebedev do you suggest porting to the release branch?

tools/perf/ucc_pt_bootstrap.h
Sergei-Lebedev: perhaps we can use ucs_machine_guid. It returns uint64_t based on mac address and host name, but it's not public currently

Sergei-Lebedev:
> @Sergei-Lebedev do you suggest porting to the release branch?

I'm not sure. Are we still sticking to the plan not to introduce new features to release branch?

Sergei-Lebedev:
bot:retest

manjugv (approved):

artemry-nv:
bot:retest

PR #174, opened 2021-04-29
CORE: fix memory leaks

Sergei-Lebedev:
## What
Fixing memory leaks in different places found by address sanitizer

### DIFF
--- a/test/gtest/coll_score/test_score_str.cc
+++ b/test/gtest/coll_score/test_score_str.cc
@@ -72,7 +72,6 @@ UCC_TEST_F(test_score_str, check_range_1)
     EXPECT_EQ(UCC_OK,
               check_range(score, UCC_COLL_TYPE_ALLTOALL, UCC_MEMORY_TYPE_CUDA,
                           RLIST({RANGE(64, 256, 10)})));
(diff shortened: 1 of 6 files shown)


test/gtest/coll_score/test_score_str.cc
```diff
@@ -71,7 +71,6 @@ UCC_TEST_F(test_score_str, check_range_1)
     EXPECT_NE(UCC_OK,
               check_range(score, UCC_COLL_TYPE_ALLTOALL, UCC_MEMORY_TYPE_HOST,
                           RLIST({RANGE(64, 256, 10)})));
-    ucc_coll_score_free(score);
```
vspetrov, line 74: Why free is removed here?

Sergei-Lebedev: because score variable is already freed in line 69

manjugv (approved):

test/gtest/coll_score/test_score_str.cc
vspetrov: i see. now check_range at line 72 accesses score after free. lets remove score_free at 69 and keep it at 74.

PR #177, opened 2021-05-04
API: completion callback

vspetrov:
## What
According to the decision made in WG: Adds the API option to set completion callback for ucc_coll_args_t. 

## Why ?
Used to implement the support for non-blocking collectives in OpenMPI coll/ucc driver (other use-cases also possible): https://github.com/vspetrov/ompi/tree/coll_ucc_v2 . It used to handle ompi/ucc requests allocation/progress/signalling.

## How ?
User can set UCC_COLL_ARGS_FIELD_CB flag to coll_args and pass a callback function which is invoked upon collective completion. The only UCC api allowed from the callback is ucc_collective_finalize.

### DIFF
--- a/src/schedule/ucc_schedule.h
+++ b/src/schedule/ucc_schedule.h
@@ -7,6 +7,7 @@
 
 #include "ucc/api/ucc.h"
 #include "utils/ucc_list.h"
+#include "utils/ucc_log.h"
 #define MAX_LISTENERS 4
 
 typedef enum {
@@ -29,7 +30,8 @@ typedef struct ucc_event_manager {
 } ucc_event_manager_t;
 
 enum {
-    UCC_COLL_TASK_FLAG_INTERNAL = UCC_BIT(0)
+    UCC_COLL_TASK_FLAG_INTERNAL = UCC_BIT(0),
+    UCC_COLL_TASK_FLAG_CB       = UCC_BIT(1)
 };
 
 typedef struct ucc_coll_task {
@@ -38,6 +40,7 @@ typedef struct ucc_coll_task {
     ucc_coll_post_fn_t           post;
     ucc_coll_triggered_post_fn_t triggered_post;
     ucc_coll_finalize_fn_t       finalize;
+    ucc_coll_callback_t          cb;
     ucc_event_manager_t          em;
     ucc_task_event_handler_p     handlers[UCC_EVENT_LAST];
     ucc_status_t               (*progress)(struct ucc_coll_task *self);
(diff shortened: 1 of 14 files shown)


src/ucc/api/ucc_def.h
```diff
@@ -63,6 +63,18 @@ typedef struct ucc_coll_req {
...
+ * @brief UCC collective completion callback
+ *
+ * The callback is invoked whenever the collective operation is completed.
+ * It is not allowed to call UCC APIs from the completion callback except
+ * for @ref ucc_collective_finalize.
+ */
+typedef struct ucc_coll_callback {
+    void (*cb)(void *data, ucc_status_t status);
+    void  *data;
+} ucc_coll_callback_t;
```
manjugv, line 77: 1. I guess the callback should be optional, right?
2. Does it have to be thread safe?

vspetrov: 1. It is optional (and passed on per collective basis, so user can decide for which colls to request the CB)
2. It is users responsibility to protect anything inside the  CB in the multi-threaded env.

manjugv: Makes sense. Let's put that in the document in the header.

manjugv: Another question - does mean every endpoint participating in the collectives should opt-in or opt-out? Can few endpoints decide to use callback and others decide to not use callback?

manjugv:
Is it targeting only to master or release branch also?

src/ucc/api/ucc_def.h
vspetrov: As we discussed in WG: we don't put any restrictions with respect to that - up to user to not introduce dead locks.

src/schedule/ucc_schedule.h
```diff
@@ -69,4 +72,23 @@ ucc_status_t ucc_schedule_init(ucc_schedule_t *schedule, ucc_context_t *ctx);
 void ucc_schedule_add_task(ucc_schedule_t *schedule, ucc_coll_task_t *task);
 ucc_status_t ucc_schedule_start(ucc_schedule_t *schedule);
 
+static inline ucc_status_t ucc_task_complete(ucc_coll_task_t *task)
+{
+    ucc_status_t status;
+    if (task->super.status < 0) {
+        /* error in task status */
+        ucc_error("failure in task %p, %s", task,
+                  ucc_status_string(task->super.status));
```
bureddy, line 81: shouldn't we call callback if requested with error status in thi case

bureddy (approved):

manjugv (approved):

PR #180, opened 2021-05-06
MC: Fix mem_query when addr is NULL

bureddy:


vspetrov (approved):
@bureddy this PR is good on its own. However, plz have a look into ucc_mc_query routine in ucc_mc.c. Looks like if we have only CPU mem component AND user will set PRE_REG_MEM=1 then mem_attr is not initialized at all (we skip all components) but the return status is UCC_OK.

TL/UCP then may try to "populate rcache" with garbage. Need to handle that as well. Do you want a separate PR or maybe update this one?

src/components/tl/ucp/tl_ucp.c
```diff
@@ -134,6 +134,10 @@ void ucc_tl_ucp_pre_register_mem(ucc_tl_ucp_team_t *team, void *addr,
     ucc_mem_attr_t mem_attr;
     ucc_status_t status;
 
+    if (addr == NULL) {
```
Sergei-Lebedev, line 137: minor, if (addr == NULL || length == 0)

Sergei-Lebedev (approved):

PR #190, opened 2021-05-13
CORE: remove ucc_tl_type_t

vspetrov:
## What
Now TLs are black boxes discovered in runtime and accessed via  component->name. CLs define the required TLs via
ucc_config_names_array (registered with parser) instead of tl_type_t bits.

## Why ?
This is the first PR the allows 3rd parties to create dynamically loadable TL plugins w/o ANY code integration into UCC.

## How ?
TLs are accessed via the "name" key which is part of .so tl library. No more "ucc_tl_type_t" that would require integration into ucc code.

### DIFF
--- a/src/utils/ucc_parser.c
+++ b/src/utils/ucc_parser.c
@@ -1,0 +1,43 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#include "ucc_parser.h"
+#include "ucc_malloc.h"
+#include "ucc_log.h"
+
+ucc_status_t ucc_config_names_array_dup(ucc_config_names_array_t *dst,
+                                        const ucc_config_names_array_t *src)
+{
+    int i;
+    dst->names = ucc_malloc(sizeof(char*) * src->count, "ucc_config_names_array");
+    if (!dst->names) {
+        ucc_error("failed to allocate %zd bytes for ucc_config_names_array",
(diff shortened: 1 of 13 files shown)


src/utils/ucc_parser.c
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#include "ucc_parser.h"
+#include "ucc_malloc.h"
+#include "ucc_log.h"
+
+ucc_status_t ucc_config_names_array_dup(ucc_config_names_array_t *dst,
```
Sergei-Lebedev, line 10: why not to use ucs_config_clone_array and ucs_config_release_array?

vspetrov: I checked: those routines do different thing: they clone the whole array of parameters (ucs_config_field_t array). Here is just want to create a copy of names array

bureddy:
@vspetrov is this PR subset of https://github.com/openucx/ucc/pull/191. i saw similar changes in that as well

vspetrov:
> @vspetrov is this PR subset of #191. i saw similar changes in that as well

#191 depends on this one. It will be rebased when 190 is merged. 191 shows how 190 is going to be used.

Sergei-Lebedev (approved):

manjugv:
How do I know that third party TLs are available or not? ucc_info will inform me?

bureddy (approved):

manjugv (approved):

PR #193, opened 2021-05-17
UCC context address query

vspetrov:
## What
Implements ucc_context_get_attr with UCC_CONTEXT_ATTR_FIELD_CTX_ADDR flag.

## Why ?
CORE API. It serves 2 purposes:
1. User can use PMIx-like (key-value storage) method for UCC address exchange.
2. Internally this will unify the address-exchange of different TLs. E.g. currently TL/UCP and TL/NCCL use 2 separate allgathers to collect addressing info. We will be able to move that to 1 common place in CORE team create (or ctx create if it has oob): we will query the TLs addresses and pack the data, do allgather once and then each TL will extract its own part of address. Additional data will be later added to the packed address: e..g ctx id hash (used for ep hash in tl/ucp currently); generic process info (bound socket/numa/core number, pid, nodename/nodehash) - will be used for hierarchical CL.

## How ?
The address layout is described in ucc_context.h

### DIFF
--- a/src/components/cl/basic/cl_basic_context.c
+++ b/src/components/cl/basic/cl_basic_context.c
@@ -63,9 +63,13 @@ UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_context_t)
 
 UCC_CLASS_DEFINE(ucc_cl_basic_context_t, ucc_cl_context_t);
 
-ucc_status_t ucc_cl_basic_get_context_attr(const ucc_base_context_t *context, /* NOLINT */
-                                           ucc_base_attr_t *attr)             /* NOLINT */
+ucc_status_t
+ucc_cl_basic_get_context_attr(const ucc_base_context_t *context, /* NOLINT */
+                              ucc_base_ctx_attr_t      *attr)
 {
-    /* TODO */
-    return UCC_ERR_NOT_IMPLEMENTED;
+    if (attr->attr.mask & UCC_CONTEXT_ATTR_FIELD_CTX_ADDR_LEN) {
+        attr->attr.ctx_addr_len = 0;
+    }
+    // attr->attr.mask & UCC_CONTEXT_ATTR_FIELD_CTX_ADDR - nothing to do
+    return UCC_OK;
 }
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -697,7 +697,7 @@ typedef struct ucc_context_attr {
     uint64_t                mask;
     ucc_context_type_t      type;
     ucc_coll_sync_type_t    sync_type;
-    ucc_context_addr_t      ctx_addr;
+    ucc_context_addr_h      ctx_addr;
     ucc_context_addr_len_t  ctx_addr_len;
 } ucc_context_attr_t;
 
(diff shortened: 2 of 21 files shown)


src/components/cl/basic/cl_basic_context.c
```diff
@@ -63,9 +63,13 @@ UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_context_t)
...
+ucc_status_t
+ucc_cl_basic_get_context_attr(const ucc_base_context_t *context, /* NOLINT */
+                              ucc_base_ctx_attr_t      *attr)
 {
-    /* TODO */
-    return UCC_ERR_NOT_IMPLEMENTED;
+    if (attr->attr.mask & UCC_CONTEXT_ATTR_FIELD_CTX_ADDR_LEN) {
+        attr->attr.ctx_addr_len = 0;
+    }
+    // attr->attr.mask & UCC_CONTEXT_ATTR_FIELD_CTX_ADDR - nothing to do
```
manjugv, line 73: Minor: dead code? plz remove it.

src/ucc/api/ucc.h
```diff
@@ -909,6 +909,27 @@ ucc_status_t ucc_context_destroy(ucc_context_h context);
...
+ *
+ *  @brief @ref ucc_context_free_attr routine releases the resources
+ *  associated with the context attributes described by @ref ucc_context_attr.
+ *
+ *  @endparblock
+ *
+ *  @return Error code as defined by ucc_status_t
+ */
+
+ucc_status_t ucc_context_free_attr(ucc_context_attr_t *context_attr);
```
manjugv, line 932: Do we need this? The context attributes are created by the library and it should destroy it.

vspetrov: But when? For example, if user asks for context address, then there is memory allocated during ucc_context_get_attr call and that stores the address. User can take the buffer with the address from attribs. When library can clear it? User needs another api to release any resources associated with attribs (in the case get_attr implies resource allocation like in our case)

src/components/cl/basic/cl_basic_context.c
vspetrov: do you mean the comment? It explains why the address is not packed (CL BASIC has no address, at least right now). The ctx_addr_len setting to 0 is valid (not dead) code and must be there.

manjugv: Yes, the comment.

src/ucc/api/ucc.h
manjugv: Why not during the context destroy? If we provide this free attributes approach, we need to make sure that free attributes is called before the context destroy.

vspetrov: agree - this makes sense, did think about it. I made the change: ctx address is packed on demand and cached on context. Released in context_destroy

src/utils/ucc_math.h
```diff
@@ -34,6 +34,23 @@ static inline size_t ucc_dt_size(ucc_datatype_t dt)
     return 0;
 }
 
+
+#define PTR_OFFSET(_ptr, _offset)                                              \
```
Sergei-Lebedev, line 38: duplication?

Sergei-Lebedev (approved):

manjugv (approved):
It looks good! You need to resolve the conflicts.

PR #191, opened 2021-05-13
CL/BASIC: use ALL available TLs by default

vspetrov:
## What
All CLs (even though we have only 1 at the moment  - basic) have the parameter UCC_CL_BASIC_TLS which defines the TLS the given CL requires. It can be ALL which is set to default for BASIC. CL/BASIC will try to use ALL tls specified with this variable (can be controlled). The logic of tl context/team/score creation in CL/BASIC is unified. 

## Why ?
After that change any 3rd party TL plugin (.so) can be just distributed separately and will be enabled via CL basic OOB according to default scoring settings.

PR #202, opened 2021-05-26
CL/BASIC: generalize get_lib_attr

vspetrov:
## What
Query all TLs when building CL/BASIC lib attributes.

## Why ?
CL Basic now (after #191 ) can operate with arbitrary TLs found in runtime. New TLs that will be added later (e.g. sharp, dpu) will not need to modify this function any more.

### DIFF
--- a/src/components/cl/basic/cl_basic_lib.c
+++ b/src/components/cl/basic/cl_basic_lib.c
@@ -28,43 +28,65 @@ UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_lib_t)
 
 UCC_CLASS_DEFINE(ucc_cl_basic_lib_t, ucc_cl_lib_t);
 
-ucc_status_t ucc_cl_basic_get_lib_attr(const ucc_base_lib_t *lib,
-                                       ucc_base_attr_t *base_attr)
+static inline ucc_status_t check_tl_lib_attr(const ucc_base_lib_t *lib,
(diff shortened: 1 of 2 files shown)


bureddy (approved):

src/components/cl/basic/cl_basic_lib.c
```diff
@@ -28,43 +28,64 @@ UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_lib_t)
...
-    status = tl_ucp_iface->lib.get_attr(NULL, &ucp_tl_attr.super);
+    ucc_tl_lib_attr_t tl_attr;
+    ucc_status_t      status;
+
+    memset(&tl_attr, 0, sizeof(tl_attr));
+    status = tl_iface->lib.get_attr(NULL, &tl_attr.super);
     if (UCC_OK != status) {
-        cl_error(lib, "failed to query cl lib attributes");
-        return status;
+        cl_warn(lib, "failed to query tl %s lib attributes",
```
Sergei-Lebedev, line 41: why not return error?

Sergei-Lebedev (approved):

src/components/cl/basic/cl_basic_lib.c
vspetrov: right, no reason - reverted it back. At first i thought that maybe we can still proceed, but now i agree. If a component failed to report its caps we can't use it anyways. So, reporting error.

PR #208, opened 2021-05-31
TEST: Fix data_finiliaze in test_bcast

lappazos:
## What
Fix free in test_bcast 

## Why ?
Un-needed free

## How ?
Add same condition as in allocation

### DIFF
--- a/test/gtest/core/test_bcast.cc
+++ b/test/gtest/core/test_bcast.cc
@@ -54,10 +54,13 @@ public:
     }
     void data_fini(UccCollCtxVec ctxs)
     {
-        for (gtest_ucc_coll_ctx_t* ctx : ctxs) {
+        for (auto r = 0; r < ctxs.size(); r++) {
+            gtest_ucc_coll_ctx_t *ctx = ctxs[r];
             ucc_coll_args_t* coll = ctx->args;
             UCC_CHECK(ucc_mc_free(coll->src.info.buffer, mem_type));
-            UCC_CHECK(ucc_mc_free(ctx->init_buf, UCC_MEMORY_TYPE_HOST));
+            if (r == coll->root) {
+                UCC_CHECK(ucc_mc_free(ctx->init_buf, UCC_MEMORY_TYPE_HOST));
+            }
             free(coll);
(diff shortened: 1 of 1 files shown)


lappazos:
bot:retest

vspetrov:
@lappazos why do we need this PR? i saw this change in @shimmybalsam PR on mpool.

lappazos:
> @lappazos why do we need this PR? i saw this change in @shimmybalsam PR on mpool.

@vspetrov just because it's a separate bug, to keep order. @shimmybalsam will simply rebase and solve the conflict

test/gtest/core/test_bcast.cc
```diff
@@ -54,10 +54,13 @@ class test_bcast : public UccCollArgs, public ucc::test
...
     void data_fini(UccCollCtxVec ctxs)
     {
-        for (gtest_ucc_coll_ctx_t* ctx : ctxs) {
+        for (auto r = 0; r < ctxs.size(); r++) {
+            gtest_ucc_coll_ctx_t *ctx = ctxs[r];
             ucc_coll_args_t* coll = ctx->args;
             UCC_CHECK(ucc_mc_free(coll->src.info.buffer, mem_type));
-            UCC_CHECK(ucc_mc_free(ctx->init_buf, UCC_MEMORY_TYPE_HOST));
+            if (r == coll->root) {
+                UCC_CHECK(ucc_mc_free(ctx->init_buf, UCC_MEMORY_TYPE_HOST));
```
vspetrov, line 62: If we are to fix, then lets do it right: the init_buf array is allocated always on host (UCC_MEMORY_TYPE_HOST is provided). We should not use ucc_mc_alloc in this case, but rather use ucc_malloc (and hence ucc_free).

vspetrov (changes requested):

test/gtest/core/test_bcast.cc
lappazos: in another PR

vspetrov (approved):

Sergei-Lebedev (approved):

PR #209, opened 2021-05-31
TEST: Fix test_obj_size

lappazos:
## What
Fix test_obj_size test 

## Why ?
UCS ucs_log_component_config_t was changed recently

## How ?
Adjust expected value


Sergei-Lebedev (approved):
Right now it's enough to just fix gtest because new field in log component config structure is used only in default log config instance. **UCX**_LOG_FILE_FILTER will be applied to both UCX and UCC logger, from UCC we don't have control on file filtering.

vspetrov (approved):

PR #210, opened 2021-05-31
CORE: Fix MT progress queue

lappazos:
## What
Fix bug in MT progress queue. In addition, move to Utils and add gtests.

## Why ?
Hang while using mt progress queue after PR 125

## How ?
Create locked queue per pool instead of global. Also, organize the structs in a more accurate way.
In addition, remove counters (seems like only hurt performance) - add unnecessary actions per call. When empty, dequeue func simply finish the iteration.

### DIFF
--- a/src/schedule/ucc_schedule.c
+++ b/src/schedule/ucc_schedule.c
@@ -26,8 +26,8 @@ ucc_status_t ucc_coll_task_init(ucc_coll_task_t *task)
 {
     task->super.status   = UCC_OPERATION_INITIALIZED;
     task->ee             = NULL;
-    task->was_progressed = 0;
     task->flags          = 0;
+    ucc_lf_queue_init_elem(&task->lf_elem);
     return ucc_event_manager_init(&task->em);
 }
 
--- a/src/schedule/ucc_schedule.h
+++ b/src/schedule/ucc_schedule.h
@@ -8,6 +8,8 @@
 #include "ucc/api/ucc.h"
 #include "utils/ucc_list.h"
 #include "utils/ucc_log.h"
+#include "utils/ucc_lock_free_queue.h"
+
 #define MAX_LISTENERS 4
 
 typedef enum {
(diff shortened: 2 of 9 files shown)


lappazos:
Replace #147

src/schedule/ucc_schedule.c
```diff
@@ -24,10 +24,10 @@ void ucc_event_manager_subscribe(ucc_event_manager_t *em, ucc_event_t event,
 
 ucc_status_t ucc_coll_task_init(ucc_coll_task_t *task)
 {
-    task->super.status   = UCC_OPERATION_INITIALIZED;
-    task->ee             = NULL;
-    task->was_progressed = 0;
-    task->flags          = 0;
+    task->super.status       = UCC_OPERATION_INITIALIZED;
+    task->ee                 = NULL;
+    task->lf_elem.was_queued = 0;
```
vspetrov, line 29: @lappazos i think we should add a generic (maybe inline) function ucc_lf_queue_elem_init instead of directly initializing was_queued. LF queueue currently is a data structure that provides all the interfaces and hides the implementation. UCC schedule should not access internals like "was_queued" imho.

src/schedule/ucc_schedule.h
```diff
@@ -49,9 +51,10 @@ typedef struct ucc_coll_task {
     ucc_ev_t                    *ev;
     void                        *ee_task;
     ucc_coll_task_t             *triggered_task;
-    /* used for progress queue */
+    /* used for st & locked mt progress queue */
     ucc_list_link_t              list_elem;
-    uint8_t                      was_progressed;
+    /* used for lf mt progress queue */
+    ucc_lf_queue_elem_t          lf_elem;
```
vspetrov, line 57: should we put list_elem and lf_elem under "union" ? they can't be used simultaneously.

vspetrov (changes requested):

vspetrov (approved):

bureddy (approved):

PR #219, opened 2021-06-15
UCP: Implementing Allreduce SRA KN offload

shimmybalsam:
## What
Allreduce SRA KN offload scratch buffers and reduce_scatter compute from cuda to host.

## Why ?
To reduce compute traffic on cuda during Allreduce_sra_knomial runtime.

## How ?
Allocating HOST scratch buffers and choosing relevant mc component to run reduce_scatter with. Then writing the data back into cuda recieve buffer.

### DIFF
--- a/.github/workflows/codestyle.yaml
+++ b/.github/workflows/codestyle.yaml
@@ -36,7 +36,7 @@ jobs:
               return 1
             fi
           fi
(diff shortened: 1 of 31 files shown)


manjugv:
Can we close this PR? or WIP?

manjugv:
WG Discussion: Val will loopback

vspetrov:
closing since it requires more work (will not work with triggered_post, needs rebase on master). Will re-open when we have stronger use-case (request)

PR #221, opened 2021-06-15
TL/NCCL: add reduce scatter and reduce

Sergei-Lebedev:
## What
Adding reduce scatter and reduce to NCCL TL and corresponding tests.

For discussion:
In case of in-place reduce-scatter how the destination vector should look like. Rank N will have result shifted by N * count?

### DIFF
--- a/src/components/tl/nccl/tl_nccl_coll.c
+++ b/src/components/tl/nccl/tl_nccl_coll.c
@@ -374,3 +374,108 @@ ucc_status_t ucc_tl_nccl_bcast_init(ucc_tl_nccl_task_t *task)
     task->super.post     = ucc_tl_nccl_bcast_start;
     return UCC_OK;
 }
+
+ucc_status_t ucc_tl_nccl_reduce_scatter_start(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_nccl_task_t *task   = ucc_derived_of(coll_task, ucc_tl_nccl_task_t);
+    ucc_tl_nccl_team_t *team   = task->team;
+    ucc_ee_h            ee     = coll_task->ee;
+    cudaStream_t        stream = (ee) ? (cudaStream_t) ee->ee_context : team->stream;
+    void               *dst    = task->args.dst.info.buffer;
+    void               *src    = task->args.src.info.buffer;
+    ucc_status_t        status = UCC_OK;
+    ncclDataType_t      dt     = ucc_to_nccl_dtype[
+                                    task->args.src.info.datatype];
+    ncclRedOp_t         op     = ucc_to_nccl_reduce_op[
+                                    task->args.reduce.predefined_op];
+    size_t              count  = task->args.src.info.count / team->size;
+
+    task->super.super.status = UCC_INPROGRESS;
+    if (UCC_IS_INPLACE(task->args)) {
+        src = task->args.dst.info.buffer;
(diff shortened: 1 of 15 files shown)


vspetrov:
@Sergei-Lebedev @manjugv how is the UCC ReduceScatter "count" field is defined in ucc.h ? Is it amount of dtypes received by each peer (the way it is in MPI), ie reflects the size of rbuf and sbuf = count*team->size ?

For the allgather it IS defined that way - the amount of data per process.

Related discussion: no matter how we define the ucc.h is suggest that internally (in base iface for CL and TL) we define "count" both for allgather and reduce-scatter as "total" count. As we already discussed with Sergey, that would allow us to use RS and AG primitives for other algs internally w/o going to allgatherv, reduce_scatterv that can be less efficient. 

If this is what we agree on then this PR might need additional changes.

manjugv:
Given that we have consolidated buffer info into one structure for all collectives, it makes sense to define it as “total”, IMO. Though this interpretation is different from MPI, this makes sense from the UCC perspective. PR #225 is an attempt to clarify this.

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -396,3 +396,108 @@ ucc_status_t ucc_tl_nccl_bcast_init(ucc_tl_nccl_task_t *task)
...
+    ucc_ee_h            ee     = coll_task->ee;
+    cudaStream_t        stream = (ee) ? (cudaStream_t) ee->ee_context : team->stream;
+    void               *dst    = task->args.dst.info.buffer;
+    void               *src    = task->args.src.info.buffer;
+    ucc_status_t        status = UCC_OK;
+    ncclDataType_t      dt     = ucc_to_nccl_dtype[
+                                    task->args.src.info.datatype];
+    ncclRedOp_t         op     = ucc_to_nccl_reduce_op[
+                                    task->args.reduce.predefined_op];
+    size_t              count  = task->args.src.info.count;
```
vspetrov, line 466: currently this is correct, but looks like according to the discussion in #225 we will require user to set always both src.info and dst.info. And then i guess it will be logic to ignore "src" entirely in case of INPLACE. Then in the algs the right way would be to take count from dst.src.info. What do you think?

test/mpi/test_reduce_scatter.cc
```diff
@@ -0,0 +1,103 @@
...
+}
+
+ucc_status_t TestReduceScatter::check()
+{
+    int comm_rank, comm_size;
+
+    MPI_Comm_rank(team.comm, &comm_rank);
+    MPI_Comm_size(team.comm, &comm_size);
+    MPI_Reduce_scatter(inplace ? MPI_IN_PLACE : check_sbuf, check_rbuf,
+                       recvcounts, ucc_dt_to_mpi(dt), ucc_op_to_mpi(op),
```
vspetrov, line 82: don't need "recvcounts", you should use MPI_Reduce_scatter_block because it matches UCC_REDUCE_SCATTER

vspetrov (approved):

manjugv (approved):

vspetrov:
need to update the src.count usage according to the API

vspetrov:
@Sergei-Lebedev plz rebase on top of master and lets merge it

Sergei-Lebedev:
@vspetrov  updated, pls take a look.
One more question, what should be the layout of recv buffer in case of inplace? NCCL stores result with offset equal to block_size*rank in that case while MPI has no offset.

PR #224, opened 2021-06-16
MC/CUDA: fp16 reduce

Sergei-Lebedev:
## What
Added support for UCC_DT_FLOAT16 in cuda memory component

### DIFF
--- a/src/components/mc/cuda/kernel/mc_cuda_reduce_multi.cu
+++ b/src/components/mc/cuda/kernel/mc_cuda_reduce_multi.cu
@@ -14,13 +14,28 @@ extern "C" {
 }
 #endif
 
-#include <assert.h>
-#include <stdio.h>
+#include "mc_cuda_reduce_ops.h"
 
 #define CUDA_REDUCE_WITH_OP(NAME, OP)                                          \
 template <typename T>                                                          \
 __global__ void UCC_REDUCE_CUDA_ ## NAME (const T *s1, const T *s2, T *d,      \
                                           size_t size, size_t count,           \
+                                          size_t ld)                           \
+{                                                                              \
+        size_t start = blockIdx.x * blockDim.x + threadIdx.x;                  \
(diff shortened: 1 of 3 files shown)


src/components/mc/cuda/kernel/mc_cuda_reduce_multi.cu
```diff
@@ -164,6 +184,11 @@ ucc_status_t ucc_mc_cuda_reduce_multi(const void *src1, const void *src2,
             DT_REDUCE_INT(int64_t, op, src1, src2, dst, size, count, ld, stream,
                          bk, th);
             break;
+        case UCC_DT_FLOAT16:
+            ucc_assert(4 == sizeof(__half));
```
vspetrov, line 188: 4->2

vspetrov (changes requested):
gtest?

src/components/mc/cuda/kernel/mc_cuda_reduce_multi.cu
Sergei-Lebedev: copy paste 🤦‍♂️

Sergei-Lebedev:
> gtest?

it would not be easy to add, need conversion functions from fp16 to fp32 and back as well as some function that will check fp16 accuracy for fp32 numbers. There are intrinsics for some cpu arch but then we need config time detection for it. Overall it seems that test will be overcomplicated and we will need tests for test

vspetrov:
> > gtest?
> 
> it would not be easy to add, need conversion functions from fp16 to fp32 and back as well as some function that will check fp16 accuracy for fp32 numbers. There are intrinsics for some cpu arch but then we need config time detection for it. Overall it seems that test will be overcomplicated and we will need tests for test

i meant for GPU, we do test mc reduce for GPU right? we can postpone it for later

vspetrov (approved):

manjugv (approved):

PR #225, opened 2021-06-16
API: Adding details for buffer info

manjugv:
To clarify https://github.com/openucx/ucc/pull/221

### DIFF
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -1382,19 +1382,36 @@ typedef enum ucc_memory_type {
     UCC_MEMORY_TYPE_UNKNOWN = UCC_MEMORY_TYPE_LAST
 } ucc_memory_type_t;
 
+/**
+ *  @ingroup UCC_COLLECTIVES_DT
+ */
 typedef struct ucc_coll_buffer_info_v {
-    void             *buffer;
-    ucc_count_t      *counts;
-    ucc_aint_t       *displacements;
-    ucc_datatype_t    datatype;
-    ucc_memory_type_t mem_type;
+    void             *buffer; /*!< Starting address of the send/recv buffer */
+    ucc_count_t      *counts; /*!< Array of counts of type @ref ucc_count_t
+                                describing the total number of elements */
+    ucc_aint_t       *displacements; /*!< Displacement array of team size and
+                                       type @ref ucc_aint_t. Entry i specifies
+                                       the displacement relative to the start
+                                       address for the incoming data(
+                                       outgoing data) for the team member i. For
+                                       send buffer the data is fetched from this
+                                       displacement and for receive buffer the
+                                       incoming data is placed at this
+                                       displacement. */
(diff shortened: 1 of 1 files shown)


src/ucc/api/ucc.h
```diff
@@ -1381,19 +1381,36 @@ typedef enum ucc_memory_type {
...
+/**
+ *  @ingroup UCC_COLLECTIVES_DT
+ */
 typedef struct ucc_coll_buffer_info {
-    void             *buffer;
-    ucc_count_t       count;
-    ucc_datatype_t    datatype;
-    ucc_memory_type_t mem_type;
+    void             *buffer;   /*!< Starting address of the send/recv buffer */
+    ucc_count_t       count;    /*!< Total number of elements in the buffer */
```
vspetrov, line 1410: @manjugv so, we are agreeing on changing the API semantics here, right? We will need to change/fix ompi/coll/ucc (in master and 5.0.x), ompi/scoll/ucc (master and 5.0.x), torch-ucc, and our internal test: gtest and test-mpi

manjugv: That is correct. This semantic makes sense to me. We could discuss this in the WG, if you have any concerns.

vspetrov: @manjugv last question: am i understanding correctly that we will require user to fill even redundant fields for consistency, ie for allgather user will set both: src.info.count AND dst.info.count (which is N*src.info.count) ? I feel it is the right thing from the consistency perspective?

src/ucc/api/ucc.h
```diff
@@ -1381,19 +1381,36 @@ typedef enum ucc_memory_type {
...
 typedef struct ucc_coll_buffer_info_v {
-    void             *buffer;
-    ucc_count_t      *counts;
-    ucc_aint_t       *displacements;
-    ucc_datatype_t    datatype;
-    ucc_memory_type_t mem_type;
+    void             *buffer; /*!< Starting address of the send/recv buffer */
+    ucc_count_t      *counts; /*!< Array of counts of type @ref ucc_count_t
+                                describing the total number of elements */
+    ucc_aint_t       *displacements; /*!< Displacement array of team size and
```
vspetrov, line 1391: @manjugv do we want to specify that displacements are given in terms of "counts of datatypes" rather than in bytes? Or is it evident enough?

vspetrov (approved):

src/ucc/api/ucc.h
manjugv: I agree, we should require it.

src/ucc/api/ucc.h
manjugv: We haven't used bytes in any of the descriptions, so, I think it is obvious that we are talking about datatype. If you think there is ambiguity. I can change the counts to this "Array of counts of type @ref ucc_count_t describing the total number of **@ref ucc_datatype_t** elements"

manjugv:
Ping @alex--m

Sergei-Lebedev (approved):

PR #227, opened 2021-06-17
CORE: call task complete for schedule

Sergei-Lebedev:
## What
Since schedule is a task need to call task_complete function for it.

## Why ?
Fixing bug when callback function is not called after collective complete


manjugv (approved):

vspetrov (approved):

manjugv:
@Sergei-Lebedev Please port this to the release branch.

PR #228, opened 2021-06-19
Topic/addr exchange at ctx create

vspetrov:
## What
Implements ucc ctx addressing exchange during context creation if OOB is provided there.

## Why ?
1. Unifies address exchanges for any potential CL/TL. This saves time for the team creation.
2. Addresses are stored at CTX once and not duplicated for each team
3. Additional information will be augmented into address (e.g. proc info: pid, socket, hosthash, etc - to build basic topo)

## How ?
If OOB is provided for ucc_context_create then the addresses of all tls/cls are packed and exchanged. Later at team creation time we only exchange proc_info which allows building rank map to ctx ranks (i assume later we can probably provide that from runtime).

Adds necessary gtest cases. Since ucc_context_create is blocking, gtest will implement collective ctx create using threads.


src/core/ucc_context.c
```diff
@@ -328,6 +328,100 @@ static ucc_status_t ucc_create_tl_contexts(ucc_context_t *ctx,
...
+        }
+        memcpy(
+            PTR_OFFSET(addr_storage->storage, max_addrlen * addr_storage->size),
+            context->attr.ctx_addr, context->attr.ctx_addr_len);
+        status = oob->allgather(
+            PTR_OFFSET(addr_storage->storage, max_addrlen * addr_storage->size),
+            addr_storage->storage, max_addrlen, oob->coll_info,
+            &addr_storage->oob_req);
+        if (UCC_OK != status) {
+            ucc_error("failed to start oob allgather\n");
```
Sergei-Lebedev, line 417: Return error

src/core/ucc_context.c
```diff
@@ -415,9 +506,16 @@ ucc_status_t ucc_context_create(ucc_lib_h lib,
         ucc_error("failed to init progress queue for context %p", ctx);
         goto error_ctx_create;
     }
-    ctx->id.host_id = ucc_local_proc.host_id;
-    ctx->id.pid     = getpid();
-    ctx->id.seq_num = ucc_context_seq_num++;
+    ctx->id.pi      = ucc_local_proc;
+    ctx->id.seq_num = ucc_atomic_fadd32(&ucc_context_seq_num, 1);
```
Sergei-Lebedev, line 510: Is atomic really needed? I thought it’s not possible to call ucc control flow api from different threads

src/core/ucc_team.h
```diff
@@ -41,4 +51,44 @@ typedef struct ucc_team {
...
+                                         unsigned long component_id)
+{
+    ucc_context_addr_header_t *h = ucc_get_team_ep_header(context, team, rank);
+    int                        i;
+    for (i = 0; i < h->n_components; i++) {
+        if (h->components[i].id == component_id) {
+            return PTR_OFFSET(h, h->components[i].offset);
+        }
+    }
+    ucc_assert(0);
```
Sergei-Lebedev, line 90: Remove

src/core/ucc_team.c
```diff
@@ -235,11 +224,102 @@ ucc_team_create_cls(ucc_context_t *context, ucc_team_t *team)
...
+            return status;
+        }
+    }
+    status = team->params.oob.req_test(team->oob_req);
+    if (status < 0) {
+        team->params.oob.req_free(team->oob_req);
+        ucc_error("oob req test failed during team proc info exchange");
+        return status;
+    }
+    else if (UCC_INPROGRESS == status) {
```
Sergei-Lebedev, line 274: Move line up

src/components/tl/ucp/tl_ucp_ep_hash.h
```diff
@@ -23,16 +23,15 @@ static inline uint32_t tl_ucp_ctx_id_hash_fn_impl(uint32_t h, uint32_t k)
 static inline khint32_t tl_ucp_ctx_id_hash_fn(ucc_context_id_t k)
 {
     uint32_t h = 0;
-    h = tl_ucp_ctx_id_hash_fn_impl(h, k.host_id);
-    h = tl_ucp_ctx_id_hash_fn_impl(h, k.pid);
+    h          = tl_ucp_ctx_id_hash_fn_impl(h, k.pi.host_hash);
```
Sergei-Lebedev, line 26: Alignment

src/core/ucc_context.h
```diff
@@ -72,9 +81,27 @@ typedef struct ucc_context_config {
...
+   If used at context level the function is blocking, otherwise it
+   is non-blocking and can return UCC_INPROGRESS. In that case it
+   must be called until UCC_OK is returned.
+
+   The addresses are stored in the addr_storage data structure.
+
+   The addressing data of rank "i" (according to OOB) can be accessed
+   with UCC_ADDR_STORAGE_RANK_HEADER macro defined below.
+*/
+ucc_status_t ucc_core_addr_exchange(ucc_context_t          *context,
```
Sergei-Lebedev, line 101: I think it’s cleaner to make this function nonblocking only. And caller should implement blocking behavior if needed.

src/core/ucc_context.c
vspetrov: yes, it turned out to be needed, if a thread  runs a separate ucc context and thus behaves as a "rank" - e.g. this happens in gtest

src/core/ucc_team.h
vspetrov: wanted to make it readable that we never get there - i'll change the assertion on positive

src/components/tl/ucp/tl_ucp_ep_hash.h
vspetrov: that is what "Clang" produced - fixed manually

Sergei-Lebedev: yeah, maybe because there is no blank line between "h" declaration and assignment

src/core/ucc_context.c
bureddy: what is the use case of this ID?

src/core/ucc_team.c
```diff
@@ -235,11 +224,101 @@ ucc_team_create_cls(ucc_context_t *context, ucc_team_t *team)
...
+    }
+    if (is_const_stride) {
+        if ((stride == 1) && (team->size == context->addr_storage.size)) {
+            team->ctx_map.type = UCC_EP_MAP_FULL;
+        }
+        else {
+            team->ctx_map.type           = UCC_EP_MAP_STRIDED;
+            team->ctx_map.strided.start  = (uint64_t)team->ctx_ranks[0];
+            team->ctx_map.strided.stride = (uint64_t)stride;
+        }
```
bureddy, line 299: minor: can we release the memory "ctx_ranks" here in case of const stride?

src/utils/ucc_proc_info.h
```diff
@@ -14,13 +14,18 @@ typedef uint32_t ucc_host_id_t;
 typedef uint32_t ucc_socket_id_t;
 
 typedef struct ucc_proc_info {
-    ucc_host_id_t   host_id;
+    ucc_host_id_t   host_hash;
     ucc_socket_id_t socket_id;
+    ucc_host_id_t   host_id;
```
bureddy, line 19: seems "host_id" is replaced with host_hash. do we nee need this ?

src/core/ucc_context.c
```diff
@@ -328,6 +328,98 @@ static ucc_status_t ucc_create_tl_contexts(ucc_context_t *ctx,
...
+            PTR_OFFSET(addr_storage->storage, max_addrlen * addr_storage->size),
+            context->attr.ctx_addr, context->attr.ctx_addr_len);
+        status = oob->allgather(
+            PTR_OFFSET(addr_storage->storage, max_addrlen * addr_storage->size),
+            addr_storage->storage, max_addrlen, oob->coll_info,
+            &addr_storage->oob_req);
+        if (UCC_OK != status) {
+            ucc_error("failed to start oob allgather");
+            return status;
+        }
```
bureddy, line 416: can we also compute ctx_rank and save it?

src/core/ucc_context.c
vspetrov: it is used to uniquely identify a participant of ucc context (a process) - for example, it is used in tl/ucp as a key to the hash table of the UCP eps.

src/core/ucc_team.c
vspetrov: yes, wanted to do that. this is why i asked if we need the actual "array" anywhere, e.g. in the sharp comm init. looks like it is not needed right now - i will free it.

src/utils/ucc_proc_info.h
vspetrov: host_id will be used in next PRs to build the subgroups for CL/HIER: host id is also a uniq number that identifies the host.  The difference from host_hash is that host_id \in [0, N-1], where N is the total number of nodes - ie., contiguous value - used to build subgroups.

bureddy (approved):

Sergei-Lebedev (approved):

src/core/ucc_team.c
vspetrov: done

PR #229, opened 2021-06-22
MC: check provided thread mode

Sergei-Lebedev:
## What
Fixing multiple bugs in MC:
- checking provided thread mode in ucc_mc_init, print warning if requested thread_mode is not consistent with previously used
- disabled 2 test cases in gtest because they depend on the order in which memory components were initialized
- fix fp16 switch exit
- added debug flags for nvcc host code if --enable-debug is set

### DIFF
--- a/src/core/ucc_mc.c
+++ b/src/core/ucc_mc.c
@@ -42,6 +42,7 @@ ucc_status_t ucc_mc_init(const ucc_mc_params_t *mc_params)
     int            i, n_mcs;
     ucc_mc_base_t *mc;
     ucc_status_t   status;
+    ucc_mc_attr_t attr;
 
     memset(mc_ops, 0, UCC_MEMORY_TYPE_LAST * sizeof(ucc_mc_ops_t *));
     memset(ee_ops, 0, UCC_EE_LAST * sizeof(ucc_ee_ops_t *));
@@ -75,6 +76,18 @@ ucc_status_t ucc_mc_init(const ucc_mc_params_t *mc_params)
                 continue;
             }
             ucc_debug("%s initialized", mc->super.name);
+        } else {
+            attr.field_mask = UCC_MC_ATTR_FIELD_THREAD_MODE;
+            status = mc->get_attr(&attr);
+            if (status != UCC_OK) {
+                return status;
+            }
+            if (attr.thread_mode < mc_params->thread_mode) {
+                ucc_warn("memory component %s was allready initilized with "
+                         "different thread mode: current tm %d, provided tm %d",
+                         mc->super.name, attr.thread_mode,
+                         mc_params->thread_mode);
+            }
         }
         mc->ref_cnt++;
         mc_ops[mc->type] = &mc->ops;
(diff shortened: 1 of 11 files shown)


src/core/ucc_mc.c
```diff
@@ -75,6 +76,18 @@ ucc_status_t ucc_mc_init(const ucc_mc_params_t *mc_params)
                 continue;
             }
             ucc_debug("%s initialized", mc->super.name);
+        } else {
+            attr.field_mask = UCC_MC_ATTR_FIELD_THREAD_MODE;
+            status = mc->get_attr(&attr);
+            if (status != UCC_OK) {
+                return status;
+            }
+            if (attr.thread_mode < mc_params->thread_mode) {
```
vspetrov, line 85: what should we do in this case? just print warn?

test/gtest/core/test_mc.cc
```diff
@@ -82,7 +82,8 @@ UCC_TEST_F(test_mc, can_alloc_and_free_host_mem)
     ucc_mc_finalize();
 }
 
-UCC_TEST_F(test_mc, can_alloc_and_free_host_mem_mt)
+// Disabled because can't reinit mc with different thread mode
+UCC_TEST_F(test_mc, DISABLED_can_alloc_and_free_host_mem_mt)
```
vspetrov, line 86: Why we can't re-init? What is the problem - finalizing mpools ? why can't we finalize them - i assume potential outstanding colls? @manjugv the question to you on a close  topic: do we alloc ucc_context_create being called while there is another outstanding UCC collective (on a team from another ucc context); i remember we don't allow concurrent ucc_ctx_create from multiple threads.

Sergei-Lebedev: it's up to memory component to do everything that needed to reinit correctly. I think we can implement it in one way or another, i just wanted to disable failing test in this pr and print warning to avoid debugging in future

vspetrov: maybe we need to open gh issue then: 1. we need to understand if we need to support such re-init, 2. if we decide we need it we have to track so that we don't forget.

vspetrov (approved):

test/gtest/core/test_mc.cc
Sergei-Lebedev: sure, I think it's not P0 task since creating multiple ucc instances with different thread levels is something rare

manjugv: > Why we can't re-init? What is the problem - finalizing mpools ? why can't we finalize them - i assume potential outstanding colls? @manjugv the question to you on a close topic: do we alloc ucc_context_create being called while there is another outstanding UCC collective (on a team from another ucc context); i remember we don't allow concurrent ucc_ctx_create from multiple threads.

I think we should allow context create by multiple threads, when library is initialized the thread multiple mode. Given that it is a blocking call and can be local operation, is there a reason not to?

vspetrov: The key here is "can be local" - you will have to specify exactly which combinations of params are allowed then for concurrent context creation. Obviously, it will not be allowed when OOB is provided.

manjugv (approved):

PR #235, opened 2021-06-23
Topic/persistent

vspetrov:
## What
Adds testing of persistent mode (multiple collective restart w/o finalize/init) into gtest. Fixes current TL/UCP collective implementations to correctly work in that case.

## Why ?
Firstly, need to support persistent colls as part of API
Secondly, we need "restartable" tasks for pipelined schedules (coming in next PRs)

### DIFF
--- a/src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
+++ b/src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
@@ -93,9 +93,24 @@ ucc_status_t ucc_tl_ucp_alltoallv_pairwise_start(ucc_coll_task_t *coll_task)
 
     UCC_TL_UCP_PROFILE_REQUEST_EVENT(coll_task, "ucp_alltoallv_pairwise_start",
                                      0);
-    task->super.super.status = UCC_INPROGRESS;
-    task->n_polls            = ucc_min(1, task->n_polls);
+    ucc_tl_ucp_task_reset(task);
 
+    ucc_tl_ucp_alltoallv_pairwise_progress(&task->super);
+    if (UCC_INPROGRESS == task->super.super.status) {
+        ucc_progress_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+        return UCC_OK;
+    }
+    return ucc_task_complete(coll_task);
+}
+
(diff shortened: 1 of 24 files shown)


src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
```diff
@@ -93,9 +93,9 @@ ucc_status_t ucc_tl_ucp_alltoallv_pairwise_start(ucc_coll_task_t *coll_task)
 
     UCC_TL_UCP_PROFILE_REQUEST_EVENT(coll_task, "ucp_alltoallv_pairwise_start",
                                      0);
-    task->super.super.status = UCC_INPROGRESS;
-    task->n_polls            = ucc_min(1, task->n_polls);
+    ucc_tl_ucp_task_reset(task);
 
+    task->n_polls = ucc_min(1, task->n_polls);
     if (UCC_TL_UCP_TEAM_CTX(team)->cfg.pre_reg_mem) {
```
Sergei-Lebedev, line 99: registration can be done once in init

Sergei-Lebedev (commented):
What about NCCL TL? I think persistent colls are not supported there
In TL UCP triggered_post also creates some internal tasks to work with streams, need to reset them as well in order to support persistent triggered post

src/components/tl/ucp/alltoallv/alltoallv_pairwise.c
vspetrov: done

vspetrov:
> What about NCCL TL? I think persistent colls are not supported there
> In TL UCP triggered_post also creates some internal tasks to work with streams, need to reset them as well in order to support persistent triggered post

Regarding NCCL:
if i understand correctly then our implementation DOES support persistent interface (w/o any optimization there): we reset task->status when we "post" tl nccl collective, and then it is launched into nccl kernel. So, the same task can be restarted any number of times - it will just be new real nccl collective each time.

On triggered: right now it works, because triggered task is allocated each time in "triggered post". In next PR i'm moving implementation of "triggered_post" to CORE (so that CL/HIER could re-use it), we can add some "persistent - triggered test". BTW currently we have 0 "triggered" tests.

manjugv (approved):

Issue #234, opened 2021-06-23
[MTT] infinity log with cuda errors


avildema:
UCX: 1.11
UCC: master
OMPI: v5.0.x

setup: GPU (cuda)

infinity print to log

```sh
[1624426315.715236] [vulcan02:9197 :0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
[1624426315.715239] [vulcan02:9197 :0] reduce_scatter_knomial.c:151  TL_UCP ERROR failed to perform dt reduction
[1624426315.715244] [vulcan02:9197 :0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
[1624426315.715247] [vulcan02:9197 :0] reduce_scatter_knomial.c:151  TL_UCP ERROR failed to perform dt reduction
[1624426315.715252] [vulcan02:9197 :0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
[1624426315.715256] [vulcan02:9197 :0] reduce_scatter_knomial.c:151  TL_UCP ERROR failed to perform dt reduction
[1624426315.715260] [vulcan02:9197 :0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
[1624426301.703205] [vulcan04:18934:0] reduce_scatter_knomial.c:151  TL_UCP ERROR failed to perform dt reduction
[1624426301.703210] [vulcan04:18934:0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
[1624426301.703214] [vulcan04:18934:0] reduce_scatter_knomial.c:151  TL_UCP ERROR failed to perform dt reduction
[1624426301.703218] [vulcan04:18934:0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
[1624426301.703221] [vulcan04:18934:0] reduce_scatter_knomial.c:151  TL_UCP ERROR failed to perform dt reduction
[1624426301.703226] [vulcan04:18934:0] mc_cuda_reduce_multi.cu:206  cuda mc ERROR cuda failed with ret:400(invalid resource handle)
```

avildema:
cmd for reproduce
```sh
   
[...]

PR #238, opened 2021-06-24
MC/CUDA: init stream on demand

Sergei-Lebedev:
## What
Fix for #234

1. Initialize cuda stream only when it's needed for the first time, that way we guarantee that we pick correct device for stream (fixing cuda error)

2. Introduce error event handler and UCC_EVENT_ERROR to propagate error from individual task to schedule (fixing infinity progress after error)

### DIFF
--- a/src/components/mc/cuda/mc_cuda.h
+++ b/src/components/mc/cuda/mc_cuda.h
@@ -67,7 +67,7 @@ typedef struct ucc_mc_cuda {
     ucc_mpool_t                    strm_reqs;
     ucc_mpool_t                    mpool;
     int                            mpool_init_flag;
-    ucc_spinlock_t                 mpool_init_spinlock;
+    ucc_spinlock_t                 init_spinlock;
     ucc_thread_mode_t              thread_mode;
     ucc_mc_cuda_strm_task_mode_t   strm_task_mode;
     ucc_mc_cuda_task_stream_type_t task_strm_type;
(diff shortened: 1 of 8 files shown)


src/components/mc/cuda/mc_cuda.h
```diff
@@ -136,4 +136,21 @@ extern ucc_mc_cuda_t ucc_mc_cuda;
 #define MC_CUDA_CONFIG                                                         \
     (ucc_derived_of(ucc_mc_cuda.super.config, ucc_mc_cuda_config_t))
 
+#define UCC_MC_CUDA_INIT_STREAM() do {                                         \
+    if (ucc_mc_cuda.stream == NULL) {                                          \
+        cudaError_t cuda_st;                                                   \
+        ucc_spin_lock(&ucc_mc_cuda.init_spinlock);                             \
+        if (ucc_mc_cuda.stream == NULL) {                                      \
+            cuda_st = cudaStreamCreateWithFlags(&ucc_mc_cuda.stream,           \
```
bureddy, line 144: should we create high_priority stream with cudaStreamCreateWithPriority()

Sergei-Lebedev: yes, but I believe it should be an option rather than creating high priority steam always.
Do you mean make it part of this PR?

bureddy: lets do it in another PR

bureddy (approved):

vspetrov (approved):

manjugv (approved):

PR #239, opened 2021-06-28
SCHEDULE: rework listeners

vspetrov:
## What
Change the implementation of event manager to support subscription on the same event with different handlers for different parent tasks

## Why ?
When we will build more complex algorithms (upcoming PRs in CL/HIER, 2 lvl allreduce alg) we will have e.g. schedule of schedules: e.g., 1 sra schedule launched after another task_T. The SRA schedule subscribes to its tasks on EVENT completed but also needs to subscribe with different handler to the task_T.

### DIFF
--- a/src/core/ucc_progress_queue_mt.c
+++ b/src/core/ucc_progress_queue_mt.c
@@ -69,10 +69,7 @@ static int ucc_pq_mt_progress(ucc_progress_queue_t *pq)
     pq->dequeue(pq, &task);
     if (task) {
         if (task->progress) {
-            status = task->progress(task);
-            if (ucc_unlikely(status < 0)) {
-                return ucc_task_error(task);
-            }
+            task->progress(task);
         }
         if (UCC_INPROGRESS == task->super.status) {
             pq->enqueue(pq, task);
--- a/src/schedule/ucc_schedule.c
+++ b/src/schedule/ucc_schedule.c
@@ -8,18 +8,26 @@
 ucc_status_t ucc_event_manager_init(ucc_event_manager_t *em)
 {
     int i;
-    for (i = 0; i < UCC_EVENT_LAST; i++) {
-        em->listeners_size[i] = 0;
+    for (i = 0; i < MAX_LISTENERS; i++) {
+        em->listeners[i].task = NULL;
     }
     return UCC_OK;
 }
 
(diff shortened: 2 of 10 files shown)


manjugv (approved):

manjugv:
@vspetrov The PR looks good. Currently, the scalability of the event manager is with listeners, and another way would be with events (IIRC, Sergey raised this point during the WG discussion)? Did you conclude on scaling with listeners being a better choice?

vspetrov:
@Sergei-Lebedev @bureddy plz review

vspetrov:
> @vspetrov The PR looks good. Currently, the scalability of the event manager is with listeners, and another way would be with events (IIRC, Sergey raised this point during the WG discussion)? Did you conclude on scaling with listeners being a better choice?

I don't think there is a big concern in terms of "scaling". It does not depend on actual scale of job (number of processes). This defines how many tasks can subscribe to each other in event manager, which is entirely the property of our algorithms. With proper algs design this number will not go high. On the other hand with the former implementation (that you call "scaling with events") we just could not support necessary functionality of serving different handlers to the same event from different parent tasks (needed in several places). So, the change is just necessary. The concern about increasing number of listeners due to "error handler event" is resolved in this PR but adding a special error handling flow.

src/schedule/ucc_schedule.c
```diff
@@ -31,6 +39,26 @@ ucc_status_t ucc_coll_task_init(ucc_coll_task_t *task)
     return ucc_event_manager_init(&task->em);
 }
 
+static ucc_status_t
+ucc_task_error_handler(ucc_coll_task_t *parent_task, //NOLINT
```
Sergei-Lebedev, line 43: remove NOLINT

src/core/ucc_progress_queue_mt.c
```diff
@@ -70,9 +70,6 @@ static int ucc_pq_mt_progress(ucc_progress_queue_t *pq)
     if (task) {
         if (task->progress) {
             status = task->progress(task);
-            if (ucc_unlikely(status < 0)) {
```
Sergei-Lebedev, line 73:
```
core/ucc_progress_queue_mt.c:72:13: warning: Value stored to 'status' is never read [clang-analyzer-deadcode.DeadStores]
            status = task->progress(task);
```

Sergei-Lebedev (approved):

vspetrov:
bot:retest

PR #248, opened 2021-07-07
Blocking a2a self copy

bureddy:
`UCC_TL_UCP_BLOCKING_A2A_SELF_COPY=1` to do blocking self copy in alltoall and alltoallv before send/recv from remote peers

### DIFF
--- a/src/components/tl/ucp/alltoall/alltoall_pairwise.c
+++ b/src/components/tl/ucp/alltoall/alltoall_pairwise.c
@@ -75,6 +75,11 @@ ucc_status_t ucc_tl_ucp_alltoall_pairwise_start(ucc_coll_task_t *coll_task)
 {
     ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
     ucc_tl_ucp_team_t *team = task->team;
+    ptrdiff_t          sbuf = (ptrdiff_t)task->args.src.info.buffer;
+    ptrdiff_t          rbuf = (ptrdiff_t)task->args.dst.info.buffer;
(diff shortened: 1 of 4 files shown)


src/components/tl/ucp/alltoall/alltoall_pairwise.c
```diff
@@ -89,10 +94,24 @@ ucc_status_t ucc_tl_ucp_alltoall_pairwise_start(ucc_coll_task_t *coll_task)
...
+    if ((UCC_TL_UCP_TEAM_CTX(team)->cfg.blocking_a2a_self_copy)) {
+        data_size = (size_t)task->args.src.info.count *
+                    ucc_dt_size(task->args.src.info.datatype);
+        UCPCHECK_GOTO(ucc_tl_ucp_recv_nb((void *)(rbuf + grank * data_size),
+                                          data_size, rmem, grank, team, task),
+                                          task, error);
+        UCPCHECK_GOTO(ucc_tl_ucp_send_nb((void *)(sbuf + grank * data_size),
+                                          data_size, smem, grank, team, task),
+                                          task, error);
+    }
```
vspetrov, line 107: @bureddy looks like these send/recv to self are still non-blocking. Is that the expected behavior we wanted? don't we need a tight while{} loop here?

bureddy: in xccl also it is non-blocking api but we have a testall() in while loop. . the main thing is we finish self-copy before send/recv other ranks. should be ok. but will change it to  busy loop

bureddy: fixed.

PR #251, opened 2021-07-08
TEST: disable NCCL and GDR

vspetrov:
## What
Disables NCCL and GDR in gtest

## Why ?
NCCL is not supported due to blocking team creation
GDR currently may result in a hang (known issue)

## How ?
internally sets env vars: UCC_TL_NCCL_TUNE=0 and UCX_IB_GPU_DIRECT_RDMA=no

### DIFF
--- a/test/gtest/common/test_ucc.cc
+++ b/test/gtest/common/test_ucc.cc
@@ -247,6 +247,14 @@ UccJob::UccJob(int _n_procs, ucc_job_ctx_mode_t _ctx_mode, ucc_job_env_t vars) :
 {
     ucc_job_env_t env_bkp;
     char *var;
+
+    /* NCCL TL is disabled since it currently can not support non-blocking
+       team creation. */
+    vars.push_back({"UCC_TL_NCCL_TUNE", "0"});
+    /* GDR is temporarily disabled due to known issue that may result
+       in a hang in the destruction flow */
+    vars.push_back({"UCX_IB_GPU_DIRECT_RDMA", "no"});
+
     for (auto &v : vars) {
         var = std::getenv(v.first.c_str());
         if (var) {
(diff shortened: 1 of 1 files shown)


test/gtest/common/test_ucc.cc
```diff
@@ -247,6 +247,14 @@ UccJob::UccJob(int _n_procs, ucc_job_ctx_mode_t _ctx_mode, ucc_job_env_t vars) :
 {
     ucc_job_env_t env_bkp;
     char *var;
+
+    /* NCCL TL is disabled since it currently can not support non-blocking
+       team creation. */
+    vars.push_back({"UCC_TL_NCCL_TUNE", "0"});
```
bureddy, line 253: is it a temporary fix to disable NCCL also? is there way to going forward fix to testing if non-blocking team create not support

vspetrov: it is temporary until we figure out how to run NCCL from single process (not sure if that is possible at all). Gtest is not that well suited for that case

bureddy (approved):

test/gtest/common/test_ucc.cc
```diff
@@ -247,6 +247,14 @@ UccJob::UccJob(int _n_procs, ucc_job_ctx_mode_t _ctx_mode, ucc_job_env_t vars) :
...
+    /* GDR is temporarily disabled due to known issue that may result
+       in a hang in the destruction flow */
+    vars.push_back({"UCX_IB_GPU_DIRECT_RDMA", "no"});
```
bureddy, line 256: instead of disabling GDR here, is it ok to disable stop nv_peer_mem on test server?

Sergei-Lebedev (approved):

vspetrov:
bot:retest

artemry-nv:
bot:retest

PR #258, opened 2021-07-14
TL/NCCL: change task status from progress only

Sergei-Lebedev:
## What
In case of  memops sync task status need to be changed only during collective progress. Otherwise task might never be dequeued from progress queue


vspetrov (approved):

bureddy (approved):

vspetrov:
bot:retest

PR #261, opened 2021-07-14
MC/CPU: openmp reduction

Sergei-Lebedev:
## What
Use openmp to do reduction in parallel.

## How ?
During configure time user can add --enable-openmp option ( by default disabled) to use parallel reduction. Number of threads controlled by UCC_MC_CPU_REDUCE_NUM_THREADS variable

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -69,6 +69,8 @@ AC_MSG_RESULT([git sha: $ucc_git_sha])
 CFLAGS_save="$CFLAGS"
 AC_PROG_CC
 AC_PROG_CXX
+: ${enable_openmp=no}
+AC_OPENMP
 AM_PROG_AS
 AC_PROG_LN_S
 AC_PROG_MKDIR_P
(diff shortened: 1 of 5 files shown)


manjugv:
@Sergei-Lebedev Do we really want to create dependency on openmp ? Also, we should not enable this in thread single mode.

Sergei-Lebedev:
> @Sergei-Lebedev Do we really want to create dependency on openmp ? Also, we should not enable this in thread single mode.

What do you mean by dependency? openmp is not required and by default disabled, it's only optimization that user can turn off.
Why thread multiple needed?

manjugv:
> > @Sergei-Lebedev Do we really want to create dependency on openmp ? Also, we should not enable this in thread single mode.
> 
> What do you mean by dependency? openmp is not required and by default disabled, it's only optimization that user can turn off.
> Why thread multiple needed?

It is still a dependency :) Can you please explain the motivation, so we can explore alternatives?

janjust:
@manjugv why would thread-multiple be an issue here, it only parallelizes compute loops.

PR #268, opened 2021-07-19
MC: fixed mc reduce multi parameters order

shimmybalsam:
## What
Switch the order between size and count parameters in ucc_mc_reduce_multi function decleration in ucc_mc.h

## Why ?
To fit to the order in which the function is implemeted in ucc_mc.c and in which it is called from ucc_dt_reduce_multi. 

## How ?
Switched size and count.

### DIFF
--- a/src/components/mc/base/ucc_mc_base.h
+++ b/src/components/mc/base/ucc_mc_base.h
@@ -106,7 +106,7 @@ typedef struct ucc_mc_ops {
                            size_t count, ucc_datatype_t dt,
                            ucc_reduction_op_t op);
     ucc_status_t (*reduce_multi)(const void *src1, const void *src2, void *dst,
-                                 size_t count, size_t size, size_t stride,
+                                 size_t n_vectors, size_t count, size_t stride,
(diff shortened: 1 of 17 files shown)


lappazos (approved):

vspetrov (commented):
Ok, so if we want to improve readability i would change "size" in the declaration to "n_vectors". It is more clear imho. And then also need to change in implementations for consistency

lappazos:
> Ok, so if we want to improve readability i would change "size" in the declaration to "n_vectors". It is more clear imho. And then also need to change in implementations for consistency

It's not improved readability, it is a mistake that the compiler passed only since both are the same type. I agree, but need to change also ucc_dt_reduce_multi variable name.

shimmybalsam:
> Ok, so if we want to improve readability i would change "size" in the declaration to "n_vectors". It is more clear imho. And then also need to change in implementations for consistency

All changed. Please check.

vspetrov:
bot:retest

vspetrov (approved):

Sergei-Lebedev (approved):

artemry-nv:
bot:retest

PR #274, opened 2021-07-23
Topic/service colls

vspetrov:
## What
- Optimize team creation: don't perform service team allocation or team_id allocation during ucc_team_create if CL/TL components don't report they need those features
- Add service allgather at tl interface
- Add service allgather implementation in the TL/UCP via allgather ring (adds support for subsets to ring allgather alg)
- Adds service coll helper layer in CORE
- Adds an option (UCC_OOB_INTERNAL) that allocates a single global service team per ucc_context (instead of 1 service team per ucc team). That team is then used for oob during CL/TL team creations (as well as team_id alloc) instead of runtime OOB. This can be useful when runtime OOB allgather is not efficient (e.g. implemented over TCP). This option can be used when context creation is in GLOBAL mode (with OOB). Still 1 allgather is made via runtime oob durin team_create - the one that builds context rank map. @manju i still insist we should expose the mapping via our UCC API. In that case i can eliminate this last OOB allgather (this is for another discussion, regardless of this PR)
- Adds gtests for service collectives

## Why ?
Optimizes team creation by: (1) not performing unneeded exchanges when possible (e.g. if only NCCL TL is used, then team_id is not required) (2) switching to internal OOB via UCX

### DIFF
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -84,16 +84,14 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
 
     {"NPOLLS", "10",
      "Number of ucp progress polling cycles for p2p requests testing",
-     ucc_offsetof(ucc_tl_ucp_context_config_t, n_polls),
-     UCC_CONFIG_TYPE_UINT},
+     ucc_offsetof(ucc_tl_ucp_context_config_t, n_polls), UCC_CONFIG_TYPE_UINT},
 
     {"OOB_NPOLLS", "20",
      "Number of polling cycles for oob allgather and service coll request",
      ucc_offsetof(ucc_tl_ucp_context_config_t, oob_npolls),
      UCC_CONFIG_TYPE_UINT},
 
-    {"PRE_REG_MEM", "0",
-     "Pre Register collective memory region with UCX",
+    {"PRE_REG_MEM", "0", "Pre Register collective memory region with UCX",
      ucc_offsetof(ucc_tl_ucp_context_config_t, pre_reg_mem),
      UCC_CONFIG_TYPE_UINT},
 
@@ -132,9 +130,10 @@ ucc_status_t ucc_tl_ucp_service_allreduce(ucc_base_team_t *team, void *sbuf,
                                           ucc_tl_team_subset_t subset,
                                           ucc_coll_task_t **task);
 
-ucc_status_t ucc_tl_ucp_service_allgather(ucc_base_team_t *team, void *sbuf, void *rbuf,
-                                          size_t msgsize, ucc_tl_team_subset_t subset,
-                                          ucc_coll_task_t **task_p);
+ucc_status_t ucc_tl_ucp_service_allgather(ucc_base_team_t *team, void *sbuf,
+                                          void *rbuf, size_t msgsize,
+                                          ucc_tl_team_subset_t subset,
+                                          ucc_coll_task_t    **task_p);
 
 ucc_status_t ucc_tl_ucp_service_test(ucc_coll_task_t *task);
 
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -88,7 +88,7 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      UCC_CONFIG_TYPE_UINT},
 
     {"OOB_NPOLLS", "20",
-     "Number of polling cycles for oob allgather request",
+     "Number of polling cycles for oob allgather and service coll request",
      ucc_offsetof(ucc_tl_ucp_context_config_t, oob_npolls),
      UCC_CONFIG_TYPE_UINT},
 
@@ -131,6 +131,10 @@ ucc_status_t ucc_tl_ucp_service_allreduce(ucc_base_team_t *team, void *sbuf,
                                           size_t count, ucc_reduction_op_t op,
                                           ucc_tl_team_subset_t subset,
                                           ucc_coll_task_t **task);
+
+ucc_status_t ucc_tl_ucp_service_allgather(ucc_base_team_t *team, void *sbuf, void *rbuf,
+                                          size_t msgsize, ucc_tl_team_subset_t subset,
+                                          ucc_coll_task_t **task_p);
 
 ucc_status_t ucc_tl_ucp_service_test(ucc_coll_task_t *task);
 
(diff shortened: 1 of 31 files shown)


vspetrov:
bot:retest

vspetrov:
bot:retest

src/core/ucc_context.c
```diff
@@ -15,26 +15,35 @@
...
+     "team ids for a single process is team_ids_pool_size*64. This parameter "
+     "is relevant when internal team id allocation takes place.",
+     ucc_offsetof(ucc_context_config_t, team_ids_pool_size),
+     UCC_CONFIG_TYPE_UINT},
+
+    {"INTERNAL_OOB", "1",
+     "Use internal OOB transport for team creation. Available for ucc_context "
+     "is configured with OOB (global mode). 0 - disable, 1 - try, 2 - force.",
+     ucc_offsetof(ucc_context_config_t, internal_oob), UCC_CONFIG_TYPE_UINT},
+
```
manjugv, line 45: 1. How about this? Modify the team params to introduce 
   a) User-defined OOB
   b) Use OOB from the context
   c) Use any OOB

Currently, we provide a) and c), but not b). I’m thinking making this dependency explicit is better than having it implicit. I’m not entirely sure, but this might also help your other proposal, where you want relation context participants and team participants. This will be advantageous if we create many teams from the same context. Otherwise, it is a burden on the programmer to handle another semantic. 

2. Once we resolve (1), is there a need for this environment variable? Using internal OOB or not becomes a purely internal decision.

manjugv:
bot:retest

Sergei-Lebedev (approved):

src/ucc/api/ucc.h
```diff
@@ -635,15 +635,18 @@ enum ucc_context_attr_field {
...
+typedef struct ucc_oob_coll {
     ucc_status_t    (*allgather)(void *src_buf, void *recv_buf, size_t size,
                                  void *allgather_info,  void **request);
     ucc_status_t    (*req_test)(void *request);
     ucc_status_t    (*req_free)(void *request);
     uint32_t        participants;
     void            *coll_info;
-}  ucc_context_oob_coll_t;
+    uint32_t        rank;
+}  ucc_oob_coll_t;
```
manjugv, line 646: - Naming - How about rank -> oob_ep ? Since there is a relation between ep (of team) and the “rank” here. 
- Also, participants -> num_oob_eps ?
- We need to add a constraint saying that “rank” is unique among the participants.
- We need to add a constraint saying that if a certain oob is used for context create, the same oob should be used for team create.
- We need to define the relation between “rank” and team’s “ep”, when we create the team.
- What would be a relation with “rank” and context’s ctx_id ? No relation?

src/components/tl/ucp/tl_ucp.c
```diff
@@ -129,9 +127,14 @@ ucc_status_t ucc_tl_ucp_populate_rcache(void *addr, size_t length,
...
                                           void *rbuf, ucc_datatype_t dt,
                                           size_t count, ucc_reduction_op_t op,
-                                          ucc_tl_team_subset_t subset,
+                                          ucc_team_subset_t subset,
                                           ucc_coll_task_t **task);
 
+ucc_status_t ucc_tl_ucp_service_allgather(ucc_base_team_t *team, void *sbuf,
+                                          void *rbuf, size_t msgsize,
+                                          ucc_team_subset_t subset,
+                                          ucc_coll_task_t    **task_p);
```
bureddy, line 136: extra spaces before variable name

src/core/ucc_context.c
```diff
@@ -475,6 +485,9 @@ ucc_status_t ucc_context_create(ucc_lib_h lib,
     b_params.estimated_num_ppn = config->estimated_num_ppn;
     b_params.prefix            = lib->full_prefix;
     b_params.thread_mode       = lib->attr.thread_mode;
+    if (params->mask & UCC_CONTEXT_PARAM_FIELD_OOB) {
+        ctx->rank = params->oob.oob_ep;
```
bureddy, line 489: is ctx->rank re-calculated and set again later (at line:579) ?

src/ucc/api/ucc.h
vspetrov: > * Naming - How about rank -> oob_ep ? Since there is a relation between ep (of team) and the “rank” here.
Done
> * Also, participants -> num_oob_eps ?
Done
> * We need to add a constraint saying that “rank” is unique among the participants.
Added desc
> * We need to add a constraint saying that if a certain oob is used for context create, the same oob should be used for team create.
Why? team oob is general is different from ctx oob (e.g., team oob can be a subset of ctx oob). We wanted to add another constraint: if team is created from multiple "GLOBAL" contexts then ALL those ctxs must have same oob. Is that what you mean here?
> * We need to define the relation between “rank” and team’s “ep”, when we create the team.
This is most interesting one. Basically, these are now 2 ways how user/runtime can pass "the rank of a process in a team" down to ucc_team_create_post: 1st way - EP_CONTIG + ep as input int parameter; 2nd way as team_params.oob.oob_ep. 
I think we should allow all combinations:
1. None is provided. No EP/EP_RANGE_CONTIG, no OOB. Then we compute EP internally and return as output.
2. Either one is provided: either EP w/o OOB (team oob is not necessary if ctx has oob and ep_map is provided for team_create) ; OR OOB is provided w/o EP - then EP is taken from oob.oob_ep
3. Both are provided. Then we demand that the input is consistent, ie oob.oob_ep == EP contig value.
4. If value of EP is provided and it is NOT contig, then it will be used as EP for the API (but it is up to internal implementation to utilize oob_ep if OOB is provided).
@manjugv does it make sense?
> * What would be a relation with “rank” and context’s ctx_id ? No relation?
Yes, no relation.

src/core/ucc_context.c
vspetrov: good catch. i've added the "assert" but forgot to remove the line 579.

src/utils/ucc_datastruct.h
```diff
@@ -5,10 +5,17 @@
...
-
+#include <ucc/api/ucc.h>
 #include <ucs/datastruct/list.h>
 #include <stdint.h>
+
 #define UCC_LIST_HEAD UCS_LIST_HEAD
 typedef uint32_t ucc_rank_t;
 #define UCC_RANK_MAX UINT32_MAX
+
+typedef struct ucc_team_subset {
```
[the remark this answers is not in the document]
vspetrov: The definition is used in ucc_tl.h. If put it to ucc_team.h then i need to include ucc_team.h in the ucc_tl.h which makes it aware of core team. this is why wanted to keep in some other header.

src/ucc/api/ucc.h
```diff
@@ -635,15 +635,18 @@ enum ucc_context_attr_field {
...
 
+typedef ucc_oob_coll_t ucc_context_oob_coll_t;
+typedef ucc_oob_coll_t ucc_team_oob_coll_t;
```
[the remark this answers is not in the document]
vspetrov: can't do that because then the def of the whole struct ucc_oob_coll_t should go to def as well. Maybe remove team/context oob altogether, keeping just ucc_oob_coll_t? @manjugv

manjugv (approved):

bureddy (approved):

PR #279, opened 2021-07-28
Topic/node hash fix

vspetrov:
## What
Fixes issues with gethostid() which may return non-uniq values if not configured properly (FB hang issue)
Additionally: 
- adds check for the uniq proc_info identifiers and prints error in case of collision (unlikely event after that fix)
- adds array storage for TL/UCP ucp_ep_h instead of khash that can be used when OOB is global and tl team has ctx_map

## How ?
Compute the 64 bit node hash using string hash function (todo: switch to UCS ucs_machine_guid when it is made public).


vspetrov:
@Sergei-Lebedev @bureddy plz review

bureddy (approved):

Sergei-Lebedev (approved):

PR #259, opened 2021-07-14
Adjust AllGather, AlltoAll count

lappazos:
## What
Changes AllGather dst.count for the number of elements in the entire dst buffer, for all ranks combined.
Changes AlltoAll dst.count & src.count for the number of elements in the entire src & dst buffers, for all ranks combined.

## Why ?
API Change in PR #225 

## How ?
Change implementation in TL and use cases in tests.

PR #280, opened 2021-07-29
V0.1.x: Adjust allgather, alltoall count

lappazos:
V0.1.x of #259


manjugv (approved):

vspetrov:
re-opened as #281 with build fix

PR #293, opened 2021-09-03
Change coll args rules for allreduce, reduce_scatter and reduce

Sergei-Lebedev:
## What
Changing coll args fields filling rules for some collectives.

## Why ?
Current spec is not consistent and has errors. For example it's not specified how to set datatype for inplace reduce for root rank.

## How ?
In new scheme we set datatype, count and memory type for all user provided buffers.


vspetrov (approved):

src/components/tl/ucp/reduce/reduce.c
```diff
@@ -13,21 +13,27 @@ ucc_status_t ucc_tl_ucp_reduce_knomial_finalize(ucc_coll_task_t *task);
...
     ucc_rank_t         vrank     = (myrank - root + team_size) % team_size;
-    ucc_memory_type_t  mtype     = args->src.info.mem_type;
     ucc_status_t       status    = UCC_OK;
-    size_t             data_size =
-        args->src.info.count * ucc_dt_size(args->src.info.datatype);
+    ucc_memory_type_t  mtype;
+    size_t             data_size;
     int                isleaf;
 
+    if (root == myrank) {
```
lappazos, line 27: i think its more readable if you'll do
ucc_coll_buffer_info_t info = (root == myrank) ? args->dst.info ? args->src.info 

and then just do the rest once

src/components/tl/ucp/reduce/reduce_knomial.c
```diff
@@ -18,27 +18,37 @@
...
-    void              *scratch_offset;
+    ucc_memory_type_t  mtype;
+    ucc_datatype_t     dt;
+    size_t             count, data_size;
+    void              *received_vectors, *scratch_offset;
     ucc_rank_t         vpeer, peer, vroot_at_level, root_at_level, pos;
     uint32_t           i;
     ucc_status_t       status;
 
+    if (root == myrank) {
```
lappazos, line 39: same as above

lappazos (commented):
Just small comment

lappazos:
@Sergei-Lebedev there is a mistake in PR name, you wrote Reduce twice

src/components/tl/ucp/reduce/reduce_knomial.c
Sergei-Lebedev: if use "?" instead of if else i will need to write same ternary operator thing 4 times and here it's just 1 if

bureddy (approved):

manjugv:
@Sergei-Lebedev I'm not sure about the context of the PR. Let's discuss this during the WG. The IN_PLACE flag was defined to primarily support MPI, so, it makes to be consistent with MPI semantics.

Sergei-Lebedev:
@manjugv updated coll args table in docs

manjugv (approved):

PR #294, opened 2021-09-06
fix: --with-docs-only to --with-only-docs

TunchinKao:
## What
fixing --with-docs-only to --with-only-docs
## Why ?
Because I encounter this error while trying to build the docs.
`configure: error: --with-only-docs was requested, but doxygen was not found`
Here is my running command 
```
./autogen.sh
 ./configure --prefix=/tmp/ucc --with-docs-only
```

And it becomes good after switch to --with-only-docs

### DIFF
--- a/.ci/scripts/install_torch.sh
+++ b/.ci/scripts/install_torch.sh
@@ -33,5 +33,5 @@ set -o pipefail
 #conda install pytorch cudatoolkit=11.0 -c pytorch-nightly
 
 pip3 install --default-timeout=900 numpy
-pip3 install --default-timeout=900 --pre torch -f https://download.pytorch.org/whl/nightly/cu110/torch_nightly.html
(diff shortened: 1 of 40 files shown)


mellanox-github:

Sergei-Lebedev:
@TunchinKao thanks for the PR. I think you indeed found a bug, but the fix is not correct.
When you configure with "--with-docs-only" option you need to have doxygen installed. So error message "configure: error: ***--with-only-docs*** was requested, but doxygen was not found" is almost good, except it should be "configure: error: ***--with-docs-only*** was requested, but doxygen was not found".
Correct fix will be to change error message in configure script https://github.com/openucx/ucc/blob/d6eb20d7703d95c91223f7f29c01ae4a9db404a9/configure.ac#L129

TunchinKao:
@Sergei-Lebedev Thanks for your comment, I just found my logic mistake. XD. I will close this one and create a new correct PR.

PR #297, opened 2021-09-08
Scoremap selection fallback

vspetrov:
## What
Implements automatic score-based fallback logic for TL/CL coll_init flow.

## Why ?
Currently when multiple TLs (CLs) are used (e.g. NCCL + UCP via CL/basic) the selection is done based on highest component score (e.g., NCCL can be chosen for cuda based colls). However, if some TL (CL) did not support certain configuration of coll_args (e.g. TL/NCCL does not support allreduce with int16_t datatype) then user would get UCC_ERR_NOT_SUPPORTED. 

Same will be used e.g. for SHARP TL for unsupported dtypes/ops. CL/HIER will also use that when some hierarchical algorithms (e.g. hybrid AR) can not support team topology, etc.

## How ?
This PR implements support for fallbacks in the scoring framework. Nothing changes in the CL/TL - everything is inside coll_score and score_map. Fallbacks are stored in the score-sorted order. So now, when NCCL can't run allreduce int16_t it will automatically fallback to TL/UCP and user will not get error. If user wants to force NCCL only this is of course always an option. 

Looks like this:
```
nnodes=1; ppn=2; mpirun  -x UCC_LOG_LEVEL=debug --map-by ppr:$ppn:node $WDIR/dgx_set_dev.sh ./install/bin/ucc_test_mpi -c allreduce   -t world 32 -M cuda -o sum -d int16 -m 32768:32768
(... part of output from 1 rank ...)
27811:0]    tl_nccl_coll.c:237  TL_NCCL DEBUG dataype is not supported
[1631130400.958187] [swx-dgx01:27811:0] ucc_coll_score_map.c:88   UCC  DEBUG coll is not supported for TL_NCCL, fallback TL_UCP
[1631130400.958208] [swx-dgx01:27811:0]        ucc_coll.c:160  UCC  DEBUG coll_init: team id 1 size 2 rank 0 ctx_rank 0: Allreduce Cuda inplace=0 bytes=32768 int16 sum, req 0x3210f00
tc=Allreduce team=world msgsize=32768 inplace=0 dt=int16 op=sum
[1631130400.958235] [swx-dgx01:27811:0]        ucc_coll.c:170  UCC  DEBUG coll_post: req 0x3210f00
```

### DIFF
--- a/src/coll_score/ucc_coll_score.c
+++ b/src/coll_score/ucc_coll_score.c
@@ -26,6 +26,12 @@ ucc_status_t ucc_coll_score_alloc(ucc_coll_score_t **score)
     return UCC_OK;
 }
 
+static inline void ucc_msg_range_free(ucc_msg_range_t *r)
+{
+    ucc_list_destruct(&r->fallback, ucc_msg_range_fb_t, ucc_free, list_elem);
+    ucc_free(r);
+}
+
 static inline ucc_status_t
 coll_score_add_range(ucc_coll_score_t *score, ucc_coll_type_t coll_type,
                      ucc_memory_type_t mem_type, size_t start, size_t end,
@@ -44,6 +50,9 @@ coll_score_add_range(ucc_coll_score_t *score, ucc_coll_type_t coll_type,
         ucc_error("failed to allocate %zd bytes for ucc_msg_range", sizeof(*r));
         return UCC_ERR_NO_MEMORY;
     }
+
+    ucc_list_head_init(&r->fallback);
+
     r->start   = start;
     r->end     = end;
     r->score   = msg_score;
@@ -65,7 +74,7 @@ coll_score_add_range(ucc_coll_score_t *score, ucc_coll_type_t coll_type,
         (ucc_container_of(next, ucc_msg_range_t, list_elem)->start < r->end)) {
         ucc_error("attempt to add overlaping range");
         ucc_list_del(&r->list_elem);
-        ucc_free(r);
+        ucc_msg_range_free(r);
         return UCC_ERR_INVALID_PARAM;
     }
     return UCC_OK;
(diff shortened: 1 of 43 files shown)


vspetrov:
@bureddy I've implemented the fallback from "post". 
1 commit added is another refactor of ucc_coll_task: we used to cache ucc_coll_args_t on task but now i'm caching ucc_base_coll_args_t (i need base args to potentially relaunch the collective from fallback flow). This change is big in terms of line-count but very simple.

2nd commit actually makes a fallback from post. Note, how it is done: i need to allocate another "fallback task" since user already got origin task pointer from init and we should not alter it. I tested it with NCCL by hacking nccl_allreduce_start (to return ERR_NO_RESOURCE).

src/schedule/ucc_schedule.h
```diff
@@ -68,6 +69,10 @@ typedef struct ucc_coll_task {
     uint8_t n_deps;
     uint8_t n_deps_satisfied;
     uint8_t n_deps_base;
+    struct {
+        ucc_list_link_t      *fb_list;
```
bureddy, line 73: just name "list". its already in fallback struct

src/coll_score/ucc_coll_score.h
```diff
@@ -19,8 +19,16 @@
 
 #define UCC_MSG_MAX UINT64_MAX
 
+typedef struct ucc_msg_range_fb {
+    ucc_list_link_t          list_elem;
+    ucc_score_t              score;
+    ucc_base_coll_init_fn_t  init;
+    ucc_base_team_t         *team;
+} ucc_msg_range_fb_t;
```
bureddy, line 27: rename this struct to generic "ucc_coll_entry" and unify this with "ucc_msg_range_t" struct with common members. this is not really fallback specific. for fallback colls you already have "ucc_list_link_t         fallback" in ucc_msg_range_t

src/components/cl/basic/cl_basic_coll.c
```diff
@@ -11,18 +11,14 @@ ucc_status_t ucc_cl_basic_coll_init(ucc_base_coll_args_t *coll_args,
...
-    ucc_cl_basic_team_t    *cl_team = ucc_derived_of(team, ucc_cl_basic_team_t);
-    ucc_base_coll_init_fn_t init;
-    ucc_base_team_t        *bteam;
-    ucc_status_t            status;
-    status =
-        ucc_coll_score_map_lookup(cl_team->score_map, coll_args, &init, &bteam);
+    ucc_cl_basic_team_t *cl_team = ucc_derived_of(team, ucc_cl_basic_team_t);
+    ucc_status_t         status;
+
+    status = ucc_coll_score_map_init(cl_team->score_map, coll_args, task);
```
bureddy, line 17: given than  we are initializing the coll->init() inside and score_map is an asrugment for initilaization, may be we can rename it to ucc_coll_init()/ WDYT?

src/coll_score/ucc_coll_score.h
```diff
@@ -134,9 +142,11 @@ ucc_status_t ucc_coll_score_build_map(ucc_coll_score_t *score,
...
-                                       ucc_base_coll_args_t    *args,
-                                       ucc_base_coll_init_fn_t *init,
-                                       ucc_base_team_t        **team);
+/* Initializes task based on args selection and score map.
+   Checks fallbacks if necessary. */
+ucc_status_t ucc_coll_score_map_init(ucc_score_map_t      *map,
+                                     ucc_base_coll_args_t *bargs,
+                                     ucc_coll_task_t     **task);
+
+ucc_status_t ucc_coll_task_post_with_fallback(ucc_coll_task_t *task);
```
bureddy, line 151: generic comment: it seems the whole score_map logic is getting complex. should we have a tool option to print all this information for debugging purposes?

src/components/cl/basic/cl_basic_coll.c
vspetrov: sounds good

src/coll_score/ucc_coll_score.h
vspetrov: do you mean printing the score_map data struct, like active protocols selection? I also thought about that. Will try smth.

vspetrov:
@bureddy @manjugv @Sergei-Lebedev I've removed the "fallback from post" for now. Let's see if we can resolve in libsharp. (i've saved the code if at some point we will see that we need that). Plz review.

bureddy:
do we need ucc_base_coll_args_t in coll_task changes given that we removed "fallback from post" ?

src/coll_score/ucc_coll_score.c
```diff
@@ -728,83 +851,117 @@ static ucc_status_t ucc_coll_score_update_one(ucc_list_link_t *dest,
...
-                if (UCC_SCORE_INVALID != rs->score) {
-                    new->score = rs->score;
+                if (UCC_SCORE_INVALID != rs->super.score) {
+                    new->super.score = rs->super.score;
                 }
-                if (rs->init) {
-                    new->init = rs->init;
-                    new->team = rs->team;
+                if (rs->super.init) {
+                    if (rs->super.init != rd->super.init) {
```
bureddy, line 894: seems Line  894-905 is duplicated many times. can we merge ucc_fallback_alloc and ucc_fallback_insert into a single functions?

vspetrov:
> do we need ucc_base_coll_args_t in coll_task changes given that we removed "fallback from post" ?

It is useful in general for several reasons: 1. base_args contains core ucc_team ptr that can be used in multiple places (e.g. if we need to dump log/debug info from request), 2. base_args will contain more info that CL can pass down to TL (in cuda_ipc branch with executor we have use case for that: need to pass executor from CL/HIER to TLs); TLs then usually need to cache it on their tasks struct, it is more convenient to store "base_args" right away.

src/coll_score/ucc_coll_score.c
vspetrov: done

bureddy (approved):

Sergei-Lebedev (approved):

src/coll_score/ucc_coll_score_map.c
```diff
@@ -54,12 +57,43 @@ ucc_status_t ucc_coll_score_map_lookup(ucc_score_map_t         *map,
...
+    ucc_status_t      status;
+
+    status = ucc_coll_score_map_lookup(map, bargs, &r);
+    if (UCC_OK != status) {
+        return status;
+    }
+
+    team   = r->super.team;
+    status = r->super.init(bargs, team, task);
+    fb     = ucc_list_head(&r->fallback, ucc_coll_entry_t, list_elem);
```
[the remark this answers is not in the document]
vspetrov: @Sergei-Lebedev regarding this "likely": again, when user launches, say, sharp + unsupported dt it will always be false hint., so i will add the check w/o "likely"

PR #299, opened 2021-09-13
CORE: Fix ucc_config_names_search

lappazos:
## What
Fix comilation error in ucc_config_names_search

## Why ?
https://github.com/openucx/ucx/pull/7379

## How ?
remove dereference


vspetrov:
@lappazos is UCX PR 7379 applied to ucx/master only? If so, we might need a configure step logic to detect the type of ucs_config_names_search api, because we want to support latest released UCX as well i think. @manjugv @Sergei-Lebedev thoughts?

shimmybalsam (approved):

vspetrov:
closed in favor of #300 (signature detection is added)

PR #300, opened 2021-09-13
UTIL: ucs_config_names_search signature

vspetrov:
## What
Fix compilation error in ucc_config_names_search

## Why ?
openucx/ucx#7379

### DIFF
--- a/config/m4/ucx.m4
+++ b/config/m4/ucx.m4
@@ -92,6 +92,24 @@ AS_IF([test "x$ucx_checked" != "xyes"],[
 
             AC_SUBST(UCX_LIBADD, "-lucp -lucm")
             AC_SUBST(UCS_LIBADD, "-lucs")
+
+            AC_TRY_COMPILE([#include <ucs/config/parser.h>],
+            [
+                ucs_config_names_array_t array;
+                const char *str = "aaa";
+                ucs_config_names_search(array, str);
+            ],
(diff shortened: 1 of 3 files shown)


lappazos (approved):

Sergei-Lebedev (approved):

shimmybalsam (approved):

bureddy (approved):

bureddy:
@vspetrov  this is breaking compatibility. now need to build and run with before/after UCX change

vspetrov:
> @vspetrov  this is breaking compatibility. now need to build and run with before/after UCX change
> 
> 

@bureddy i don't get it. What breaks compat? I thought i made this pr exactly to keep support for both ucx versions. And detection is automatic.

bureddy:
I mean building UCC with UCX version v1 api but, run time using UCX version api v2.

vspetrov:
> I mean building UCC with UCX version v1 api but, run time using UCX version api v2.

I see, you are right. I think we already had such an issue with UCS cache api changes. I think UCX should detect runtime mismatch, no? Otherwise, UCX introduced the same binary incompat, right?

yosefe:
UCS API (unlike UCP/UCT) does not guarantee binary or API compatibility so need to adjust UCC for changes.
In this case, a better fix would be looping over ucs_config_names_array_t directly, instead of using the helper function ucs_config_names_search()

Issue #302, opened 2021-09-14
Semantics and Description of Teams Params


manjugv:
Addressing the changes introduced in #274

PR #303, opened 2021-09-17
API: Add/Clarify Semantics of team params

manjugv:
Addressing Issue #302


src/ucc/api/ucc.h
```diff
@@ -1012,8 +1012,42 @@ typedef struct ucc_team_p2p_conn {
...
+      * When set to this value, the collective participants shall initialize and
+      * invoke the operation in the same order.
+      */
+    UCC_COLLECTIVE_INIT_AND_POST_ORDERED    = 4,
+
+    /**
+      * When set to this value, the collective participants shall initialize and
+      * invoke the operation in any order.
+      */
+    UCC_COLLECTIVE_INIT_AND_POST_UNORDERED  = 5
```
bureddy, line 1050: I think it's confusing.  given that it is enum, what is semantics on post ordering if users set UCC_COLLECTIVE_INIT_ORDERED ?

src/ucc/api/ucc.h
```diff
@@ -1012,8 +1012,42 @@ typedef struct ucc_team_p2p_conn {
...
 } ucc_post_ordering_t;
```
bureddy, line 1051: is there case where INIT is ordered but post is unordered?

bureddy: is it make sense to use bitmask instead of enum here?

src/ucc/api/ucc.h
```diff
@@ -1097,22 +1131,113 @@ typedef struct ucc_ep_map_t {
...
+     * . The oob is used for exchanging information between the team
+     * participants during team creation. The user is
+     * responsible for implementing the oob operation. The relation between @ref
+     * ucc_team_params.ep and @ref ucc_oob_coll.oob_ep is defined as below:
+     *
+     * - When both are not provided. The library is responsible for generating the ep,
+     * which can be then queried via the @ref ucc_team_get_attr interface. This
+     * requires, however,
+     * ucc_params_t ep_map to be set and context created by @ref ucc_oob_coll.
+     * The behavior is
```
bureddy, line 1193: can you fill the warp the text in the line

src/ucc/api/ucc.h
manjugv: > is there case where INIT is ordered but post is unordered?

The MPI persistent collectives in MPI 4.0 require those semantics. When we set to the value "UCC_COLLECTIVE_POST_ORDERED", there is no constraint on the ordering of Init. Should we specify the order for init in this case? I left it undefined intentionally.

manjugv: > is it make sense to use bitmask instead of enum here?

I don't have a preference.

manjugv: > is there case where INIT is ordered but post is unordered?

This semantic is required for MPI 4.0's persistence collectives. When UCC_COLLECTIVE_INIT_ORDERED is set, the post can be ordered or unordered. Is this enough or we should tighten it up?

vspetrov: @manjugv one thing i don't understand: if, say, user sets "UCC_COLLECTIVE_INIT_ORDERED" only. Do this imply POST = unordered (basically post can by "any" but for implementation this means it has to support potentially unordered posts) ?

src/ucc/api/ucc.h
```diff
@@ -1012,8 +1012,42 @@ typedef struct ucc_team_p2p_conn {
  *  @ingroup UCC_TEAM_DT
  */
 typedef enum {
-    UCC_COLLECTIVE_POST_ORDERED     = 0,
-    UCC_COLLECTIVE_POST_UNORDERED   = 1
+
+    /**
+      * When set to this value, the collective participants shall invoke the operation
```
vspetrov, line 1017: invoke->post

src/ucc/api/ucc.h
vspetrov, line 1050: Can we really support that unless it is a "tagged" collective?

src/ucc/api/ucc.h
```diff
@@ -1097,22 +1131,113 @@ typedef struct ucc_ep_map_t {
...
+     * participants during team creation. The user is
+     * responsible for implementing the oob operation. The relation between @ref
+     * ucc_team_params.ep and @ref ucc_oob_coll.oob_ep is defined as below:
+     *
+     * - When both are not provided. The library is responsible for generating the ep,
+     * which can be then queried via the @ref ucc_team_get_attr interface. This
+     * requires, however,
+     * ucc_params_t ep_map to be set and context created by @ref ucc_oob_coll.
+     * The behavior is
+     * undefined, when neither @ref ucc_team_params.ep or @ref
```
vspetrov, line 1194: When nothing is set, should we call it "undef" behavior or just call it error? We can add the check (in the team_create control path) and through the error.

src/ucc/api/ucc.h
```diff
@@ -1097,22 +1131,113 @@ typedef struct ucc_ep_map_t {
...
+     * is undefined.
+     */
     ucc_team_oob_coll_t     oob;
+
+    /** @ref ucc_team_params.p2p_conn is a callback function for the gathering
+      * the point-to-point communication information.
+      */
     ucc_team_p2p_conn_t     p2p_conn;
+
+    /** @ref ucc_team_params.mem_params provides an ability to attach a buffer
```
vspetrov, line 1216: from the description of mem_params here several things are not clear:
1. What type/kind of memory can user pass? Which UCC_MEMORY_TYPEs (gpu/cpu) ? Should it be pinned?
2. What the library can do with that memory? - after team is created (user passed the memory into the ucc), does the ownership goes entirely to libucc? when is it returned to used, upon team destruction?
3. Should/can this memory be used as coll_args dst/src fields?

In general, an example of usage for 1-sided colls would be good.

src/ucc/api/ucc.h
manjugv: If it is error, we need to check and throw an error. If undefined, the onus is on the user.

src/ucc/api/ucc.h
manjugv: Yes. But there are edge cases, where we need tag-based collectives.

src/ucc/api/ucc.h
manjugv: Correct, it can be either ordered or unordered. Ordered Init and unordered post is needed for MPI's persistent collective. The question I have, do we need to be explicit about the post being ordered or unordered?

src/ucc/api/ucc.h
manjugv: @vspetrov We are reworking on how to use this parameter. As a part of #309, we will propose the change. Let's discuss this in today's meeting.

src/ucc/api/ucc.h
manjugv: @vspetrov Did we agree on leaving the semantics as it is or should we add INIT_AND_UNORDERED?

src/ucc/api/ucc.h
manjugv: On this PR, I will not modify the params. The other PR (#310 ) will address any changes in this field, ok?

src/ucc/api/ucc.h
manjugv: UCC WG: Agreed not to add.

src/ucc/api/ucc.h
manjugv: Done

src/ucc/api/ucc.h
manjugv: Done

bureddy (approved):

PR #287, opened 2021-08-11
TL/UCP: Fix bug in ep close

lappazos:
## What
Change ep close proccess and fix small bug in gtest.

## Why ?
https://github.com/openucx/ucx/pull/7140
Causing hang in great without the change

## How ?
Change ep close according to UCX examples, with consideration to special cases where we don't have OOB and need to ignore errors in ep close, which are expected

PR #304, opened 2021-09-21
TL/UCP: fix eps array storage

vspetrov:
## What
Fixes array eps storage in TL/UCP

## Why ?
The flow was not enabled before due to incorrect check during tl ucp context create.

### DIFF
--- a/src/components/tl/ucp/tl_ucp_ep.c
+++ b/src/components/tl/ucp/tl_ucp_ep.c
@@ -51,47 +51,56 @@ ucc_status_t ucc_tl_ucp_connect_team_ep(ucc_tl_ucp_team_t         *team,
     return ucc_tl_ucp_connect_ep(ctx, ep, addr);
 }
 
+/* Finds next non-NULL ep in the storage and returns that handle
+   for closure. In case of "hash" storage it pops the item,
+   in case of "array" sets it to NULL */
+static inline ucp_ep_h get_next_ep_to_close(ucc_tl_ucp_context_t *ctx, int *i)
+{
+    ucp_ep_h   ep = NULL;
+    ucc_rank_t size;
+
+    if (ctx->eps) {
+        size = (ucc_rank_t)ctx->super.super.ucc_context->params.oob.n_oob_eps;
+        while (NULL == ep && (*i) < size) {
+            ep           = ctx->eps[*i];
+            ctx->eps[*i] = NULL;
+            (*i)++;
+        }
+    } else {
+        ep = tl_ucp_hash_pop(ctx->ep_hash);
+    }
+    return ep;
+}
+
 void ucc_tl_ucp_close_eps(ucc_tl_ucp_context_t *ctx)
 {
+     int                          i = 0;
      ucp_ep_h                     ep;
      ucs_status_t                 status;
-     void **                      close_reqs;
-     void *                       close_req;
-     int                          i, close_reqs_counter = 0;
+     ucs_status_ptr_t             close_req;
+     ucp_request_param_t          param;
 
-     close_reqs = (void **)ucc_malloc(sizeof(void *) * kh_size(ctx->ep_hash),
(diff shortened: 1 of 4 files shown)


src/components/tl/ucp/tl_ucp_ep.h
```diff
@@ -44,7 +44,12 @@ static inline ucc_status_t ucc_tl_ucp_get_ep(ucc_tl_ucp_team_t *team, ucc_rank_t
     ucc_status_t               status;
 
     if (ctx->eps) {
-        ctx_rank = ucc_get_ctx_rank(team->super.super.team, rank);
+        /* Core super.super.team ptr is NULL for service_team
+           which has scope == UCC_CL_LAST + 1*/
+        ucc_assert((NULL != team->super.super.team) ||
```
bureddy, line 49: define a macro` IS_SERVICE_TEAM(team)`

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -95,3 +95,64 @@ void ucc_tl_ucp_close_eps(ucc_tl_ucp_context_t *ctx)
...
 }
+
+static void ucc_tl_ucp_close_eps_array(ucc_tl_ucp_context_t *ctx)
+{
+     ucp_ep_h                     ep;
+     ucs_status_t                 status;
+     void **                      close_reqs;
+     void *                       close_req;
+     int                          i, close_reqs_counter = 0;
+     int                          size;
```
bureddy, line 106: extra spaces between type and names?

manjugv (approved):

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -95,3 +95,64 @@ void ucc_tl_ucp_close_eps(ucc_tl_ucp_context_t *ctx)
...
+             // to sync closure, and we can ignore errors, which are expected
+         }
+     }
+     for (i = 0; i < close_reqs_counter; i++) {
+         close_req = close_reqs[i];
+         // TODO: Should we put a timer? in UCX, some examples have timer, and some don't
+         do {
+             ucp_worker_progress(ctx->ucp_worker);
+             status = ucp_request_check_status(close_req);
+         } while (status == UCS_INPROGRESS && close_req);
```
Sergei-Lebedev, line 141: how close_req can be null here?

Sergei-Lebedev (approved):

vspetrov:
@Sergei-Lebedev @bureddy plz re-review. The close_ep is re-implemented: old logic added by @lappazos in #287 (O(N) close_reqs allocation) is not needed. we can process eps 1 by 1. in this PR i unified ep_close for 2 cases: with hash and array storage. it is much simpler now.

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -51,47 +51,52 @@ ucc_status_t ucc_tl_ucp_connect_team_ep(ucc_tl_ucp_team_t         *team,
     return ucc_tl_ucp_connect_ep(ctx, ep, addr);
 }
 
+static inline ucp_ep_h get_next_ep(ucc_tl_ucp_context_t *ctx, int *i)
```
Sergei-Lebedev, line 54: I'm not sure get_next_ep should be a function, I think better move code inside close_eps because 
1. Where else we can use that function?
2. It does different things for different types of storage, for array it just returns next ep but for hash it additionally removes ep from storage

Sergei-Lebedev (approved):

src/components/tl/ucp/tl_ucp_ep.c
vspetrov: it is used twice in the ucc_tl_ucp_close_eps fn: i didn't want to duplicate it. It does different things indeed and only used for eps close flow. Will it be better if i change "static inline" to helper macro?

PR #307, opened 2021-09-24
API: coll timeout option

vspetrov:
## What
Adds the TIMEOUT option to the ucc_coll_args

### DIFF
--- a/src/core/ucc_progress_queue_mt.c
+++ b/src/core/ucc_progress_queue_mt.c
@@ -7,9 +7,11 @@
 #include "ucc_progress_queue.h"
 #include "utils/ucc_malloc.h"
 #include "utils/ucc_log.h"
+#include "utils/ucc_time.h"
 #include "utils/ucc_spinlock.h"
 #include "utils/ucc_list.h"
 #include "utils/ucc_lock_free_queue.h"
+#include "utils/ucc_coll_utils.h"
 
 typedef struct ucc_pq_mt {
     ucc_progress_queue_t super;
@@ -35,6 +37,7 @@ static void ucc_pq_locked_mt_enqueue(ucc_progress_queue_t *pq,
 static void ucc_pq_mt_enqueue(ucc_progress_queue_t *pq, ucc_coll_task_t *task)
 {
     ucc_pq_mt_t *pq_mt      = ucc_derived_of(pq, ucc_pq_mt_t);
+
     ucc_lf_queue_enqueue(&pq_mt->lf_queue, &task->lf_elem);
 }
 
@@ -63,15 +66,29 @@ static void ucc_pq_mt_dequeue(ucc_progress_queue_t *pq,
 
 static int ucc_pq_mt_progress(ucc_progress_queue_t *pq)
 {
-    int              n_progressed = 0;
+    int              n_progressed =  0;
+    double           timestamp    = -1;
     ucc_coll_task_t *task;
     ucc_status_t     status;
+
     pq->dequeue(pq, &task);
     if (task) {
         if (task->progress) {
             task->progress(task);
         }
         if (UCC_INPROGRESS == task->super.status) {
+            if (UCC_COLL_TIMEOUT_REQUIRED(task->args)) {
+                if (timestamp < 0) {
+                    timestamp = ucc_get_time();
+                }
+                if (ucc_unlikely(timestamp - task->start_time >
+                                 task->args.timeout)) {
+                    task->super.status = UCC_ERR_TIMED_OUT;
+                    ucc_task_complete(task);
+                    return UCC_ERR_TIMED_OUT;
+                }
+            }
+
             pq->enqueue(pq, task);
             return n_progressed;
         }
--- a/src/schedule/ucc_schedule.h
+++ b/src/schedule/ucc_schedule.h
@@ -9,6 +9,7 @@
 #include "utils/ucc_list.h"
 #include "utils/ucc_log.h"
 #include "utils/ucc_lock_free_queue.h"
+#include "utils/ucc_coll_utils.h"
 
 #define MAX_LISTENERS 4
 
@@ -68,6 +69,8 @@ typedef struct ucc_coll_task {
     uint8_t n_deps;
     uint8_t n_deps_satisfied;
     uint8_t n_deps_base;
+    double  start_time; /* timestamp of the start time:
+                           either post or triggered_post */
 } ucc_coll_task_t;
 
 typedef struct ucc_context ucc_context_t;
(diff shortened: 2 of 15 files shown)


src/ucc/api/ucc_status.h
```diff
@@ -39,6 +39,7 @@ typedef enum {
     UCC_ERR_NO_RESOURCE                 =   -5,
     UCC_ERR_NO_MESSAGE                  =   -6, /*!< General purpose return code without specific error */
     UCC_ERR_NOT_FOUND                   =   -7,
+    UCC_ERR_TIMED_OUT                   =   -8,
```
manjugv, line 42: Should we make this an error or status? It can still run to completion correctly, right?

src/ucc/api/ucc.h
```diff
@@ -1391,12 +1391,22 @@ typedef enum {
...
+    UCC_COLL_ARGS_FLAG_TIMEOUT              = UCC_BIT(6)  /*!<If set, the collective
+                                                            elapsed time is being measured
+                                                            and compared vs the provided
+                                                            timeout value. UCC_ERR_TIMED_OUT
+                                                            will be returned from collective
+                                                            test if elapsed time exceeds
+                                                            the timeout.
+                                                            Note, the status is not guaranteed
+                                                            to be global on all the processes
+                                                            participating in the collective*/
```
manjugv, line 1409: I'm guessing you are saying that the timeout applies only to the calling thread.

vspetrov: yes, "reported" (via collective_test) timeout is valid only the calling thread. How should i better formulate that?

src/ucc/api/ucc_status.h
vspetrov: do you mean making it >0 ? maybe. @Sergei-Lebedev @bureddy  wdyt?

src/core/ucc_progress_queue_mt.c
```diff
@@ -27,6 +29,10 @@ static void ucc_pq_locked_mt_enqueue(ucc_progress_queue_t *pq,
 {
     ucc_pq_mt_locked_t *pq_mt = ucc_derived_of(pq, ucc_pq_mt_locked_t);
 
+    if (UCC_HAS_TIMEOUT(task->args)) {
```
bureddy, line 32: UCC_HAS_TIMEOUT/UCC_COLL_TIMEOUT_REQUIRED

src/core/ucc_progress_queue_mt.c
```diff
@@ -27,6 +29,10 @@ static void ucc_pq_locked_mt_enqueue(ucc_progress_queue_t *pq,
...
+        task->started = ucc_get_accurate_time();
```
bureddy, line 33: s/ucc_get_accurate_time/ucc_get_time ?

src/ucc/api/ucc_status.h
bureddy: it should be error right? user does not care and check if coll is completed or not after timeout

src/utils/ucc_time.h
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
```
bureddy, line 4: why can't we use ucs/time/time.h ?

src/core/ucc_progress_queue_mt.c
bureddy: why can't we set the start time coll_post instead if conditions in the progress func?

src/utils/ucc_time.h
vspetrov: it is not exposed as public header. only time_def.h is. time.h contains arch specific implementation.

src/utils/ucc_time.h
```diff
@@ -0,0 +1,23 @@
...
+#define UCC_TIME_H_
+#include "config.h"
+#include <sys/time.h>
+
+#define UCC_USEC_PER_SEC   1000000ul     /* Micro */
+
+/**
+ * @return The current accurate time, in seconds.
+ */
+static inline double ucc_get_accurate_time()
```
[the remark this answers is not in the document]
vspetrov: i thought about that: we can, if we copy-paste implementations from UCS (ucs_arch_read_hres_clock). Should we?

src/ucc/api/ucc_status.h
manjugv: Typically with error, if we detect library doesn't continue the execution. In this case, we are continuing the execution.

src/ucc/api/ucc.h
manjugv: @vspetrov I'm not sure, if this better :) `"If set and the elapsed time after call invocation is greater than @ref ucc_coll_args_t.timeout, the library returns UCC_ERR_TIMED_OUT on the calling thread. There is no impact on threads or processes that does not set the value. "`

vspetrov:
@bureddy moved timer setup into coll_post/trigged_post

src/core/ucc_progress_queue_mt.c
```diff
@@ -63,15 +66,28 @@ static void ucc_pq_mt_dequeue(ucc_progress_queue_t *pq,
...
     ucc_coll_task_t *task;
     ucc_status_t     status;
+
     pq->dequeue(pq, &task);
     if (task) {
         if (task->progress) {
             task->progress(task);
         }
         if (UCC_INPROGRESS == task->super.status) {
+            if (UCC_COLL_TIMEOUT_REQUIRED(task->args)) {
```
bureddy, line 80: ucc_unlikely() ?

src/schedule/ucc_schedule.h
```diff
@@ -68,6 +69,7 @@ typedef struct ucc_coll_task {
     uint8_t n_deps;
     uint8_t n_deps_satisfied;
     uint8_t n_deps_base;
+    double  started; /* timstamp of the start time */
```
bureddy, line 72: s/started/start_time ?

src/utils/ucc_coll_utils.h
```diff
@@ -29,6 +30,17 @@
...
 
+#define UCC_COLL_TIMEOUT_REQUIRED(_args) \
+    (((_args).mask & UCC_COLL_ARGS_FIELD_FLAGS) && \
+     ((_args).flags & UCC_COLL_ARGS_FLAG_TIMEOUT))
+
+#define UCC_COLL_SET_TIMEOUT(_task, _timeout) do {           \
+        (_task)->args.mask   |= UCC_COLL_ARGS_FIELD_FLAGS;   \
+        (_task)->args.flags  |= UCC_COLL_ARGS_FLAG_TIMEOUT;  \
+        (_task)->args.timeout = _timeout;                    \
+        (_task)->started = ucc_get_time();                   \
```
bureddy, line 41: align =

src/schedule/ucc_schedule.h
Sergei-Lebedev, line 72: s/timstamp/timestamp

Sergei-Lebedev (approved):

src/core/ucc_progress_queue_mt.c
vspetrov: i don't think this is right hint. E.g., for the application which actually uses TM (FB) it will always produce wrong hint.

bureddy (approved):

PR #311, opened 2021-09-30
CORE: coll seq_num

vspetrov:
## What
Adds running sequence number per ucc team.

## Why ?
Current use case: this significantly simplifies debug of the hanging apps (because it is easy to find late process from the log).

### DIFF
--- a/src/core/ucc_coll.c
+++ b/src/core/ucc_coll.c
@@ -199,10 +199,12 @@ UCC_CORE_PROFILE_FUNC(ucc_status_t, ucc_collective_init,
         task->cb = coll_args->cb;
         task->flags |= UCC_COLL_TASK_FLAG_CB;
     }
+    task->seq_num = team->seq_num++;
     if (ucc_global_config.log_component.log_level >= UCC_LOG_LEVEL_DEBUG) {
         char coll_debug_str[256];
         ucc_coll_str(&op_args, coll_debug_str, sizeof(coll_debug_str));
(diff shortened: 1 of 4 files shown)


Sergei-Lebedev (approved):

src/core/ucc_coll.c
```diff
@@ -193,10 +193,12 @@ UCC_CORE_PROFILE_FUNC(ucc_status_t, ucc_collective_init,
         task->cb = coll_args->cb;
         task->flags |= UCC_COLL_TASK_FLAG_CB;
     }
+    task->seq_num = team->seq_num++;
```
bureddy, line 196: do you need atomic inc here?

vspetrov: not needed i think since multiple threads are not allowed to call collective_init concurrently on the same team.

bureddy: why did we do in HCOLL? Ex: https://github.com/Mellanox/hcoll/blob/master/src/hcoll/ml/coll_ml_bcast.c#L1356

vspetrov: it came from coll/ml. i think it was not necessary.

bureddy (approved):

src/core/ucc_coll.c
manjugv: I'm not sure, if we had well-defined thread model there, particular from ML POV.

manjugv (approved):

PR #317, opened 2021-10-07
CORE: fix timeout for schedule

Sergei-Lebedev:
## What
Fixes timeout error for collectives that are represented by schedule or pipelined schedule

### DIFF
--- a/src/schedule/ucc_schedule_pipelined.c
+++ b/src/schedule/ucc_schedule_pipelined.c
@@ -14,6 +14,7 @@ static ucc_status_t ucc_frag_start_handler(ucc_coll_task_t *parent,
     ucc_schedule_t *frag               = ucc_derived_of(task, ucc_schedule_t);
     ucc_status_t    status;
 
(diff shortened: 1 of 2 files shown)


src/schedule/ucc_schedule_pipelined.c
```diff
@@ -14,6 +14,7 @@ static ucc_status_t ucc_frag_start_handler(ucc_coll_task_t *parent,
     ucc_schedule_t *frag               = ucc_derived_of(task, ucc_schedule_t);
     ucc_status_t    status;
 
+    task->start_time = parent->start_time;
```
vspetrov, line 17: is it enough or we need to sert mask/flag for timeout on children task? (UCC_COLL_SET_TIMEOUT macro?)
Do you need to check if parent has timeout set?

vspetrov (commented):
do we need to extend gtest/test_timeout.cc with , e.g., sra allreduce to check that (even though those timeout tests are disabled by default) ?

PR #318, opened 2021-10-08
CORE: move triggered post to common

vspetrov:
## What
Moves implementation of ucc_collective_triggered_post from tl/ucp to core

## Why ?
Implementation is used in other components as well: e.g., cl/hier


src/core/ucc_coll.c
```diff
@@ -223,23 +224,167 @@ ucc_status_t ucc_collective_post(ucc_coll_req_h request)
...
+        status = ucc_mc_ee_task_post(task->ee->ee_context,
+                                     task->ee->ee_type, &task->ee_task);
+        if (ucc_unlikely(status != UCC_OK)) {
+            ucc_error("error in ee task post, %s", ucc_status_string(status));
+            task->super.status = status;
+            return status;
+        }
+
+        /* TODO: mpool */
+        post_event = ucc_malloc(sizeof(ucc_ev_t), "event");
```
Sergei-Lebedev, line 312: when post_event will be freed?

bureddy: the consumer (ucc_ee_get_event) should call ucc_ee_ack_event

bureddy (approved):

src/core/ucc_coll.c
vspetrov: yes, i did that

PR #323, opened 2021-10-18
TL/UCP: One-sided A2A

wfaderhold21:
## What
Adding a one-sided alltoall implementation for the UCP TL. The algorithm implemented is a naive, linear algorithm with synchronization on exit.  

## Why ?
This will add a basic, one-sided alltoall algorithm, which does not exist.

### DIFF
--- a/src/components/tl/ucp/alltoall/alltoall.c
+++ b/src/components/tl/ucp/alltoall/alltoall.c
@@ -11,6 +11,21 @@
 ucc_status_t ucc_tl_ucp_alltoall_pairwise_start(ucc_coll_task_t *task);
 ucc_status_t ucc_tl_ucp_alltoall_pairwise_progress(ucc_coll_task_t *task);
 
+ucc_status_t ucc_tl_ucp_alltoall_onesided_start(ucc_coll_task_t *task);
+ucc_status_t ucc_tl_ucp_alltoall_onesided_progress(ucc_coll_task_t *task);
+
+ucc_base_coll_alg_info_t
+    ucc_tl_ucp_alltoall_algs[UCC_TL_UCP_ALLTOALL_ALG_LAST + 1] = {
+        [UCC_TL_UCP_ALLTOALL_ALG_PAIRWISE] =
+            {.id   = UCC_TL_UCP_ALLTOALL_ALG_PAIRWISE,
+             .name = "pairwise",
+             .desc = "pairwise two-sided implementation"},
+        [UCC_TL_UCP_ALLTOALL_ALG_ONESIDED] =
+            {.id   = UCC_TL_UCP_ALLTOALL_ALG_ONESIDED,
+             .name = "onesided",
+             .desc = "naive, linear one-sided implementation"},
+        [UCC_TL_UCP_ALLTOALL_ALG_LAST] = {.id = 0, .name = NULL, .desc = NULL}};
+
 ucc_status_t ucc_tl_ucp_alltoall_init(ucc_tl_ucp_task_t *task)
 {
     ucc_status_t status;
@@ -36,3 +51,40 @@ ucc_status_t ucc_tl_ucp_alltoall_pairwise_init(ucc_base_coll_args_t *coll_args,
 out:
     return status;
 }
+
+ucc_status_t ucc_tl_ucp_alltoall_onesided_init(ucc_base_coll_args_t *coll_args,
+                                               ucc_base_team_t *     team,
+                                               ucc_coll_task_t **    task_h)
+{
+    ucc_tl_ucp_team_t *tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
+    ucc_tl_ucp_task_t *task;
+    ucc_status_t       status;
+
+    ALLTOALL_TASK_CHECK(coll_args->args, tl_team);
+
+    if (!(coll_args->args.mask & UCC_COLL_ARGS_FIELD_GLOBAL_WORK_BUFFER)) {
+        tl_debug(UCC_TL_TEAM_LIB(tl_team),
+                 "global work buffer not provided nor associated with team");
+        status = UCC_ERR_NOT_SUPPORTED;
+        goto out;
+    }
+
+    if (coll_args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) {
+        if (!(coll_args->args.flags & UCC_COLL_ARGS_FLAG_MEM_MAPPED_BUFFERS)) {
+            tl_debug(UCC_TL_TEAM_LIB(tl_team),
+                     "non memory mapped buffers are not supported");
+            status = UCC_ERR_NOT_SUPPORTED;
+            goto out;
+        }
+    }
+
+    task    = ucc_tl_ucp_init_task(coll_args, team);
+    *task_h = &task->super;
+
+    task->super.post     = ucc_tl_ucp_alltoall_onesided_start;
+    task->super.progress = ucc_tl_ucp_alltoall_onesided_progress;
+    status               = UCC_OK;
+out:
+    return status;
+}
+
--- a/src/components/tl/ucp/alltoall/alltoall.c
+++ b/src/components/tl/ucp/alltoall/alltoall.c
@@ -63,14 +63,14 @@ ucc_status_t ucc_tl_ucp_alltoall_onesided_init(ucc_base_coll_args_t *coll_args,
     ALLTOALL_TASK_CHECK(coll_args->args, tl_team);
 
     if (!(coll_args->args.mask & UCC_COLL_ARGS_FIELD_GLOBAL_WORK_BUFFER)) {
-        tl_debug(UCC_TL_TEAM_LIB(tl_team),
+        tl_error(UCC_TL_TEAM_LIB(tl_team),
                  "global work buffer not provided nor associated with team");
         status = UCC_ERR_NOT_SUPPORTED;
         goto out;
     }
     if (coll_args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) {
         if (!(coll_args->args.flags & UCC_COLL_ARGS_FLAG_MEM_MAPPED_BUFFERS)) {
-            tl_debug(UCC_TL_TEAM_LIB(tl_team),
+            tl_error(UCC_TL_TEAM_LIB(tl_team),
                      "non memory mapped buffers are not supported");
             status = UCC_ERR_NOT_SUPPORTED;
             goto out;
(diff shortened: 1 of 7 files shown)


src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,78 @@
...
+    ucc_tl_ucp_put_nb((void *)(src + start * nelems), (void *)dest, nelems,
+                      start, team, task);
+    ucc_tl_ucp_atomic_inc(pSync, start, team, task);
+    ucc_tl_ucp_ep_flush(start, team, task);
+    
+    for (peer = (start + 1) % npes; peer != start; peer = (peer + 1) % npes) {
+        ucc_tl_ucp_put_nb((void *)(src + peer * nelems), (void *)dest, nelems,
+                          peer, team, task);
+        ucc_tl_ucp_atomic_inc(pSync, peer, team, task);
+        ucc_tl_ucp_ep_flush(peer, team, task);
```
manjugv, line 63: Do you really need a flush after every put?

wfaderhold21: @manjugv the flush can be there or after the loop, but there should be a flush to guarantee completion of operation...

wfaderhold21:
@vspetrov @manjugv should we update the score calculations in this PR or a separate PR? I recall mention of needing to add another dimension (i.e., onesided) because this algorithm could be called when one-sided collectives had not been setup on a context, correct?

vspetrov:
> @vspetrov @manjugv should we update the score calculations in this PR or a separate PR? I recall mention of needing to add another dimension (i.e., onesided) because this algorithm could be called when one-sided collectives had not been setup on a context, correct?

yes i think we can do that in separate PR because it will require some thinking.

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,81 @@
...
+ucc_status_t ucc_tl_ucp_alltoall_onesided_progress(ucc_coll_task_t *ctask)
+{
+    ucc_tl_ucp_task_t *task   = ucc_derived_of(ctask, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t *team   = TASK_TEAM(task);
+    ptrdiff_t          src    = (ptrdiff_t)TASK_ARGS(task).src.info.buffer;
+    ptrdiff_t          dest   = (ptrdiff_t)TASK_ARGS(task).dst.info.buffer;
+    ucc_rank_t         mype   = UCC_TL_TEAM_RANK(team);
+    ucc_rank_t         npes   = UCC_TL_TEAM_SIZE(team);
+    size_t             nelems = TASK_ARGS(task).src.info.count;
+    int *              phase  = &task->barrier.phase;
```
vspetrov, line 45: barrier?

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,81 @@
...
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_ucp_alltoall_onesided_progress(ucc_coll_task_t *ctask)
+{
+    ucc_tl_ucp_task_t *task   = ucc_derived_of(ctask, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t *team   = TASK_TEAM(task);
+    ptrdiff_t          src    = (ptrdiff_t)TASK_ARGS(task).src.info.buffer;
+    ptrdiff_t          dest   = (ptrdiff_t)TASK_ARGS(task).dst.info.buffer;
+    ucc_rank_t         mype   = UCC_TL_TEAM_RANK(team);
```
vspetrov, line 42: minor: lets use "rank" instead of "pe" so that internally all UCC collectives looked similar. i know PE is shmem term but here lets keep it uniform.

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,81 @@
...
+    /* TODO: change when support for library-based work buffers is complete */
+    pSync = TASK_ARGS(task).global_work_buffer;
+
+    if (*phase == 1) {
+        goto wait;
+    }
+
+    nelems = (nelems / npes) * ucc_dt_size(TASK_ARGS(task).src.info.datatype);
+    dest   = dest + mype * nelems;
+    ucc_rank_t start = (mype + 1) % npes;
```
vspetrov, line 58: lets move local var declaration to the beginning

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,81 @@
...
+    *phase = 1;
+
+wait:
+    if (*pSync < npes - 1) {
+        return UCC_INPROGRESS;
+    }
+
+    *pSync                   = -1;
+    task->super.super.status = UCC_OK;
+    ucc_task_complete(ctask);
```
vspetrov, line 79: @manjugv @wfaderhold21 i've got a question regarding completion semantics here. Looking at the code: we only check that incoming puts are completed (we only check our local psync counter) before reporting a task status = UCC_OK. This means that outstanding puts might still not be completed. And this in-turn means that user (shmem) can't still modify send buffer. But we already report a completion. So, how does shmem know when it is safe to re-use src buffer?

wfaderhold21: @vspetrov You are correct. Looking back over this, it does not meet the specification requirements for shmem for completion, which include the ability to reuse the src buffer and dest buffer should be completely updated. I've updated the code (also code in PR309) and believe this would be correct now.

shimmybalsam (approved):

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,79 @@
...
+    dest   = dest + grank * nelems;
+    ucc_tl_ucp_put_nb((void *)(src + start * nelems), (void *)dest, nelems,
+                      start, team, task);
+    ucc_tl_ucp_atomic_inc(pSync, start, team, task);
+
+    for (peer = (start + 1) % gsize; peer != start; peer = (peer + 1) % gsize) {
+        ucc_tl_ucp_put_nb((void *)(src + peer * nelems), (void *)dest, nelems,
+                          peer, team, task);
+        ucc_tl_ucp_atomic_inc(pSync, peer, team, task);
+    }
```
manjugv, line 68: @wfaderhold21 You are making an assumption of ordered delivery here.  Ideally, you want to add a fence.

manjugv (approved):

manjugv:
@wfaderhold21 Have you tested with multiple outstanding collective operations? You might see better throughput or bandwidth.

src/components/tl/ucp/alltoall/alltoall.c
```diff
@@ -36,3 +51,37 @@ ucc_status_t ucc_tl_ucp_alltoall_pairwise_init(ucc_base_coll_args_t *coll_args,
...
+                                               ucc_coll_task_t **    task_h)
+{
+    ucc_tl_ucp_team_t *tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
+    ucc_tl_ucp_task_t *task;
+    ucc_status_t       status;
+
+    ALLTOALL_TASK_CHECK(coll_args->args, tl_team);
+
+    if (!(coll_args->args.mask & UCC_COLL_ARGS_FIELD_GLOBAL_WORK_BUFFER)) {
+        tl_debug(UCC_TL_TEAM_LIB(tl_team),
```
Sergei-Lebedev, line 66: it should be tl_error

src/components/tl/ucp/alltoall/alltoall.c
```diff
@@ -36,3 +51,37 @@ ucc_status_t ucc_tl_ucp_alltoall_pairwise_init(ucc_base_coll_args_t *coll_args,
...
+                 "global work buffer not provided nor associated with team");
+        status = UCC_ERR_NOT_SUPPORTED;
+        goto out;
+    }
+    if (coll_args->args.mask & UCC_COLL_ARGS_FIELD_FLAGS) {
+        if (!(coll_args->args.flags & UCC_COLL_ARGS_FLAG_MEM_MAPPED_BUFFERS)) {
+            tl_debug(UCC_TL_TEAM_LIB(tl_team),
```
Sergei-Lebedev, line 73: tl_error

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,79 @@
...
+        if (*pSync < gsize - 1 || task->send_completed < task->send_posted) {
+            ucp_worker_progress(UCC_TL_UCP_TEAM_CTX(team)->ucp_worker);
+            return UCC_INPROGRESS;
+        }
+        goto completed;
+    }
+
+    nelems = (nelems / gsize) * ucc_dt_size(TASK_ARGS(task).src.info.datatype);
+    dest   = dest + grank * nelems;
+    ucc_tl_ucp_put_nb((void *)(src + start * nelems), (void *)dest, nelems,
```
Sergei-Lebedev, line 60: pls add check for return value of ucc_tl_ucp_put_nb and ucc_tl_ucp_atomic_inc

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -0,0 +1,79 @@
...
+                      start, team, task);
+    ucc_tl_ucp_atomic_inc(pSync, start, team, task);
+
+    for (peer = (start + 1) % gsize; peer != start; peer = (peer + 1) % gsize) {
```
Sergei-Lebedev, line 64: i think it's better to move this loop and ucc_tl_ucp_put_nb above to ucc_tl_ucp_alltoall_onesided_start because anyway that code will be executed only once at collective start. So we can simplify progress function and reduce number of additional checks in it

vspetrov:
@manjugv @wfaderhold21 per WG discussion last week: i recall we decided that this PR will also add some correctness tests, right?

manjugv:
> @manjugv @wfaderhold21 per WG discussion last week: i recall we decided that this PR will also add some correctness tests, right?

IIRC, we decided to bring it as a separate PR. @wfaderhold21  What do you prefer?

wfaderhold21:
@manjugv my notes agree that the correctness tests were to be a separate PR. I prefer that but if anyone is against it, I can add here..

vspetrov:
@wfaderhold21 if adding corr tests separately is more convenient, we can do that as long as the new code path is disabled by default. plz rebase and fix CI/Linter failures if any.

vspetrov:
@wfaderhold21 plz rebase. is everything else addressed?

wfaderhold21:
@vspetrov I believe so. new path should be disabled by default and all comments were addressed

Sergei-Lebedev:
ok to test

vspetrov:
bot:retest

vspetrov (approved):

PR #336, opened 2021-11-02
TL/CUDA: implement tl cuda

Sergei-Lebedev:
## What
Implementation of new TL based on CUDA IPC transport a.k.a NVLink. This PR adds general logic to create lib, context and team objects as well as implementation of alltoall collective.
Further steps:
- support for triggered post
- support for CL HIER
- support for different NVLink topologies
- more collectives

## Why ?
In some cases we cannot use CUDA IPC through UCP TL
Performance of Alltoall on 1 Node, 4 PPN
| Size    | TL UCP (us) | TL CUDA (us) |
|---------|-------------|--------------|
| 4096    | 103.58      | 55.33        |
| 8192    | 174.51      | 55.97        |
| 16384   | 90.74       | 59.73        |
| 32768   | 91.56       | 61.60        |
| 65536   | 94.36       | 67.02        |
| 131072  | 100.62      | 72.01        |
| 262144  | 112.01      | 89.31        |
| 524288  | 144.72      | 113.65       |
| 1048576 | 233.94      | 176.28       |
| 2097152 | 384.55      | 300.24       |
| 4194304 | 716.55      | 524.44       |

### DIFF
--- a/src/components/tl/cuda/Makefile.am
+++ b/src/components/tl/cuda/Makefile.am
@@ -16,6 +16,7 @@ sources =             \
 	tl_cuda_team.c    \
 	tl_cuda_coll.c    \
 	tl_cuda_cache.c   \
+	tl_cuda_topo.c    \
 	$(alltoall)
 
 module_LTLIBRARIES = libucc_tl_cuda.la
--- a/src/components/tl/cuda/Makefile.am
+++ b/src/components/tl/cuda/Makefile.am
@@ -1,0 +1,30 @@
+#
+# Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+#
+
+if TL_CUDA_ENABLED
+alltoall =                       \
+	alltoall/alltoall.h          \
+	alltoall/alltoall.c          \
+	alltoall/alltoall_ce.c
+
+sources =             \
+	tl_cuda.h         \
+	tl_cuda.c         \
+	tl_cuda_lib.c     \
+	tl_cuda_context.c \
+	tl_cuda_team.c    \
+	tl_cuda_coll.c    \
+	tl_cuda_cache.c   \
+	$(alltoall)
+
+module_LTLIBRARIES = libucc_tl_cuda.la
+libucc_tl_cuda_la_SOURCES  = $(sources)
+libucc_tl_cuda_la_CPPFLAGS = $(AM_CPPFLAGS) $(BASE_CPPFLAGS) $(CUDA_CPPFLAGS)
+libucc_tl_cuda_la_CFLAGS   = $(BASE_CFLAGS)
+libucc_tl_cuda_la_LDFLAGS  = -version-info $(SOVERSION) --as-needed $(CUDA_LDFLAGS)
+libucc_tl_cuda_la_LIBADD   = $(CUDA_LIBS) $(UCC_TOP_BUILDDIR)/src/libucc.la
+
+include $(top_srcdir)/config/module.am
+
+endif
--- a/src/components/tl/cuda/alltoall/alltoall_ce.c
+++ b/src/components/tl/cuda/alltoall/alltoall_ce.c
@@ -7,6 +7,7 @@
 #include "alltoall.h"
 #include "core/ucc_mc.h"
 #include "tl_cuda_cache.h"
+#include "tl_cuda_topo.h"
 
 enum {
     ALLTOALL_CE_STAGE_SYNC, /*< Wait for free SYNC segment */
@@ -34,11 +35,11 @@ ucc_status_t ucc_tl_cuda_alltoall_setup(ucc_tl_cuda_task_t *task)
     ucc_status_t                 status;
     ucc_rank_t                   i;
 
-    sync->ptr = task->alltoall_ce.mem_info.ptr;
-    sync->length = task->alltoall_ce.mem_info.length;
-    sync->offset = task->alltoall_ce.mem_info.offset;
-    memcpy(&sync->handle, &task->alltoall_ce.mem_info.handle,
-           sizeof(cudaIpcMemHandle_t));
+    memcpy(&sync->mem_info_src, &task->alltoall_ce.mem_info_src,
+           sizeof(ucc_tl_cuda_mem_info_t));
+    memcpy(&sync->mem_info_dst, &task->alltoall_ce.mem_info_dst,
+           sizeof(ucc_tl_cuda_mem_info_t));
+
     CUDACHECK_GOTO(cudaEventRecord(sync->ipc_event_local, team->stream),
                    exit_err, status, UCC_TL_TEAM_LIB(team));
     status = ucc_tl_cuda_shm_barrier_start(UCC_TL_TEAM_RANK(team), task->bar);
@@ -52,7 +53,8 @@ ucc_status_t ucc_tl_cuda_alltoall_setup(ucc_tl_cuda_task_t *task)
     }
 
     for (i = 0; i < UCC_TL_TEAM_SIZE(team); i++) {
-        if (i == UCC_TL_TEAM_RANK(team)) {
+        if (i == UCC_TL_TEAM_RANK(team) ||
+            !ucc_tl_cuda_topo_is_direct(team, team->topo, UCC_TL_TEAM_RANK(team), i)) {
             continue;
         }
         peer_sync = TASK_SYNC(task, i);
@@ -61,9 +63,19 @@ ucc_status_t ucc_tl_cuda_alltoall_setup(ucc_tl_cuda_task_t *task)
             status = UCC_ERR_NO_MESSAGE;
             goto exit_err;
         }
-        status = ucc_tl_cuda_map_memhandle(peer_sync->ptr, peer_sync->length,
-                                           peer_sync->handle,
-                                           &task->alltoall_ce.peer_map_addr[i],
+        status = ucc_tl_cuda_map_memhandle(peer_sync->mem_info_src.ptr,
+                                           peer_sync->mem_info_src.length,
+                                           peer_sync->mem_info_src.handle,
+                                           &task->alltoall_ce.peer_map_addr_src[i],
+                                           cache);
+        if (UCC_OK != status) {
+            ucc_error("ucc_cuda_ipc_map_memhandle failed");
+            return UCC_ERR_INVALID_PARAM;
+        }
+        status = ucc_tl_cuda_map_memhandle(peer_sync->mem_info_dst.ptr,
+                                           peer_sync->mem_info_dst.length,
+                                           peer_sync->mem_info_dst.handle,
+                                           &task->alltoall_ce.peer_map_addr_dst[i],
                                            cache);
         if (UCC_OK != status) {
             ucc_error("ucc_cuda_ipc_map_memhandle failed");
@@ -80,35 +92,57 @@ static ucc_status_t ucc_tl_cuda_alltoall_ce_post_copies(ucc_tl_cuda_task_t *task
 {
     ucc_tl_cuda_team_t *team = TASK_TEAM(task);
     ucc_coll_args_t    *args = &TASK_ARGS(task);
-    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, UCC_TL_TEAM_RANK(team));
+    ucc_rank_t          rank = UCC_TL_TEAM_RANK(team);
+    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, rank);
     ucc_tl_cuda_sync_t *peer_sync;
     size_t send_len;
-    ucc_rank_t i, peer;
+    ucc_rank_t i, peer, psrc, pdst;
     void *src, *dst;
     ucc_status_t status;
 
     send_len = (args->src.info.count / UCC_TL_TEAM_SIZE(team)) *
                ucc_dt_size(args->src.info.datatype);
     for (i = 0; i < UCC_TL_TEAM_SIZE(team); i++) {
-        peer = (UCC_TL_TEAM_RANK(team) + i) % UCC_TL_TEAM_SIZE(team);
+        peer = (rank + i) % UCC_TL_TEAM_SIZE(team);
+        if (!ucc_tl_cuda_topo_is_direct(team, team->topo, rank, peer)) {
+            continue;
+        }
         peer_sync = TASK_SYNC(task, peer);
-        if (peer == UCC_TL_TEAM_RANK(team)) {
+        if (peer == rank) {
             src = args->src.info.buffer;
         } else {
-            src = PTR_OFFSET(task->alltoall_ce.peer_map_addr[peer],
-                             peer_sync->offset);
+            src = PTR_OFFSET(task->alltoall_ce.peer_map_addr_src[peer],
+                             peer_sync->mem_info_src.offset);
             CUDACHECK_GOTO(cudaStreamWaitEvent(team->stream,
                                                sync->data[peer].ipc_event_remote,
                                                0),
                            exit, status, UCC_TASK_LIB(task));
         }
-        src = PTR_OFFSET(src, UCC_TL_TEAM_RANK(team) * send_len);
+        src = PTR_OFFSET(src, rank * send_len);
         dst = PTR_OFFSET(args->dst.info.buffer, peer * send_len);
         CUDACHECK_GOTO(cudaMemcpyAsync(dst, src, send_len,
                                        cudaMemcpyDeviceToDevice, team->stream),
                        exit, status, UCC_TASK_LIB(task));
     }
 
+    for (i = 0; i < team->topo->num_proxies; i++) {
+        if (team->topo->proxies[i].proxy == rank) {
+            psrc = team->topo->proxies[i].src;
+            pdst = team->topo->proxies[i].dst;
+            peer_sync = TASK_SYNC(task, psrc);
+            src = PTR_OFFSET(task->alltoall_ce.peer_map_addr_src[psrc],
+                             peer_sync->mem_info_src.offset);
+            src = PTR_OFFSET(src, pdst * send_len);
+            peer_sync = TASK_SYNC(task, pdst);
+            dst = PTR_OFFSET(task->alltoall_ce.peer_map_addr_dst[pdst],
+                             peer_sync->mem_info_dst.offset);
+            dst = PTR_OFFSET(dst, psrc * send_len);
+            CUDACHECK_GOTO(cudaMemcpyAsync(dst, src, send_len,
+                                           cudaMemcpyDeviceToDevice,
+                                           team->stream),
+                           exit, status, UCC_TASK_LIB(task));
+        }
+    }
     status = ucc_mc_ee_event_post(team->stream, task->alltoall_ce.copy_done,
                                   UCC_EE_CUDA_STREAM);
 exit:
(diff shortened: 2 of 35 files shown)


src/components/tl/cuda/Makefile.am
```diff
@@ -0,0 +1,27 @@
+#
+# Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+#
+
+alltoall =                       \
+	alltoall/alltoall.h          \
+	alltoall/alltoall.c          \
+	alltoall/alltoall_ce.c
```
vspetrov, line 8: what does "_ce" suffix stand for?

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+}
+
+ucc_status_t ucc_tl_cuda_alltoall_ce_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_cuda_task_t *task   = ucc_derived_of(coll_task, ucc_tl_cuda_task_t);
+    ucc_tl_cuda_team_t *team   = TASK_TEAM(task);
+    ucc_tl_cuda_sync_t *sync   = TASK_SYNC(task, team->rank);
+    ucc_rank_t          n_done = 0;
+    ucc_tl_cuda_sync_t *peer_sync;
+    ucc_status_t status;
```
vspetrov, line 29: local vars align

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+    ucc_rank_t i;
+
+    status = ucc_mc_ee_event_test(task->alltoall_ce.copy_done,
+                                  UCC_EE_CUDA_STREAM);
+    if (status != UCC_OK) {
+        task->super.super.status = status;
+        goto exit;
+    }
+    sync->seq_num[1] = task->seq_num;
+    __sync_synchronize();
```
vspetrov, line 39: both lines below (_sync_synchronize and asm) were POC hacks. What we need is ARCH independent "write memory barrier". I think we can probably start with __sync_synchronize(). 2nd (_asm_) is compile level barrier. i don't even think we need it - i added it when we were seeing races long ago and it just left there. @bureddy what do you think, is __sync_synchronize (hw mem barrier) is enough?

Another think, both are GCC extensions - need to think if that's a problem.

Again, the right solution imho is to have utils/arch/x86.h (ppc.h/arm.h) and there to define ucc_wmb() or ucc_isync() (similar to ucs/arch). Probably this can be done in other PR.

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+    return task->super.super.status;
+}
+
+ucc_status_t ucc_tl_cuda_alltoall_setup(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
+    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, team->rank);
+    volatile ucc_tl_cuda_sync_t *peer_sync;
+    ucc_tl_cuda_cache_t *cache;
```
vspetrov, line 61: align vars

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+exit:
+    return task->super.super.status;
+}
+
+ucc_status_t ucc_tl_cuda_alltoall_setup(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
+    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, team->rank);
+    volatile ucc_tl_cuda_sync_t *peer_sync;
```
vspetrov, line 60: why need to have "volatile" ? isn't it not enough to have volatile "seq_num" inside the sync struct?

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+    ucc_tl_cuda_cache_t *cache;
+    ucc_mem_attr_t mem_attr;
+    ucc_status_t status;
+    size_t data_len;
+    ucc_rank_t i;
+
+    data_len = ucc_dt_size(args->src.info.datatype) * args->src.info.count;
```
vspetrov, line 67: maybe all these lines 67-81 can go to some helper fn ucc_tl_cuda_map_local_buffer.
I recall the same code will be used for alltoallv

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+
+}
+
+ucc_status_t ucc_tl_cuda_alltoall_ce_start(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_cuda_task_t *task = ucc_derived_of(coll_task, ucc_tl_cuda_task_t);
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_status_t status;
+
+    status = ucc_tl_cuda_alltoall_setup(task);
```
vspetrov, line 162: do we need to split alltoall_setup into 2 functions?
1st: local memory_setup - can be done during tl_cuda_alltoall_init
2nd: writing input dependency event + shm exchange - happens either in start or in early_triggered_post (using old terminology)

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,181 @@
...
+    }
+exit:
+    return task->super.super.status;
+}
+
+ucc_status_t ucc_tl_cuda_alltoall_setup(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
+    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, team->rank);
```
vspetrov, line 59: after you getting you "sync" based on seq_num you have to make sure that old/previous collective (you have 8/max_concurrent on the fly) has completed. Otherwise you risk overwriting the shm and corrupting data.

src/components/tl/cuda/tl_cuda.c
```diff
@@ -0,0 +1,90 @@
...
+                                   ucc_base_team_t *team,
+                                   ucc_coll_task_t **task);
+
+ucc_status_t ucc_tl_cuda_team_get_scores(ucc_base_team_t   *tl_team,
+                                         ucc_coll_score_t **score_p);
+
+UCC_TL_IFACE_DECLARE(cuda, CUDA);
+
+
+// static inline ucc_context_addr_header_t *
```
vspetrov, line 63: remove

src/components/tl/cuda/tl_cuda.h
```diff
@@ -0,0 +1,150 @@
...
+    int device;
+    int shm;
+} ucc_tl_cuda_rank_id_t;
+
+typedef struct ucc_tl_cuda_sync_data {
+    cudaEvent_t ipc_event_remote;
+} ucc_tl_cuda_sync_data_t;
+
+typedef struct ucc_tl_cuda_sync {
+    uint32_t                 seq_num[16];
```
vspetrov, line 111: looks like you are using only 2 seq_nums ? why need 16?

src/components/tl/cuda/tl_cuda.h
```diff
@@ -0,0 +1,150 @@
...
+UCC_CLASS_DECLARE(ucc_tl_cuda_team_t, ucc_base_context_t *,
+                  const ucc_base_team_params_t *);
+
+typedef struct ucc_tl_cuda_task {
+    ucc_coll_task_t super;
+    uint32_t        seq_num;
+    uint32_t        coll_id;
+    union {
+        struct {
+            void *peer_map_addr[8];
```
vspetrov, line 144: at least have a define MAX_PEERS 8.
soon we will need to have it dynamic imho.

src/components/tl/cuda/tl_cuda_cache.c
```diff
@@ -0,0 +1,326 @@
...
+{
+    ucc_tl_cuda_cache_purge(cache);
+    ucs_pgtable_cleanup(&cache->pgtable);
+    pthread_rwlock_destroy(&cache->lock);
+    free(cache->name);
+    free(cache);
+}
+
+#if ENABLE_CACHE
+static void ucc_tl_cuda_cache_invalidate_regions(ucc_tl_cuda_cache_t *cache,
```
vspetrov, line 118: @Sergei-Lebedev @bureddy can this cache stuff be abstracted as generic peer_based cache? I see a use case for KNEM regions cache in TL/SHM ?

src/components/tl/cuda/tl_cuda_coll.h
```diff
@@ -0,0 +1,67 @@
...
+
+#define TASK_SYNC(_task, _rank)                                                \
+    ({                                                                         \
+        ucc_tl_cuda_team_t *_team = TASK_TEAM(_task);                          \
+        UCC_TL_CUDA_TEAM_SYNC(_team, _rank, (_task)->coll_id);                 \
+    })
+
+static inline void ucc_tl_cuda_task_reset(ucc_tl_cuda_task_t *task)
+{
+    task->super.super.status = UCC_INPROGRESS;
```
vspetrov, line 25: "reset" is tricky imho. Originally "reset" was added to support persistent behavior or frags restart in fragmented schedules (with pipeline depth < num_of frags). Since task has a fixed seq_num during its lifetime, after "reset" you will have all remote shm sync values equal to that seq_num from the previous execution. THis will imply sync will be incorrect.

I guess we should do task->seq_num++ each reset, while task->coll_id will be fixed to use same sync each time. wdyt?

src/components/tl/cuda/Makefile.am
Sergei-Lebedev: ce is copy engine

src/components/tl/cuda/alltoall/alltoall_ce.c
Sergei-Lebedev: I don't think so, from cppreference "any attempt to read or write to an object whose type is volatile-qualified through a non-volatile lvalue results in undefined behavior"

src/components/tl/cuda/alltoall/alltoall_ce.c
Sergei-Lebedev: yes, it makes sense, but I'm not sure how the interface for such ucc_tl_cuda_map_local_buffer should look like. Let's do it for next PR when we will have another collective?

src/components/tl/cuda/alltoall/alltoall_ce.c
Sergei-Lebedev: done

src/components/tl/cuda/alltoall/alltoall_ce.c
Sergei-Lebedev: fixed

src/components/tl/cuda/tl_cuda_coll.h
Sergei-Lebedev: fixed

src/components/tl/cuda/alltoall/alltoall_ce.c
Sergei-Lebedev: fixed

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,250 @@
...
+        status = ucc_tl_cuda_map_memhandle(peer_sync->mem_info_src.ptr,
+                                           peer_sync->mem_info_src.length,
+                                           peer_sync->mem_info_src.handle,
+                                           &task->alltoall_ce.peer_map_addr_src[i],
+                                           cache);
+        if (UCC_OK != status) {
+            ucc_error("ucc_cuda_ipc_map_memhandle failed");
+            return UCC_ERR_INVALID_PARAM;
+        }
+        status = ucc_tl_cuda_map_memhandle(peer_sync->mem_info_dst.ptr,
```
vspetrov, line 77: so now you have to map both src and dst? is this needed only for proxy or always? does this imply overheads when "proxy" is not required ?

src/components/tl/cuda/tl_cuda_coll.h
```diff
@@ -0,0 +1,119 @@
...
+    task->super.team         = &team->super.super;
+    ucc_tl_cuda_task_reset(task);
+    return task;
+}
+
+static inline void ucc_tl_cuda_task_put(ucc_tl_cuda_task_t *task)
+{
+    // ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    // ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, UCC_TL_TEAM_RANK(team));
+    // uint32_t max_concurrent;
```
vspetrov, line 52: remove comments

src/components/tl/cuda/tl_cuda_team.c
```diff
@@ -0,0 +1,303 @@
...
+    if (UCC_TL_TEAM_RANK(self) == 0) {
+        shm_id = shmget(IPC_PRIVATE, ctrl_size, IPC_CREAT | 0666);
+        if (shm_id < 0) {
+            tl_error(tl_context->lib, "failed to shmget with IPC_PRIVATE, "
+                     "size %zd, IPC_CREAT errno: %d(%s)", ctrl_size,
+                     errno, strerror(errno));
+            status = UCC_ERR_NO_MEMORY;
+            goto exit_err;
+        }
+        self->sync = shmat(shm_id, NULL, 0);
```
vspetrov, line 60: if shmget /shmat fails then rank 0 will return with error and never go to oob -> dead lock on other ranks? lets use negative shm_id and make rank 0 always go to allgather to process global error state.

vspetrov:
@bureddy plz have a look at this PR. From my side the PR looks ready. Let's have another pair of eyes on it. We need this in order to proceed with 2lvl AR and A2A algs on GPU.

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,250 @@
...
+
+}
+
+ucc_status_t ucc_tl_cuda_alltoall_ce_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_cuda_task_t *task = ucc_derived_of(coll_task, ucc_tl_cuda_task_t);
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_status_t        status;
+
+    if (task->alltoall_ce.stage == ALLTOALL_CE_STAGE_SYNC) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: done

src/components/tl/cuda/alltoall/alltoall_ce.c
```diff
@@ -0,0 +1,250 @@
...
+           sizeof(ucc_tl_cuda_mem_info_t));
+
+    CUDACHECK_GOTO(cudaEventRecord(sync->ipc_event_local, team->stream),
+                   exit_err, status, UCC_TL_TEAM_LIB(team));
+    ucc_memory_bus_fence();
+    status = ucc_tl_cuda_shm_barrier_start(UCC_TL_TEAM_RANK(team), task->bar);
+    if (ucc_unlikely(status != UCC_OK)) {
+        goto exit_err;
+    }
+    while (ucc_tl_cuda_shm_barrier_test(UCC_TL_TEAM_RANK(team), task->bar) ==
```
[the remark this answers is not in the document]
Sergei-Lebedev: fixed, alltoall setup is now nonblocking

src/components/tl/cuda/alltoall/alltoall_ce.c
Sergei-Lebedev: right, dst buffer registration is needed only if we use proxy rank however with registration cache overhead is not big. Let's postpone this change for another PR, there are some other optimization I need to add

src/components/tl/cuda/tl_cuda_team.c
Sergei-Lebedev: done

Sergei-Lebedev:
@vspetrov all comments fixed

vspetrov (approved):

src/components/tl/cuda/tl_cuda_coll.c
```diff
@@ -0,0 +1,76 @@
...
+ucc_status_t ucc_tl_cuda_shm_barrier_init(ucc_rank_t size, ucc_rank_t rank,
+                                          ucc_tl_cuda_shm_barrier_t *barrier)
+{
+    if (rank == 0) {
+        barrier->size = size;
+        barrier->count = 0;
+        barrier->sense = 0;
+    }
+    barrier->state[rank]       = UCC_OK;
+    barrier->local_sense[rank] = 1;
```
bureddy, line 49: why do you need this array?

src/components/tl/cuda/tl_cuda.h
```diff
@@ -0,0 +1,189 @@
...
+    cudaIpcMemHandle_t  handle;
+} ucc_tl_cuda_mem_info_t;
+
+typedef struct ucc_tl_cuda_sync {
+    uint32_t                 seq_num[2];
+    ucc_tl_cuda_mem_info_t   mem_info_src;
+    ucc_tl_cuda_mem_info_t   mem_info_dst;
+    cudaEvent_t              ipc_event_local;
+    cudaIpcEventHandle_t     ev_handle;
+    ucc_tl_cuda_sync_data_t  data[1];
```
bureddy, line 141: why array of size "1"?

src/components/tl/cuda/tl_cuda.h
```diff
@@ -0,0 +1,189 @@
...
+#include "tl_cuda_ep_hash.h"
+#include <cuda_runtime.h>
+
+#ifndef UCC_TL_CUDA_DEFAULT_SCORE
+#define UCC_TL_CUDA_DEFAULT_SCORE 40
+#endif
+
+#define UCC_TL_CUDA_MAX_PEERS 8
+#define UCC_TL_CUDA_SUPPORTED_COLLS                                            \
+    (UCC_COLL_TYPE_ALLTOALL)
```
bureddy, line 22: how about alltoallv?

src/components/tl/cuda/tl_cuda_team.c
```diff
@@ -0,0 +1,316 @@
...
+            /* status is set to UCC_OK to let rank 0 proceed and notify
+              other ranks about error */
+            status = UCC_OK;
+            goto exit_err;
+        }
+        self->sync = shmat(shm_id, NULL, 0);
+        self->bar  = (ucc_tl_cuda_shm_barrier_t*)UCC_TL_CUDA_TEAM_SYNC(self, 0,
+                                                       lib->cfg.max_concurrent);
+        shmctl(shm_id, IPC_RMID, NULL);
+        if (self->sync == (void *)-1) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: added ucc_sysv_alloc and ucc_sysv_free functions to utils, it now handles error checking similar to ucs

src/components/tl/cuda/tl_cuda_team.c
```diff
@@ -0,0 +1,316 @@
...
+    if (self->ids) {
+        if (self->sync != (void*)-1) {
+            for (i = 0; i < lib->cfg.max_concurrent; i++) {
+                for (j = 0; j < UCC_TL_TEAM_SIZE(self); j++) {
+                    if (j == UCC_TL_TEAM_RANK(self)) {
+                        continue;
+                    }
+                    sync = UCC_TL_CUDA_TEAM_SYNC(self, j, i);
+                    if (sync->data[j].ipc_event_remote) {
+                        cudaEventDestroy(sync->data[j].ipc_event_remote);
```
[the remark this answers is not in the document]
Sergei-Lebedev: done

src/components/tl/cuda/tl_cuda_coll.c
Sergei-Lebedev: it's used to handle the case of multiple barriers, official name of the algorithm is barrier with sense reversal

src/components/tl/cuda/tl_cuda.h
Sergei-Lebedev: it's actually of size TEAM_SIZE, but it's defined as an array of 1 element for convenience

src/components/tl/cuda/tl_cuda.h
Sergei-Lebedev: alltoallv is relatively easy to add once alltoall is merged, I just don't want to make this PR bigger

bureddy (approved):

PR #337, opened 2021-11-03
BUILD: fix clang tidy warnings

Sergei-Lebedev:
## What
Fixing following clang-tidy warnings

-  allreduce/allreduce_rab.c: 100:5: warning: 3rd function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage]
- scatter/scatter_knomial.c:61:28:  warning: Value stored to 'sbuf' during its initialization is never read [clang-analyzer-deadcode.DeadStores]
- core/ucc_coll.c:314:56: warning: parameter 'ev' is unused [misc-unused-parameters]
- coll_score/ucc_coll_score.c:123:9: warning: Potential memory leak [clang-analyzer-unix.Malloc]

### DIFF
--- a/src/coll_score/ucc_coll_score.c
+++ b/src/coll_score/ucc_coll_score.c
@@ -119,8 +119,9 @@ static ucc_status_t ucc_fallback_alloc(ucc_score_t              score,
 
     fb = ucc_malloc(sizeof(*fb), "fallback");
     if (ucc_unlikely(!fb)) {
+        ucc_error("failed to allocate %zd bytes for fallback",
+                  sizeof(ucc_coll_entry_t));
         *_fb = NULL;
-        ucc_error("failed to allocate %zd bytes for fallback", sizeof(*fb));
         return UCC_ERR_NO_MEMORY;
     }
     fb->score = score;
@@ -140,6 +141,7 @@ static inline void ucc_fallback_insert(ucc_list_link_t  *list,
     ucc_list_for_each(f, list, list_elem) {
         if (fb->score == f->score && fb->init == f->init &&
             fb->team == f->team) {
+            ucc_free(fb);
             /* same fallback: skip */
             return;
         }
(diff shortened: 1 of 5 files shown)


src/components/cl/hier/allreduce/allreduce_rab.c
```diff
@@ -54,6 +54,11 @@ UCC_CL_HIER_PROFILE_FUNC(ucc_status_t, ucc_cl_hier_allreduce_rab_init,
         goto out;
     }
 
+    if (!SBGP_ENABLED(cl_team, NODE) && !SBGP_ENABLED(cl_team, NODE_LEADERS)) {
```
vspetrov, line 57: this cannot happen - it is rather ucc_assert. Either of those SBGP must be enabled.

Sergei-Lebedev: I know, it's just to suppress warning. ucc_assert will not work because it's no op for release build

vspetrov: can we do smth with NOLINT tricks ?

Sergei-Lebedev: updated, is it better now?

src/coll_score/ucc_coll_score.c
```diff
@@ -131,7 +132,8 @@ static ucc_status_t ucc_fallback_alloc(ucc_score_t              score,
 }
 
 static inline void ucc_fallback_insert(ucc_list_link_t  *list,
-                                       ucc_coll_entry_t *fb)
+                                       ucc_coll_entry_t *fb,
+                                       int free_fb)
```
bureddy, line 136: seems always called with free_fb = 1. is it needed?

vspetrov (approved):

src/coll_score/ucc_coll_score.c
Sergei-Lebedev: not really needed, maybe only to explicitly say that this function will remove object
@vspetrov wdyt?

vspetrov: @Sergei-Lebedev sounds good to me

Sergei-Lebedev: @bureddy fixed

bureddy (approved):

PR #345, opened 2021-11-10
TOOLS: add reduce_scatter to perftest

Sergei-Lebedev:
## What
add reduce_scatter to perftest

### DIFF
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -1194,10 +1194,10 @@ enum ucc_team_attr_field {
  */
 enum ucc_team_flags {
     UCC_TEAM_FLAG_COLL_WORK_BUFFER             = UCC_BIT(0) /*< If set, this indicates
-                                                                the user will provide
-                                                                a scratchpad buffer for
-                                                                use in one-sided
-                                                                collectives. Otherwise,
+                                                                the user will provide 
+                                                                a scratchpad buffer for 
+                                                                use in one-sided 
+                                                                collectives. Otherwise, 
                                                                 an internal buffer will
                                                                 used. */
 };
@@ -1696,13 +1696,7 @@ typedef enum {
     UCC_COLL_ARGS_FLAG_IN_PLACE             = UCC_BIT(0), /*!< If set, the
                                                             output buffer is
                                                             identical to the
-                                                            input buffer.
-                                                            Note, for inplace
-                                                            reduce scatter
-                                                            data is placed in
-                                                            output buffer with
-                                                            offset according
-                                                            to team id. */
+                                                            input buffer.*/
     UCC_COLL_ARGS_FLAG_PERSISTENT           = UCC_BIT(1), /*!< If set, the
                                                             collective is
                                                             considered
@@ -1743,9 +1737,9 @@ typedef enum {
                                                             to be global on all the processes
                                                             participating in the collective.*/
     UCC_COLL_ARGS_FLAG_MEM_MAPPED_BUFFERS   = UCC_BIT(7)  /*!< If set, both src
-                                                            and dst buffers
-                                                            reside in a memory
-                                                            mapped region.
+                                                            and dst buffers 
+                                                            reside in a memory 
+                                                            mapped region. 
                                                             Useful for one-sided
                                                             collectives. */
 } ucc_coll_args_flags_t;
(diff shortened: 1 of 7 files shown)


tools/perf/ucc_pt_coll_reduce_scatter.cc
```diff
@@ -0,0 +1,92 @@
...
+                        ucc_pt_comm *communicator) : ucc_pt_coll(communicator)
+{
+    has_inplace_   = true;
+    has_reduction_ = true;
+    has_range_     = true;
+    has_bw_        = true;
+
+    coll_args.coll_type = UCC_COLL_TYPE_REDUCE_SCATTER;
+    coll_args.mask = 0;
+    if (is_inplace) {
```
vspetrov, line 19: 1 question: last time we discussed INPLACE for RS in WG we agreed that there can be 2 modes (result in the beginning of the buffer or with the offset). Which inplace is it? What is out default mode in UCC?

vspetrov:
Lets document RS INPLACE in ucc.h. Default UCC behavior is MPI-like: result of operation is stored in the Beginning of recv buffer.

Sergei-Lebedev:
> Lets document RS INPLACE in ucc.h. Default UCC behavior is MPI-like: result of operation is stored in the Beginning of recv buffer.

added

Sergei-Lebedev:
also fixed allgather bandwidth calculation in perftest

tools/perf/ucc_pt_coll_reduce_scatter.cc
```diff
@@ -0,0 +1,92 @@
...
+    coll_args.dst.info.mem_type = mt;
+}
+
+ucc_status_t ucc_pt_coll_reduce_scatter::init_coll_args(size_t count,
+                                                        ucc_coll_args_t &args)
+{
+    size_t size;
+    ucc_status_t st;
+
+    if (count % comm->get_size() != 0) {
```
vspetrov, line 38: @Sergei-Lebedev i was running this PR with tl/ucp reduce-scatter and faced a problem that the test is effectively disabled for many cases. E.g., running on 1 jazz node with ppn=28. It would be skipped due to this condition.

I suggest we change it this way:
if (count < comm->get_size()) {
    skip;
}
count = count - (count % comm->get_size());

Sergei-Lebedev: @vspetrov this expected, size should be multiple of number of ranks. Maybe for reduce_scatter and scatter make count to be number of elements to receive?

vspetrov: sure, that is essentially the same. i'm just saying lets not disable test for odd rank counts. we can make count = count * comm->get_size()

manjugv (approved):

tools/perf/ucc_pt_coll_reduce_scatter.cc
Sergei-Lebedev: fixed

src/ucc/api/ucc.h
```diff
@@ -1696,7 +1696,13 @@ typedef enum {
     UCC_COLL_ARGS_FLAG_IN_PLACE             = UCC_BIT(0), /*!< If set, the
                                                             output buffer is
                                                             identical to the
-                                                            input buffer.*/
+                                                            input buffer.
+                                                            Note, for inplace
+                                                            reduce scatter
```
vspetrov, line 1701: I think we need to document that somehow differently. Firstly, why "offset according to team id" ? did you mean to say "according to rank", but we don't really have "rank" defined ... Secondly, need to clearly defined what is the "offset" = rank * dst.info.count/team_size * dt_size. Thirdly, same applies to reduce-scatterv but offset is dst.info_v.displacements[rank] * dt_size. Maybe we should add all this to manual but not put here in ucc.h (looks too much) ?? @manjugv

vspetrov (approved):

src/ucc/api/ucc.h
Sergei-Lebedev: > I think we need to document that somehow differently. Firstly, why "offset according to team id" ? did you mean to say "according to rank", but we don't really have "rank" defined ... 

right, we don't have rank defined, that's why "according to team id"

vspetrov: "team id" is not defined either. i suggest we move detailed description to manual.

vspetrov: Or lets write:
default INPLACE behavior for UCC_COLL_TYPE_REDUCE_SCATTER and UCC_COLL_TYPE_REDUCE_SCATTERV is to place result into destination buffer at the offset rank * dst.info.count/team_size * dt_size or dst.info_v.displacements[rank] * dt_size respectively (where "rank" is the integer id of a process defined during team creation).

@manjugv plz help, i know you wanted to avoid using "rank" term but now it is very difficult to even write specification of the collective behavior without that term. For example we have info_v.counts/displacements arrays that are indexed using ranks of processes (not enpoint values), but we never defined rank itself.

manjugv: @Sergei-Lebedev @vspetrov ep=rank, when ep is ordered and contiguous. 

For reduce_scatter, we can write like this. If we agree on this, we can write to the rest of the collectives.

**The UCC_COLL_TYPE_REDUCE_SCATTER first performs a data element-wise reduction on the src buffer and then scatters the result to the dst buffer. If the size of src buffer is “count * dt_size”, where dt_size is the number of bytes for the data type element and count is the number of elements of that datatype. Assuming the result can be divided into “n” blocks, the ith block is placed in the receive buffer of endpoint “I”. Like other collectives, for this collective, the “ep” should be ordered and contiguous. 

When INPLACE is set for UCC_COLL_TYPE_REDUCE_SCATTER, the receive buffers act as both send and receive buffer.**

vspetrov: Sounds good to me

Sergei-Lebedev:
After WG discussion we decided to remove any docs changes from this PR. @manjugv will push another PR with proper collectives description

PR #348, opened 2021-11-13
API: new datatype definition

vspetrov:
## What
Changes the UCC.h datatype and reduce_op definitions to support custom dtypes

## Why ?
Current API does not allow support for derived types.

## How ?
dtype and op are 8bytes (uint64_t) values.
Predefined dtypes are still the same from the user perspective (UCC_DT_INT32, etc).
ucc_dt_make_contig, ucc_dt_make_generic

custom op: ucc_op_make_userdefined

Required OMPI change: vspetrov/ompi@9746d0c - MINIMAL

Example of Working code that does UCC_allreduce over custom dtype (struct of fields int+float) and custom reduce op:
https://gist.github.com/vspetrov/1a4cb03b1018ed742e5762768da925db

### DIFF
--- a/src/components/tl/ucp/allreduce/allreduce.h
+++ b/src/components/tl/ucp/allreduce/allreduce.h
@@ -21,16 +21,6 @@ ucc_status_t ucc_tl_ucp_allreduce_init(ucc_tl_ucp_task_t *task);
 #define UCC_TL_UCP_ALLREDUCE_DEFAULT_ALG_SELECT_STR                            \
     "allreduce:0-4k:@0#allreduce:4k-inf:@1"
 
-#define CHECK_USERDEFINED_OP(_args, _team)                                     \
-    do {                                                                       \
-        if (_args.mask & UCC_COLL_ARGS_FIELD_USERDEFINED_REDUCTIONS) {         \
-            tl_error(UCC_TL_TEAM_LIB(_team),                                   \
-                     "userdefined reductions are not supported yet");          \
-            status = UCC_ERR_NOT_SUPPORTED;                                    \
-            goto out;                                                          \
-        }                                                                      \
-    } while (0)
-
 #define CHECK_SAME_MEMTYPE(_args, _team)                                       \
     do {                                                                       \
         if (!UCC_IS_INPLACE(_args) &&                                          \
@@ -43,7 +33,6 @@ ucc_status_t ucc_tl_ucp_allreduce_init(ucc_tl_ucp_task_t *task);
     } while (0)
 
 #define ALLREDUCE_TASK_CHECK(_args, _team)                                     \
-    CHECK_USERDEFINED_OP((_args), (_team));                                    \
     CHECK_SAME_MEMTYPE((_args), (_team));
 
 ucc_status_t ucc_tl_ucp_allreduce_knomial_init(ucc_base_coll_args_t *coll_args,
--- a/src/core/ucc_mc.h
+++ b/src/core/ucc_mc.h
@@ -110,13 +110,17 @@ static inline ucc_status_t ucc_mc_reduce_userdefined(void *src1, void *src2,
                                                      size_t count, size_t stride,
                                                      ucc_dt_generic_t *dt)
 {
-    int i;
+    ucc_reduce_cb_params_t params = {
+        .src1      = src1,
+        .src2      = src2,
+        .dst       = dst,
+        .n_vectors = n_vectors,
+        .count     = count,
+        .stride    = stride,
+        .dt        = dt
+    };
 
-    dt->ops.reduce.cb(src1, src2, dst, count, dt->ops.reduce.ctx);
-    for (i = 1; i < n_vectors; i++) {
-        dt->ops.reduce.cb(PTR_OFFSET(src2, stride*i), dst, dst, count,
-                           dt->ops.reduce.ctx);
-    }
+    dt->ops.reduce.cb(&params);
     return UCC_OK;
 }
 
@@ -159,7 +163,7 @@ ucc_dt_reduce_multi_alpha(void *src1, void *src2, void *dst, size_t n_vectors,
 {
     /* reduce_multi is used for OP_AVG implementation that can only be
        used with predefined dtypes */
-    ucc_assert(!UCC_DT_IS_PREDEFINED(dt));
+    ucc_assert(UCC_DT_IS_PREDEFINED(dt));
     return ucc_mc_reduce_multi_alpha(src1, src2, dst, n_vectors, count,
                                      stride, dt, args->op,
                                      vector_op, alpha, mem_type);
--- a/src/core/ucc_mc.h
+++ b/src/core/ucc_mc.h
@@ -8,6 +8,8 @@
 
 #include "ucc/api/ucc.h"
 #include "components/mc/base/ucc_mc_base.h"
+#include "core/ucc_dt.h"
+#include "utils/ucc_math.h"
 
 ucc_status_t ucc_mc_init(const ucc_mc_params_t *mc_params);
 
@@ -103,17 +105,34 @@ ucc_status_t ucc_mc_reduce_multi_alpha(
     size_t stride, ucc_datatype_t dtype, ucc_reduction_op_t reduce_op,
     ucc_reduction_op_t vector_op, double alpha, ucc_memory_type_t mem_type);
 
-static inline ucc_status_t ucc_dt_reduce(const void *src1, const void *src2,
+static inline ucc_status_t ucc_mc_reduce_userdefined(void *src1, void *src2,
+                                                     void *dst, size_t n_vectors,
+                                                     size_t count, size_t stride,
+                                                     ucc_dt_generic_t *dt)
+{
+    int i;
+
+    dt->ops.reduce.cb(src1, src2, dst, count, dt->ops.reduce.ctx);
+    for (i = 1; i < n_vectors; i++) {
+        dt->ops.reduce.cb(PTR_OFFSET(src2, stride*i), dst, dst, count,
+                           dt->ops.reduce.ctx);
+    }
+    return UCC_OK;
+}
+
+static inline ucc_status_t ucc_dt_reduce(void *src1, void *src2,
                                          void *dst, size_t count,
                                          ucc_datatype_t dt,
                                          ucc_memory_type_t mem_type,
                                          ucc_coll_args_t *args)
 {
-    if (args->mask & UCC_COLL_ARGS_FIELD_USERDEFINED_REDUCTIONS) {
-        return UCC_ERR_NOT_SUPPORTED; //TODO
+    if (!UCC_DT_IS_PREDEFINED(dt)) {
+        ucc_assert(UCC_DT_HAS_REDUCE(dt));
+        return ucc_mc_reduce_userdefined(src1, src2, dst, 1, count,
+                                         0, ucc_dt_to_generic(dt));
     } else {
-        return ucc_mc_reduce(src1, src2, dst, count, dt,
-                             args->reduce.predefined_op, mem_type);
+        return ucc_mc_reduce(src1, src2, dst, count,
+                             dt, args->op, mem_type);
     }
 }
 
@@ -122,11 +141,13 @@ ucc_dt_reduce_multi(void *src1, void *src2, void *dst, size_t n_vectors,
                     size_t count, size_t stride, ucc_datatype_t dt,
                     ucc_memory_type_t mem_type, ucc_coll_args_t *args)
 {
-    if (args->mask & UCC_COLL_ARGS_FIELD_USERDEFINED_REDUCTIONS) {
-        return UCC_ERR_NOT_SUPPORTED; //TODO
+    if (!UCC_DT_IS_PREDEFINED(dt)) {
+        ucc_assert(UCC_DT_HAS_REDUCE(dt));
+        return ucc_mc_reduce_userdefined(src1, src2, dst, n_vectors, count,
+                                         stride, ucc_dt_to_generic(dt));
     } else {
         return ucc_mc_reduce_multi(src1, src2, dst, n_vectors, count, stride,
-                                   dt, args->reduce.predefined_op, mem_type);
+                                   dt, args->op, mem_type);
     }
 }
 
@@ -136,11 +157,11 @@ ucc_dt_reduce_multi_alpha(void *src1, void *src2, void *dst, size_t n_vectors,
                           ucc_reduction_op_t vector_op, double alpha,
                           ucc_memory_type_t mem_type, ucc_coll_args_t *args)
 {
-    if (args->mask & UCC_COLL_ARGS_FIELD_USERDEFINED_REDUCTIONS) {
-        return UCC_ERR_NOT_SUPPORTED; //TODO
-    }
+    /* reduce_multi is used for OP_AVG implementation that can only be
+       used with predefined dtypes */
+    ucc_assert(!UCC_DT_IS_PREDEFINED(dt));
     return ucc_mc_reduce_multi_alpha(src1, src2, dst, n_vectors, count,
-                                     stride, dt, args->reduce.predefined_op,
+                                     stride, dt, args->op,
                                      vector_op, alpha, mem_type);
 }
 
--- a/src/core/ucc_mc.h
+++ b/src/core/ucc_mc.h
@@ -120,9 +120,8 @@ static inline ucc_status_t ucc_mc_reduce_userdefined(void *src1, void *src2,
         .dt        = dt
     };
 
-    dt->ops.reduce.cb(&params);
-    return UCC_OK;
-}
+    return dt->ops.reduce.cb(&params);}
+
 
 static inline ucc_status_t ucc_dt_reduce(void *src1, void *src2,
                                          void *dst, size_t count,
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -258,7 +258,7 @@ typedef struct ucc_reduce_cb_params {
     void             *src2;      /*< input buffer: represents n_vectors buffers with
                                    offset "stride" between them */
     void             *dst;       /*< destination buffer */
-    size_t            n_vectors; /*< number of vetors from src2 to reduce */
+    size_t            n_vectors; /*< number of vectors from src2 to reduce */
     size_t            count;     /*< number of elements in one vector */
     size_t            stride;    /*< stride in bytes between the vectors in src2 */
     ucc_dt_generic_t *dt;        /*< pointer to user-defined datatype used for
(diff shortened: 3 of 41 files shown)


src/components/tl/ucp/allreduce/allreduce.h
```diff
@@ -43,7 +33,6 @@ ucc_status_t ucc_tl_ucp_allreduce_init(ucc_tl_ucp_task_t *task);
     } while (0)
 
 #define ALLREDUCE_TASK_CHECK(_args, _team)                                     \
-    CHECK_USERDEFINED_OP((_args), (_team));                                    \
```
bureddy, line 46: where is the check if user-defined "op" is provided for user-defined data types?

src/ucc/api/ucc_def.h
```diff
@@ -158,4 +160,21 @@ typedef size_t ucc_context_addr_len_t;
...
 
+/**
+ * @ingroup UCC_DATATYPES
+ * @brief Helper enum for generic/predefined datatype representation
+ *
+ */
+typedef enum {
+    UCC_DATATYPE_PREDEFINED = 0,
+    UCC_DATATYPE_GENERIC    = UCC_BIT(0),
+    UCC_DATATYPE_SHIFT      = 1,
```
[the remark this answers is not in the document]
vspetrov: agree will change, but what is safe ? 4 bits ?

src/components/tl/ucp/allreduce/allreduce.h
vspetrov: we don't need the check for user-defined op because now user-defined op is defined as part of user-defined type (flag on generic dt). So it is done during dt_reduce: https://github.com/openucx/ucc/pull/348/files#diff-1c664cedc0cc7a5a15ac577b975d287012b301de57fb55b92015a6842629d4d2R129

I check for generic dt and this always implies generic ops. This way internally i have only 1 runtime check in algorithms

src/ucc/api/ucc_def.h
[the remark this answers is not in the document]
bureddy: Maybe 3 bits to be in line with ucx ?

src/components/tl/ucp/allreduce/allreduce.h
bureddy: just checking if there is param validation check if reduce.op and reduce.context is there for  UCC_GENERIC_DT_OPS_FLAG_REDUCE

vspetrov: Oh i see: you mean when running ucc_dt_reduce ? I can add the check but it would be only to detect user bugs. The reason is: if user launches collective with reduction (allreduce, reduce, reduce_scatter(v)) and passes generic dtype into it then according to UCC spec user MUST provided custom op as well attached to the custom dtype. Otherwise, we can't get to ucc_dt_reduce with generic dt.

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+#define   UCC_DT_INT64    UCC_PREDEFINED_DT(3)
+#define   UCC_DT_INT128   UCC_PREDEFINED_DT(4)
+#define   UCC_DT_UINT8    UCC_PREDEFINED_DT(5)
+#define   UCC_DT_UINT16   UCC_PREDEFINED_DT(6)
+#define   UCC_DT_UINT32   UCC_PREDEFINED_DT(7)
+#define   UCC_DT_UINT64   UCC_PREDEFINED_DT(8)
+#define   UCC_DT_UINT128  UCC_PREDEFINED_DT(9)
+#define   UCC_DT_FLOAT16  UCC_PREDEFINED_DT(10)
+#define   UCC_DT_FLOAT32  UCC_PREDEFINED_DT(11)
+#define   UCC_DT_FLOAT64  UCC_PREDEFINED_DT(12)
```
manjugv, line 194: BFLOAT16?

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
- *  unsigned integers of various sizes, float 16, 32, and 64, and user-defined
- *  datatypes. The UCC_DT_USERDEFINED represents the user-defined datatype. The
- *  UCC_DT_OPAQUE is used to represent the user-defined datatypes for
- *  user-defined reductions. When UCC_DT_OPAQUE is used, the library passes the
- *  data to the user-defined reductions without any modifications.
+ *  collective and reduction operations. The standard operations (predefined)
+ *  are signed and unsigned integers of various sizes, float 16, 32, and 64, and
+ *  user-defined datatypes. User-defined datatypes cab created using
+ *  @ref ucc_dt_create_generic call and can support user-defined reduction
+ *  operations. Standard (predefined) reduction operations can be used only with
```
manjugv, line 174: cab? Typo? "User-defined datatypes are created using @ref ucc_dt_create generic interfaces. The User-defined datatypes can be used for both predefined and user-defined reduction operations, while predefined datatypes can only be used for predefined reduction operations."

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
  *  Description
  *
  *  @ref ucc_datatype_t represents the datatypes supported by the UCC library’s
- *  collective and reduction operations. The standard operations are signed and
- *  unsigned integers of various sizes, float 16, 32, and 64, and user-defined
- *  datatypes. The UCC_DT_USERDEFINED represents the user-defined datatype. The
- *  UCC_DT_OPAQUE is used to represent the user-defined datatypes for
- *  user-defined reductions. When UCC_DT_OPAQUE is used, the library passes the
- *  data to the user-defined reductions without any modifications.
+ *  collective and reduction operations. The standard operations (predefined)
```
manjugv, line 170: We can remove "standard operations" and use the term "predefined", IMO. If we a reference to term standard somewhere, let's replace it.

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+     * The pointer refers to application defined start-to-pack routine.
+     *
+     * @param [in]  context        User-defined context.
+     * @param [in]  buffer         Buffer to pack.
+     * @param [in]  count          Number of elements to pack into the buffer.
+     *
+     * @return  A custom state that is passed to the subsequent
+     *          @ref ucc_generic_dt_ops::pack "pack()" routine.
+     */
+    void* (*start_pack)(void *context, const void *buffer, size_t count);
```
manjugv, line 253: Did we decide to keep both start_pack and pack to have consistency with UCX or for some other reason?

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+     * @param [in]  dst      result buffer
+     * @param [in]  count    number of dtypes to reduce
+     * @param [in]  context  user-defined context as defined
+     *                       by @ref ucc_generic_dt_ops::reduce.ctx
+     */
+
+    struct {
+        void (*cb)(void *src1, void *src2, void *dst,
+                   ucc_count_t count, void *context);
+        void *ctx;
```
manjugv, line 353: There are so many contexts in the library. Should we change the name to something specific? "user_ctx" ?

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+ * "object" which provides a table of routines defining the operations for
+ * generic datatype manipulation. Typically, generic datatypes are used for
+ * integration with datatype engines provided with MPI implementations (MPICH,
+ * Open MPI, etc).
+ * The application is responsible for releasing the @a datatype_p  object using
+ * @ref ucc_dt_destroy "ucc_dt_destroy()" routine.
+ *
+ * @param [in]  ops          Generic datatype function table as defined by
+ *                           @ref ucc_generic_dt_ops_t .
+ * @param [in]  context      Application defined context passed to this
```
manjugv, line 373: context -> user_context or something else. Same reasoning as above: https://github.com/openucx/ucc/pull/348/files#r753809286

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+    UCC_OP_LAND,
+    UCC_OP_LOR,
+    UCC_OP_LXOR,
+    UCC_OP_BAND,
+    UCC_OP_BOR,
+    UCC_OP_BXOR,
+    UCC_OP_MAXLOC,
+    UCC_OP_MINLOC,
+    UCC_OP_AVG,
+   UCC_OP_LAST
```
manjugv, line 422: format

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+        void *ctx;
+    } reduce;
+} ucc_generic_dt_ops_t;
+
+
+/**
+ * @ingroup UCC_DATATYPE
+ * @brief Create a generic datatype.
+ *
+ * This routine creates a generic datatype object.
```
manjugv, line 362: Space after the sentence..

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+     * The pointer refers to application defined unpack routine.
+     *
+     * @param [in]  state          State as returned by
+     *                             @ref ucc_generic_dt_ops::start_unpack
+     *                             "start_unpack()" routine.
+     * @param [in]  offset         Virtual offset in the input stream.
+     * @param [in]  src            Source to unpack the data from.
+     * @param [in]  length         Length to unpack.
+     *
+     * @return UCS_OK or an error if unpacking failed.
```
Sergei-Lebedev, line 316: UCS_OK -> UCC_OK

src/ucc/api/ucc.h
```diff
@@ -199,33 +167,260 @@ typedef enum {
...
+     * @param [in]  src1     input buffer
+     * @param [in]  src2     input buffer
+     * @param [in]  dst      result buffer
+     * @param [in]  count    number of dtypes to reduce
+     * @param [in]  context  user-defined context as defined
+     *                       by @ref ucc_generic_dt_ops::reduce.ctx
+     */
+
+    struct {
+        void (*cb)(void *src1, void *src2, void *dst,
```
Sergei-Lebedev, line 351: for discussion, does it make sense to add memory type argument for reduction callback to support custom gpu reduction operations

src/components/tl/ucp/allreduce/allreduce.h
vspetrov: @bureddy is that what you meant?

src/ucc/api/ucc.h
vspetrov: @manjugv plz note, with this proposal if user wants CUSTOM reduction even on pre-defined dtype he will still need to wrap the simple dtype into "generic". This is some additional actions from user (w/o perf penalty though since it is light-weight) but it simplifies internal dt/op checks logic - this is why we went with this proposal.

vspetrov: So, typo fixed. the description is still basically the same.

src/ucc/api/ucc.h
vspetrov: i believe consistency with UCX and easier integration at tl/ucp level were reasons during WG call.

src/ucc/api/ucc.h
vspetrov: i called it "cb_ctx" : it is clear than it is the void *ctx that is passed to cb function.

src/ucc/api/ucc.h
vspetrov: @manjugv sorry, don't get it : what space?

src/ucc/api/ucc.h
vspetrov: i think it adds the flexibility - makes sense to extend. @manjugv @bureddy wdyt?

vspetrov: i extended the PR currently with mem_type in reduce_cb, so that you could check how it looks

Sergei-Lebedev: now I'm thinking of it, probably it's better to have one function pointer per memory type, because we might need to call this function from inside kernel so function should be __device__ function, wdyt?

vspetrov: @Sergei-Lebedev this complicates things a lot actually, but i see your point.
1st: I don't think that passing a device pointer through the host code (which is UCC interface) is even possible: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#function-pointers
2nd: Even if it was, then we would have to define a notion of "device" code in UCC.h which we don't have currently

On the other hand, with just mem_type as argument we can't really support "triggered" post with custom reductions. But we will be able to support MPI semantics still (user can launch kernel from its host reduce_cb). So, mem_type can serve that last case. But passing device code ptr is tricky imho.

Sergei-Lebedev: > 1st: I don't think that passing a device pointer through the host code (which is UCC interface) is even possible: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#function-pointers

Right, taking pointer from host is illegal, but I think it should be possible with cudamemcpyfromsymbol
> 2nd: Even if it was, then we would have to define a notion of "device" code in UCC.h which we don't have currently
Why do we need notion of device code and memory type is not enough?

vspetrov: What is the flow from users perspective?
User implements a cuda kernel which perfors reduction. Then user would use smth like cudaGetSymbolAddress inside his code and pass this as reduce.cb to ucc. Is that correct understanding?

What i meant with "notion of device code" is: we need to document exactly what reduce.cb can store. In most cases it is just a ptr to cpu function, in some other case it can be a ptr to a cpu function that runs a kernel on CUDA memory (and other memory types), 3rd case it can be a ptr obtained via cudaGetSymbolAddress. Say we add more flags to ucc_datatype_generic to distinguish between these cases. Now for case 3 you documentation basically must specify: the reduce.cb ptr stores a ptr to "device" code function that UCC will invoke internally also from the device.

src/ucc/api/ucc.h
manjugv: It seems like you are not using the space up to 80 chars.

src/ucc/api/ucc.h
manjugv: @vspetrov @Sergei-Lebedev  Memory type is a good addition.

vspetrov: Per discussion: 
we remove mem_type for now,
specify explicitly that only host fn ptr is supported (in the future can extend with flag),
change signature to support reduce_multi,
pack all params into struct for flexibility

src/ucc/api/ucc.h
```diff
@@ -199,34 +180,283 @@ typedef enum {
...
+     * @brief User-defined reduction callback
+     *
+     * The pointer refers to user-defined reduction routine.
+     *
+     * @param [in]  src1     input buffer
+     * @param [in]  src2     input buffer
+     * @param [in]  dst      result buffer
+     * @param [in]  count    number of dtypes to reduce
+     * @param [in]  cb_ctx   user-defined context as defined
+     *                       by @ref ucc_generic_dt_ops::reduce.cb_ctx
```
bureddy, line 385: remove arg description from here and add it ucc_reduce_cb_params_t struct members

src/components/tl/ucp/allreduce/allreduce.h
bureddy: yes. I meant to check the param valiation as per spec i.e is user provided custom op with custom dtype or not

src/core/ucc_mc.h
```diff
@@ -103,17 +105,38 @@ ucc_status_t ucc_mc_reduce_multi_alpha(
...
+        .count     = count,
+        .stride    = stride,
+        .dt        = dt
+    };
+
+    dt->ops.reduce.cb(&params);
+    return UCC_OK;
+}
+
+static inline ucc_status_t ucc_dt_reduce(void *src1, void *src2,
```
bureddy, line 127: why "const" is removed for src1 , src2?

src/ucc/api/ucc.h
```diff
@@ -199,34 +180,283 @@ typedef enum {
...
+} ucc_generic_dt_ops_flags_t;
+
+/**
+ * @ingroup UCC_DATATYPE
+ * @brief Descriptor of user-defined reduction callback
+ *
+ * This structure is the argument to the reduce.cb callback. It must implement
+ * the reduction of n_vectors + 1 data vectors each containing "count" elements.
+ * First vector is "src1", other n_vectors have start address
+ * v_j = src + count * dt_extent * stride * j.
```
bureddy, line 249: s/scc1/src2

src/core/ucc_mc.h
vspetrov: we don't have it for reduce multi neither, the reason is: src2 and dst can actually overlap (however in a very special way that still allows dst[j] = sum (src + stride * i ) [j], the dst can be equal to src2 , e.g. - it will be "inplace" reduce). This is why don't have const in reduce_cb as well.

src/ucc/api/ucc.h
vspetrov: done

src/components/tl/ucp/allreduce/allreduce.h
vspetrov: after discussion: we guard user for all the cases, e.g. if user sets REDUCE flag on generic dt but forgets to set reduce.cb and leaves garbage there. so, leaving as is.

bureddy (approved):

src/ucc/api/ucc.h
```diff
@@ -199,34 +180,281 @@ typedef enum {
...
+     * @ingroup UCC_DATATYPE
+     * @brief User-defined reduction callback
+     *
+     * The pointer refers to user-defined reduction routine.
+     *
+     * @param [in]  params reduction descriptor
+     */
+
+    struct {
+        void (*cb)(const ucc_reduce_cb_params_t *params);
```
Sergei-Lebedev, line 387: why return type is void and not ucc_status_t?

src/utils/ucc_log.h
```diff
@@ -115,10 +116,6 @@ static inline const char* ucc_datatype_str(ucc_datatype_t dt)
         return "int128";
     case UCC_DT_UINT128:
         return "uint128";
-    case UCC_DT_USERDEFINED:
```
Sergei-Lebedev, line 118: we still need to return some str in case of userdefined for example to print it in ucc_coll_str, add
```
if (!UCC_DT_IS_PREDEFINED(dt)) {
    return "userdefined";
}
```

Sergei-Lebedev (approved):

src/ucc/api/ucc.h
vspetrov: no big reason i guess. former reduce_cb in ucc.h used to be void probably because this way it is done in MPI spec. But i think we can make it ucc_status_t for generality.

PR #350, opened 2021-11-16
CORE: automatic TLs build discovery

vspetrov:
## What
Automatic TLs discovery during build process.
new (optional) configure parameter --with-tls= to specify list of tls to build.

## Why ?
When 3rd party implements custom TL integration should be entirely independent from open-source upstream UCC cude, including build.

### DIFF
--- a/config/m4/check_tls.m4
+++ b/config/m4/check_tls.m4
@@ -1,0 +1,22 @@
+#
+# Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+# See file LICENSE for terms.
+#
+
+AC_DEFUN([CHECK_TLS],[
+    AC_ARG_WITH([tls],
+        [AS_HELP_STRING([--with-tls=(tl1,tl2 | all)], [Enable build of TLs])],
+        [TLS_REQUIRED=${with_tls}], [TLS_REQUIRED=all])
+    m4_include([config/m4/tls_list.m4])
+])
+
+AC_DEFUN([CHECK_TLS_REQUIRED], [
+    tl_name=$1
+    CHECKED_TL_REQUIRED=n
+    AS_IF([ test "$TLS_REQUIRED" = "all" ], [CHECKED_TL_REQUIRED=y],
+    [
+       for t in $(echo ${TLS_REQUIRED} | tr "," " "); do
+           AS_IF([ test "$t" == "$tl_name" ], [CHECKED_TL_REQUIRED=y], [])
+       done
+    ])
+])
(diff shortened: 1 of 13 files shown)


src/components/tl/nccl/configure.m4
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2021      Mellanox Technologies. All rights reserved.
+# $COPYRIGHT$
+# Additional copyrights may follow
+
+CHECK_NCCL
```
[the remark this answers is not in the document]
vspetrov: done

config/m4/check_tls.m4
```diff
@@ -0,0 +1,22 @@
+#
+# Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+# See file LICENSE for terms.
+#
+
+AC_DEFUN([CHECK_TLS],[
+    AC_ARG_WITH([tls],
```
bureddy, line 7: what if transport is specified in "tls" but forgot/missed --with-<tls>

vspetrov: i don't understand. what is the configure cmd line you are talking about ?
If "--with-tls" is not provided then it equals to "all". All that are found under src/components/tls

bureddy: I mean "--with-tls=ucp,nccl,sharp " specified but not  --with-nccl and --with-sharp

vspetrov: first param only defines the TLs that are tried to be compiled. if with-tls has nccl nccl will first run CHECK_NCCL from nccl/configure.m4 and if it is not available then tl/nccl is not compiled. bottom line "--with-nccl" is used inside CHECK_NCCL which in turn only performed if nccl is part of required tls list.

bureddy (approved):

Sergei-Lebedev (approved):

manjugv:
@vspetrov 

1. Can you please fill these FAQs? 
https://github.com/openucx/ucc/wiki/FAQ#8-how-to-compile-all-tls
https://github.com/openucx/ucc/wiki/FAQ#9-how-to-compile-a-specific-tl

2. Can you update the copyrights? Also, the copyright in build files and "c" files are different.

manjugv:
@vspetrov Maybe it is worthwhile to port this PR to the release. Thoughts?

vspetrov:
@manjugv i'm not sure about port into v1.0.x. It is not a bugfix and we don't have any explicit request for it. I don't see it is necessary. Regarding copyrights: i've updated some. We probably need to review it separately (do we need to change to NV copyrights?)

vspetrov:
bot:retest

manjugv:
> @manjugv i'm not sure about port into v1.0.x. It is not a bugfix and we don't have any explicit request for it. I don't see it is necessary.

That is fine.

PR #353, opened 2021-11-23
TL/UCP: reduce_scatter

vspetrov:
## What
Enables reduce-scatter implementation in TL/UCP.

## Why ?
Main API. Also RS is needed for 2lvl Allreduce.

## How ?
- Enables Knomial RS to be used directly for reduce-scatter (not only as part of sra knomial allreduce)
- Implements ring reduce-scatter. Bi or uni-directional (option, default bi-directional). 
- RS knomial falls back to ring for non-power K case
- Adds gtests


vspetrov:
replacing with #361

PR #367, opened 2021-12-09
UTIL: ucc cpu arch specific functions

Sergei-Lebedev:
## What
Adding CPU architecture specific functions

## Why ?
Required to implement TL CUDA and TL SHM

## How ?
Copy of needed functions from UCS

### DIFF
--- a/src/utils/arch/aarch64/cpu.h
+++ b/src/utils/arch/aarch64/cpu.h
@@ -1,5 +1,5 @@
 /**
-* Copyright (C) Mellanox Technologies Ltd. 2001-2015.  ALL RIGHTS RESERVED.
+* Copyright (C) Mellanox Technologies Ltd. 2001-2021.  ALL RIGHTS RESERVED.
 * Copyright (C) ARM Ltd. 2016-2020.  ALL RIGHTS RESERVED.
 * Copyright (C) Stony Brook University. 2016-2020.  ALL RIGHTS RESERVED.
 *
@@ -30,5 +30,9 @@
 #define ucc_memory_bus_fence()        ucc_aarch64_dmb(oshsy)
 #define ucc_memory_bus_store_fence()  ucc_aarch64_dmb(oshst)
 #define ucc_memory_bus_load_fence()   ucc_aarch64_dmb(oshld)
+
+#define ucc_memory_cpu_fence()        ucc_aarch64_dmb(ish)
+#define ucc_memory_cpu_store_fence()  ucc_aarch64_dmb(ishst)
+#define ucc_memory_cpu_load_fence()   ucc_aarch64_dmb(ishld)
 
 #endif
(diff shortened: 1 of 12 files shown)


src/utils/arch/aarch64/cpu.h
```diff
@@ -0,0 +1,34 @@
+/**
+* Copyright (C) Mellanox Technologies Ltd. 2001-2015.  ALL RIGHTS RESERVED.
```
vspetrov, line 2: 2021

vspetrov (approved):

shimmybalsam (approved):

shimmybalsam (commented):
Please add isync function as well

src/utils/arch/aarch64/cpu.h
Sergei-Lebedev: fixed

Sergei-Lebedev:
> Please add isync function as well

added

src/utils/ucc_compiler_def.h
```diff
@@ -25,6 +25,11 @@
 #define ucc_likely        ucs_likely
 #define ucc_unlikely      ucs_unlikely
 
+/**
+ * Prevent compiler from reordering instructions
+ */
+#define ucc_compiler_fence()       asm volatile(""::: "memory")
```
vspetrov, line 31: @Sergei-Lebedev i think i was not right saying that Shimmy needs just a compiler instruction barrier. We need arch specific CPU instruction barrier.
in hcoll we use libocoms which defines isync:
power: https://github.com/uccs/libocoms/blob/4aa64d0105317badb244e7433ea29aa03a014507/ocoms/sys/powerpc/atomic.h#L34

arm: https://github.com/uccs/libocoms/blob/4aa64d0105317badb244e7433ea29aa03a014507/ocoms/sys/arm64/atomic.h#L76

@bureddy we need an instruction sync primitive for shared memory collectives (remember ppc data race) - does ucx/ucs have that or we need to copy paste from libocoms?

manjugv:
@Sergei-Lebedev I'm missing context here. Why do you need to make a copy of these functions?

Sergei-Lebedev:
> @Sergei-Lebedev I'm missing context here. Why do you need to make a copy of these functions?

currently those functions are not part of ucs headers and it's unlikely they ever be public

manjugv:
> > @Sergei-Lebedev I'm missing context here. Why do you need to make a copy of these functions?
> 
> currently those functions are not part of ucs headers and it's unlikely they ever be public

Do you know why is it not public?

manjugv:
@Sergei-Lebedev Once @bureddy confirms it is ready to go.

src/utils/ucc_compiler_def.h
bureddy: ucx/ucs  ucc_memory_cpu_load_fence() should be enough for pcc shared memory.

bureddy: here was the requirement  on PPC
```

producer 
//store data
for( .............)
 shared_array[i] = 0;                                                                                                                  
ppc_write_fence();   //an lwsync is sufficient for DDR memory writes
//Notify consumers
shared_flag = 1;

Consumer
while (!shared_flag);    //blocking wait on the shared flag (ideally should be non-blocking)
isync();                             //Don't let the speculation unit move instructions ahead of the while loop check
//access the shared_array
int x = shared_array[y];     //read the updated shared state

```

bureddy (approved):

PR #371, opened 2021-12-14
UTIL: fix ucs mpool v1.13.x

vspetrov:
Otherwise segv happens on uninit obj_str if some reqs are not returned to mpool

### DIFF
--- a/src/utils/ucc_mpool.c
+++ b/src/utils/ucc_mpool.c
@@ -43,7 +43,7 @@ ucc_status_t ucc_mpool_init(ucc_mpool_t *mp, size_t priv_size, size_t elem_size,
                             ucc_mpool_ops_t *ops, ucc_thread_mode_t tm,
                             const char *name)
 {
-    ucs_mpool_ops_t *ucs_ops = ucc_malloc(sizeof(*ucs_ops), "mpool_ops");
+    ucs_mpool_ops_t *ucs_ops = ucc_calloc(1, sizeof(*ucs_ops), "mpool_ops");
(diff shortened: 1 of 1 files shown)


src/utils/ucc_mpool.c
```diff
@@ -57,6 +57,9 @@ ucc_status_t ucc_mpool_init(ucc_mpool_t *mp, size_t priv_size, size_t elem_size,
     ucs_ops->chunk_release = ucc_mpool_chunk_release_wrapper;
     ucs_ops->obj_init      = NULL;
     ucs_ops->obj_cleanup   = NULL;
+#if HAVE_UCS_MPOOL_OPS_T_OBJ_STR
+    ucs_ops->obj_str       = NULL;
```
Sergei-Lebedev, line 61: maybe instead do calloc for ucs_ops or memset 0 to avoid conditional compilation?

vspetrov: right, i can do that. i just thought to be explicit in case we want to define obj_str to smth (same way ucx uses it). don't really have a strong opinion. can switch to memset. both fixes are straight forward. @bureddy wdyt?

bureddy: lets use memset to have a cleaner code.  we can add config check if we have use case to set obj_str

Sergei-Lebedev (approved):

bureddy (approved):

manjugv:
WG Discussion: Port to v1.0x branch

PR #309, opened 2021-09-28
TL/UCP: add one-sided collective support

wfaderhold21:
## What
This PR will provide the necessary communication support for one-sided collectives in the TL/UCP component.

## Why ?
Currently, there is support for two-sided collectives. This provides support for one-sided collectives that will map to PGAS programming models including OpenSHMEM.

PR #374, opened 2021-12-17
TEST/GTEST: adds onesided alltoall test

wfaderhold21:
## What
Adds some onesided and onesided alltoall unit tests

## Why ?
This adds the tests missing from PR #309 and #323.

### DIFF
--- a/test/gtest/common/test_ucc.cc
+++ b/test/gtest/common/test_ucc.cc
@@ -42,7 +42,7 @@ UccProcess::~UccProcess()
     EXPECT_EQ(UCC_OK, ucc_context_destroy(ctx_h));
     EXPECT_EQ(UCC_OK, ucc_finalize(lib_h));
     if (ctx_params.mask & UCC_CONTEXT_PARAM_FIELD_MEM_PARAMS) {
-        for (auto i = 0; i < 3; i++) {
+        for (auto i = 0; i < UCC_TEST_N_MEM_SEGMENTS; i++) {
             ucc_free(onesided_buf[i]);
         }
     }
@@ -405,7 +405,7 @@ void proc_context_create_mem_params(UccProcess_h proc, int id,
     ucc_status_t         status;
     ucc_context_config_h ctx_config;
     std::stringstream    err_msg;
-    ucc_mem_map_t        map[3];
+    ucc_mem_map_t        map[UCC_TEST_N_MEM_SEGMENTS];
 
     status = ucc_context_config_read(proc->lib_h, NULL, &ctx_config);
     if (status != UCC_OK) {
@@ -428,7 +428,7 @@ void proc_context_create_mem_params(UccProcess_h proc, int id,
     proc->ctx_params.oob.n_oob_eps         = ta->n_procs;
     proc->ctx_params.oob.oob_ep            = id;
     proc->ctx_params.mem_params.segments   = map;
-    proc->ctx_params.mem_params.n_segments = 3;
+    proc->ctx_params.mem_params.n_segments = UCC_TEST_N_MEM_SEGMENTS;
     status = ucc_context_create(proc->lib_h, &proc->ctx_params, ctx_config,
                                 &proc->ctx_h);
     ucc_context_config_release(ctx_config);
--- a/test/gtest/common/test_ucc.cc
+++ b/test/gtest/common/test_ucc.cc
@@ -414,7 +414,7 @@ void proc_context_create_mem_params(UccProcess_h proc, int id,
     }
     for (auto i = 0; i < UCC_TEST_N_MEM_SEGMENTS; i++) {
         proc->onesided_buf[i] =
-            ucc_malloc(UCC_TEST_MEM_SEGMENT_SIZE, "onesided_buffer");
+            ucc_calloc(UCC_TEST_MEM_SEGMENT_SIZE, 1, "onesided_buffer");
         EXPECT_NE(proc->onesided_buf[i], nullptr);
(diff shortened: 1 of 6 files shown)


Sergei-Lebedev:
ok to test

Sergei-Lebedev:
bot:retest

Sergei-Lebedev (approved):

test/gtest/common/test_ucc.cc
```diff
@@ -391,13 +399,60 @@ void proc_context_create(UccProcess_h proc, int id, ThreadAllgather *ta, bool is
     throw std::runtime_error(err_msg.str());
 }
 
+void proc_onesided_context_create(UccProcess_h proc, int id,
```
vspetrov, line 402: the name "proc_onesided_context_create" is misleading imho. I would rather called it proc_context_create_mem_params

test/gtest/common/test_ucc.cc
```diff
@@ -391,13 +399,60 @@ void proc_context_create(UccProcess_h proc, int id, ThreadAllgather *ta, bool is
...
+    ucc_context_config_h ctx_config;
+    std::stringstream    err_msg;
+    ucc_mem_map_t        map[3];
+
+    status = ucc_context_config_read(proc->lib_h, NULL, &ctx_config);
+    if (status != UCC_OK) {
+        err_msg << "ucc_context_config_read failed";
+        goto exit_err;
+    }
+    for (auto i = 0; i < 3; i++) {
```
vspetrov, line 415: i don't like 3 and 1<<20 being hardcoded and used in multiple places. Lets define macros: UCC_TEST_N_MEM_SEGMENTS, UCC_TEST_MEM_SEGMENT_SIZE

test/gtest/core/test_alltoall.cc
```diff
@@ -64,6 +64,66 @@ class test_alltoall : public UccCollArgs, public ucc::test
...
+        ctxs.resize(nprocs);
+        for (auto i = 0; i < nprocs; i++) {
+            ucc_coll_args_t *coll =
+                (ucc_coll_args_t *)calloc(1, sizeof(ucc_coll_args_t));
+
+            ctxs[i] =
+                (gtest_ucc_coll_ctx_t *)calloc(1, sizeof(gtest_ucc_coll_ctx_t));
+            ctxs[i]->args = coll;
+
+            sbuf        = job->procs[i]->onesided_buf[0];
```
vspetrov, line 83: Is this correct - taking procs[i] ?
The team being tested can be arbitrary subset of Job: e.g., job size is 15 ranks, and team consists of ranks (1,3,5,7,9). Then, the nprocs here is team size = 5, but actual processes from job should be job->procs[1], [3], [5], [7], [9] and so on.

vspetrov (changes requested):

test/gtest/common/test_ucc.cc
wfaderhold21: done

test/gtest/common/test_ucc.cc
wfaderhold21: done

test/gtest/core/test_alltoall.cc
wfaderhold21: You're correct, the code was incorrect. 

I've updated the code to use the team instead of the job. I believe that should be correct.

test/gtest/core/test_alltoall.cc
```diff
@@ -35,35 +42,56 @@ class test_alltoall : public UccCollArgs, public ucc::test
...
-                coll->mask  |= UCC_COLL_ARGS_FIELD_FLAGS;
-                coll->flags |= UCC_COLL_ARGS_FLAG_IN_PLACE;
-                UCC_CHECK(ucc_mc_memcpy(coll->dst.info.buffer, ctxs[i]->init_buf,
-                                        ucc_dt_size(dtype) * single_rank_count * nprocs,
-                                        mem_type, UCC_MEMORY_TYPE_HOST));
+            if (is_onesided) {
+                sbuf        = team->procs[i].p->onesided_buf[0];
+                rbuf        = team->procs[i].p->onesided_buf[1];
+                work_buf    = (long *)team->procs[i].p->onesided_buf[2];
+                work_buf[0] = -1; /* initialize for onesided a2a algorithm */
```
vspetrov, line 55: @wfaderhold21 why this initialization to -1 is required? is it generic requirement for the user if onesided "work_buf" is provided or is it gtest specific?

wfaderhold21: @vspetrov It's a generic requirement. The a2a algorithm is expecting the value to start at -1.

vspetrov: @manjugv looks like this is then part of API/behavior which needs to be documented. Also i think we must have special constant for that (similar to shmem) instead of "-1" value.

manjugv: @wfaderhold21 "-1" might be an odd choice. Should we define a constant instead like we do in SHMEM?

wfaderhold21: @vspetrov @manjugv I've added a line in the API stating that the buffer must be initialized to 0 and updated a2a code to match. Is this sufficient?

manjugv: As discussed offline, If "0" works for all the use-cases, this is the least intrusive way to do it, IMO. @vspetrov does this work for you?

vspetrov: it does for me

vspetrov (approved):

PR #384, opened 2022-01-11
DOCS: Adding Collective Operations Semantics

manjugv:
The collective operation semantics are not documented. This PR is an effort to capture the current semantics briefly.


docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+participants of the team. All participants of the team are required to call the
+collective operation. Each participant is represented by the endpoint that is
+unique to the team used for the collective operation. This section provides a
+set of routines for launching, progressing, and completing the collective
+operations.
 
-\b Invocation semantics: The ucc\_collective\_init routine is a non-blocking collective operation to initialize the buffers, operation type, reduction type, and other information required for the collective operation. All participants of the team should call the initialize operation. The routine returns once the participants enter the collective initialize operation. The collective operation is invoked using a ucc\_collective\_post operation. ucc\_collective\_init\_and\_post operation initializes as well as post the collective operation.
+**Invocation semantics**: The ucc\_collective\_init routine is a non-blocking
+collective operation to initialize the buffers, operation type, reduction type,
+and other information required for the collective operation. All participants of
```
vspetrov, line 12: _The routine returns once the participants enter the collective initialize operation._

This is a strong statement - i never realized we have such a constraint. Does this mean that ucc_collective_init is potentially blocking, ie it says that the call "may not return" until everybody enter collective. I think this is incorrect.

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+participants enter the collective initialize operation. The collective operation
+is invoked using a ucc\_collective\_post operation.
+ucc\_collective\_init\_and\_post operation initializes as well as post the
+collective operation.
 
-\b Collective type: The collective operation supported by UCC is defined by the enumeration ucc\_coll\_type\_t. It supports three types of collective operations: (a) UCC\_{ALLTOALL,ALLTOALLV, ALLGATHER, ALLGATHERV, ALLREDUCE, REDUCE_SCATTER, REDUCE_SCATTERV, BARRIER} operations where all participants contribute to the results and receive the results (b) UCC\_{REDUCE, GATHER, GATHERV, FANIN} where all participants contribute to the result and one participant receives the result. The participant receiving the result is designated as root. (c) UCC\_{BROADCAST, SCATTER, SCATTERV, FANOUT} where one participant contributes to the result, and all participants receive the result. The participant contributing to the result is designated as root.
+**Collective Type**: The collective operation supported by UCC is defined by the
+enumeration ucc\_coll\_type\_t. The semantics are briefly described here,
```
vspetrov, line 20: are -> is (briefly described)

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+however in most cases it agrees with the semantics of collective operations in
+the popular programming models such as MPI and OpenSHMEM. When they differ, the
+semantics changes are documented. All collective operations operate on the team.
```
vspetrov, line 23: operate -> are executed (otherwise "operations operate")

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+For the collective operations defined in the enumeration, all participants of
```
vspetrov, line 24: "participants participate"

Lets rephrase: All team members participate in collective operation.

First part "For the collective operations defined in the enumeration" is unnecessary imho, since we don't support anything else (which is not defined in enumeration)

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+the team participate in the collective operations. Further the team should be
+created with endpoints, where the “eps” should be ordered and contiguous.
```
vspetrov, line 26: _Further the team should be created with endpoints, where the “eps” should be ordered and contiguous._

Is this a requirement? I thought that potentially we can support creation of teams with non-contig eps. What is meant here?

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+operations where all participants contribute to the results and receive the
+results (b) UCC\_{REDUCE, GATHER, GATHERV, FANIN} where all participants
+contribute to the result and one participant receives the result. The
+participant receiving the result is designated as root. (c) UCC\_{BROADCAST,
+SCATTER, SCATTERV, FANOUT} where one participant contributes to the result, and
+all participants receive the result. The participant contributing to the result
+is designated as root.
 
-\b Ordering: The team can be configured for ordered collective operations or unordered collective operations. For unordered collectives, the user is required to provide the “tag”, which is an unsigned 64-bit integer. 
++ The UCC\_COLL\_TYPE\_BCAST operation moves the data from the src buffer of
```
vspetrov, line 38: lets remove "src" and "dest" words here, because it is a bit misleading. In coll_args for bcast we init args.src.info.buffer for ALL ranks, root and non root (args.dst.info is not initialized).

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
 
-\b Synchronized and Non-Synchronized Collectives: In the synchronized collective model, on entry, the participants cannot read or write to other participants without ensuring all participants have entered the collective operation. On the exit of the collective operation, the participants may exit after all participants have completed the reading or writing to the buffers.
++ The UCC\_COLL\_TYPE\_BARRIER synchronizes all participants of the collective
+operation. In this routine, first, each participant waits for all other
+participants to enter the operation. Then, once it learns the entry of all other
+participants into the operation, it exits the operation completing it locally.
 
-In the non-synchronized collective model, on entry, the participants can read or write to other participants. If the input and output buffers are defined on the team and RMA operations are used for data transfer, it is the responsibility of the user to ensure the readiness of the buffer. On exit, the participants may exit once the read and write to the local buffers are completed. 
++ The UCC\_COLL\_TYPE\_FANIN gathers the data from all participants in the team
```
vspetrov, line 46: FANIN/FANOUT don't move data. These are essentially synchronizing primitives (like UCC_COLL_TYPE_BaRRIER). Fanin can be viewed as 0 byte reduce/gather, and fanout is 0 byte bcast/scatter.

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+The dest buffer on root and src buffer on non-root participants is not defined.
+
++ In the UCC\_COLL\_TYPE\_GATHER operation, each participant of the collective
+operation sends data to the root participant. All participants send the same
+amount of data (block\_size) to the root. The size of the block is
+“dt\_elem\_size * count”. The total amount of data received by the root is equal
+to block_size * num\_participants. Here, the “count” represents the number of
+data elements. The
+"dt_elem_size" represents the size of the data element in bytes. The
+"num_participants" define the number of participants in the team. The data on
```
vspetrov, line 62: define -> defines

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+the root is placed in the receive buffer ordered by the “ep” ordering. For
+example, if the participants’ endpoints are ordered as “ep\_a” to “ep\_n”, the
```
vspetrov, line 64: How do we define "ordering" of non-contiguous EPs ? Say, ep is just a pointer - each process might have different ordering then (local pointers may differ).

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+data from the participant with ep_i is placed as an “ith” block on the receive
+buffer.
+
++ In the UCC\_COLL\_TYPE\_SCATTER operation, the root participant of the
+collective operation sends data to all other participants. It sends the same
+amount of data (block_size) to all participants. The size of the block
+(block_size) is “dt_elem_size * count”. The total amount of data received by the
```
vspetrov, line 71: received -> sent

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
++ The UCC\_COLL\_TYPE\_ALLREDUCE first performs an element-wise reduction on the
+src buffers of all participants. Then the result is distributed to all
+participants. After the operation, the results are available on the dst buffer
+of all participants. The size of src buffer and dst buffer is the same for all
+participants. The size of src buffer and dst buffer is the same, which is equal
+to “dt_elem_size * count”. Here, the “count” represents the number of data
+elements. The "dt_elem_size" represents the size of the data element in bytes.
+
++ The UCC\_COLL\_TYPE\_REDUCE\_SCATTER first performs a element-wise
+reduction on the src buffer and then scatters the result to the dst buffer. If
```
vspetrov, line 102: The sentence starts with "IF" but there is no "then"

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+the size of src buffer is “count * dt_elem_size”, where dt_elem_size is the
+number of bytes for the data type element and count is the number of elements of
+that datatype. Assuming the result can be divided into “n” blocks, the ith block
```
vspetrov, line 105: For ReduceScatter we must agree:
1. Either we only support the case when count % team_size == 0 -> MPI case.
2. If we are to support other case then we need to define how the remainder (count % team_size) is distributed. Then we must document it. E.g., the remainder can be distributed evenly among first (count % team_size) ranks - ie, each of those ranks will get +1 element

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,150 @@
...
+is placed in the receive buffer of endpoint “i”. Like other collectives, for
+this collective, the “ep” should be ordered and contiguous.
+
```
vspetrov, line 108: Missing description for: allgather

vspetrov (changes requested):

docs/doxygen/collectives.md
manjugv: I see. Small change to what you are saying - we don't have to write it as a 0-byte msg. Probably, we could write it as sync msg and leave it to the implementation, if it wants to use 0-byte or more that is up to the implementation.

I will rewrite the semantics. Let's see if it will capture the use cases.

docs/doxygen/collectives.md
manjugv: Huh .. The If is an error there.

docs/doxygen/collectives.md
manjugv: Done.

docs/doxygen/collectives.md
manjugv: Do we have use case for 2?

manjugv: Discussion with Val: Let's support case 1.

vspetrov (approved):

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,169 @@
...
+(block_size) is “dt_elem_size * count”. The total amount of data sent by the
+root is equal to block_size * num\_participants. Here, the “count” represents the
+number of data elements. The "dt_elem_size" represents the size of the data
+element in bytes. The "num_participants" represents the number of participants in
+the team.
+
++ In the UCC\_COLL\_TYPE\_ALLTOALL collective operation, all participants
+exchange a fixed amount of the data. For a given participant, the size of data
+in src buffer is “size”, where size is dt\_elem\_size * count * num_participants.
+Here, the “count” represents the number of data elements. The "dt_elem_size"
```
Sergei-Lebedev, line 97: i think changing to "Here, the “count” represents the number of data elements per destination" would be more clear

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,169 @@
...
+buffer.
+
+For UCC\_COLL\_TYPE\_BCAST operation, setting INPLACE flag has no impact.
+
+**The "v" Variant Collective Types**: The UCC\_COLL\_TYPE\_{ALLTOALLV, SCATTERV,
+GATHERV, and REDUCE\_SCATTERV} operations add flexibility to their counter
+parts (.i.e., ALLTOALL, SCATTER, GATHER, and REDUCE\_SCATTER) in that the
+location of data for the send and receive are specified by displacement arrays.
+
+**Reduction Types**: The reduction operation supported by UCC\_{ALLREDUCE,
```
Sergei-Lebedev, line 141: add reduce scatter: "supported by UCC\_{ALLREDUCE, REDUCE, REDUCE\_SCATTER, REDUCE\_SCATTERV"

docs/doxygen/collectives.md
```diff
@@ -1,20 +1,169 @@
...
+REDUCE} operation is defined by the enumeration ucc\_op\_t. The valid datatypes
```
Sergei-Lebedev, line 142: ucc\_op\_t is no longer exists, should be ucc\_reduction\_op\_t

manjugv:
@Sergei-Lebedev Addressed the feedback. I hadn't touched some parts of the text where you had suggested fixes. The fixes were small enough, I fixed in this PR rather than doing another PR.

Sergei-Lebedev (approved):

PR #388, opened 2022-01-13
TEST: mpi test fixes

vspetrov:
## What
- Always init args.coll_type of a test_case (force that by moving to constructor). Otherwise segv may happen during TestCase::str  (ucc_coll_type_str(args.coll_type)) if test_case was skipped w/o setting coll_type.
- Make MPI_Reduce_scatter_block in TestReduceScatter non-blocking to avoid potential dead-lock
- Remove check_sbuf (3 reasons: 1st it is unnecessary, 2nd it always takes 1 entry from mc mpool away, 3rd it complicates test_cases init/set/check logic).


bureddy (approved):

vspetrov:
bot:retest

vspetrov:
bot:retest

Sergei-Lebedev (approved):

PR #393, opened 2022-01-19
TL/UCP: piggy-back data exchange for one-sided

wfaderhold21:
## What
Removes the two extra data exchanges performed during one-sided context creation.  

## Why ?
There can be a performance improvement for context create when reducing the number of data exchanges. 

Performance difference using up to 32 nodes with 40 PPN between current master version and this PR. Times are in milliseconds:
| Number of Nodes | Master | This PR|
|-------------------------|-----------|------------|
|2|335.443|328.389|
|4|355.145|324.561|
|8|384.267|350.165|
|16|441.401|416.414|
|32|548.325|465.787|

_NOTE: one-sided memory buffers were allocated to 128 MB in order to highlight communication time._

### DIFF
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -473,7 +473,9 @@ ucc_status_t ucc_tl_ucp_get_context_attr(const ucc_base_context_t *context,
                 tl_error(ctx->super.super.lib, "failed to pack onesided information");
                 return ucc_status;
             }
-            memcpy(PTR_OFFSET(attr->attr.ctx_addr, ctx->ucp_addrlen), packed_data, packed_length);
+            if (ucc_likely(NULL != packed_data)) {
+                memcpy(PTR_OFFSET(attr->attr.ctx_addr, ctx->ucp_addrlen), packed_data, packed_length);
+            }
         }
     }
     if (attr->attr.mask & UCC_CONTEXT_ATTR_FIELD_WORK_BUFFER_SIZE) {
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -229,7 +229,7 @@ ucc_status_t ucc_tl_ucp_rinfo_destroy(ucc_tl_ucp_context_t *ctx)
     ucc_free(ctx->remote_info);
     ucc_free(ctx->rkeys);
     ctx->remote_info = NULL;
-    ctx->rkeys = NULL;
+    ctx->rkeys       = NULL;
 
     return UCC_OK;
 }
@@ -295,13 +295,13 @@ ucc_status_t ucc_tl_ucp_ctx_remote_populate(ucc_tl_ucp_context_t * ctx,
                                             ucc_mem_map_params_t   map,
                                             ucc_context_oob_coll_t oob)
 {
-    uint32_t                   size             = oob.n_oob_eps;
-    uint64_t                   nsegs            = map.n_segments;
-    ucp_mem_map_params_t       mmap_params;
-    ucp_mem_h                  mh;
-    ucs_status_t               status;
-    ucc_status_t               ucc_status;
-    int                        i;
+    uint32_t             size  = oob.n_oob_eps;
+    uint64_t             nsegs = map.n_segments;
+    ucp_mem_map_params_t mmap_params;
+    ucp_mem_h            mh;
+    ucs_status_t         status;
+    ucc_status_t         ucc_status;
+    int                  i;
 
     if (size < 2) {
         tl_error(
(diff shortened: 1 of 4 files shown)


src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -579,10 +459,24 @@ ucc_status_t ucc_tl_ucp_get_context_attr(const ucc_base_context_t *context,
...
+        packed_length = ctx->ucp_addrlen + ctx->n_rinfo_segs * (sizeof(size_t) * 3);
+        for (int i = 0; i < ctx->n_rinfo_segs; i++) {
+            packed_length += ctx->remote_info[i].packed_key_len;
+        }
+        attr->attr.ctx_addr_len = packed_length; 
     }
     if (attr->attr.mask & UCC_CONTEXT_ATTR_FIELD_CTX_ADDR) {
         memcpy(attr->attr.ctx_addr, ctx->worker_address, ctx->ucp_addrlen);
+        if (NULL != ctx->remote_info) {
+            ucc_status = ucc_tl_ucp_ctx_remote_pack_data(ctx, &packed_data, &packed_length);
```
vspetrov, line 471: isn't it a mem leak here? you allocate packed_data inside ucc_tl_ucp_ctx_remote_pack_data call , do memcpy but never free the packed_data array. Do you even need it? The core level will first call this get_attr routine with just UCC_CONTEXT_ATTR_FIELD_CTX_ADDR_LEN  bit and query the required space, then it will provide you a buffer that you can directly use.

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -392,30 +315,36 @@ ucc_status_t ucc_tl_ucp_ctx_remote_populate(ucc_tl_ucp_context_t * ctx,
...
         tl_error(ctx->super.super.lib, "failed to allocated %zu bytes",
-                 sizeof(ucc_tl_ucp_remote_info_t *) * size);
+                 sizeof(ucp_rkey_h *) * size);
         return UCC_ERR_NO_MEMORY;
     }
-
-    remote_info[rank] = (ucc_tl_ucp_remote_info_t *)ucc_calloc(
+    for (i = 0; i < size; i++) {
+        ctx->rkeys[i] =
+            (ucp_rkey_h *)ucc_calloc(nsegs, sizeof(ucp_rkey_h), "rkey_per_seg");
```
vspetrov, line 327: why to use double memory allocation for rkeys? Why not allocating rkys = malloc(size * nsegs * sizeof(ucp_rkey_h)) and then use an access macro, smth like
#define TL_UCP_REMOTE_RKEY(_ctx, _rank, _seg) (_ctx)->rkeys[ _seg * _ctx->size + _rank]

src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -161,33 +161,52 @@ ucc_tl_ucp_resolve_p2p_by_va(ucc_tl_ucp_team_t *team, void *va, ucp_ep_h *ep,
...
+
+    *segment  = -1;
+    core_rank = ucc_ep_map_eval(UCC_TL_TEAM_MAP(team), peer);
+    ucc_assert(UCC_TL_CORE_TEAM(team));
+    peer = ucc_get_ctx_rank(UCC_TL_CORE_TEAM(team), core_rank);
+
+    h = UCC_ADDR_STORAGE_RANK_HEADER(
+        &ctx->super.super.ucc_context->addr_storage, peer);
+    base_offset =
+        (ptrdiff_t)PTR_OFFSET(h, h->components[0].offset + ctx->ucp_addrlen);
```
vspetrov, line 182: are you assuming that tl/ucp component is loaded first ? (you use directly h->components[0]). I don't think it is correct. You should use a fn: ucc_get_team_ep_addr:

```
 85 static inline void *ucc_get_team_ep_addr(ucc_context_t *context,
  86                                          ucc_team_t *team, ucc_rank_t rank,
  87                                          unsigned long component_id)
  88 {
  89     ucc_context_addr_header_t *h    = ucc_get_team_ep_header(context, team,
  90                                                              rank);
  91     void                      *addr = NULL;
  92     int                        i;
  93     for (i = 0; i < h->n_components; i++) {
  94         if (h->components[i].id == component_id) {
  95             addr = PTR_OFFSET(h, h->components[i].offset);
  96             break;
  97         }
  98     }
  99     ucc_assert(NULL != addr);
 100     return addr;
 101 }
```

vspetrov (changes requested):

src/components/tl/ucp/tl_ucp_context.c
wfaderhold21: You are correct. I've removed leak as you have suggested.

Sergei-Lebedev (approved):

PR #394, opened 2022-01-20
TL/UCP: enable UCX active messages

gvallee:
## What
Enable UCX active messages.

## Why ?
A plugin is using UCX active messages which can only be enabled if the feature is activated when calling `ucp_init()`.

## How ?
N/A

### DIFF
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -42,7 +42,7 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
 
     ucp_params.field_mask =
         UCP_PARAM_FIELD_FEATURES | UCP_PARAM_FIELD_TAG_SENDER_MASK;
-    ucp_params.features = UCP_FEATURE_TAG;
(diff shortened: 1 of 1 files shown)


manjugv:
Do you need this in v1.0x branch ?

manjugv (approved):

vspetrov:
@gvallee My only concern is: does requesting AM support has any effect on performance or resource consumption of UCX? If yes, then we probably want to make it conditional (ie, if requested by plugin).

gvallee:
> Do you need this in v1.0x branch ?

It can wait until 1.1.x or later.

gvallee:
> @gvallee My only concern is: does requesting AM support has any effect on performance or resource consumption of UCX? If yes, then we probably want to make it conditional (ie, if requested by plugin).

The UCX team told us that it is just fine and i did not ever see any performance degradation due to it. Algorithms benefiting from AM can even be more efficient than similar ones relying on TM. Did you experience such a performance degradation?

vspetrov (approved):

Sergei-Lebedev (approved):

PR #396, opened 2022-01-21
CL/HIER: 2lvl alltoallv

vspetrov:
## What
Adds implementation of 2lvl alltoallv in the CL/HIER. 

## Why ?
Allows splitting a2av among TLs (e.g., TL/CUDA for intranode + TL/UCP for internode)

## How ?
CL/HIER allocates 2 sbgps: NODE and FULL, allocates a storage for counts/displs for 2 tasks and initializes it. Then the 2 tasks are launched concurrently.

### DIFF
--- a/src/components/cl/hier/alltoallv/alltoallv.c
+++ b/src/components/cl/hier/alltoallv/alltoallv.c
@@ -1,0 +1,240 @@
+/**
+ * Copyright (C) Mellanox Technologies Ltd. 2022.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "alltoallv.h"
+#include "../cl_hier_coll.h"
+#include "core/ucc_team.h"
+
+ucc_base_coll_alg_info_t
+    ucc_cl_hier_alltoallv_algs[UCC_CL_HIER_ALLTOALLV_ALG_LAST + 1] = {
+        [UCC_CL_HIER_ALLTOALLV_ALG_NODE_SPLIT] =
+            {.id   = UCC_CL_HIER_ALLTOALLV_ALG_NODE_SPLIT,
+             .name = "node_split",
+             .desc = "splitting alltoallv into two concurrent a2av calls"
+                     " withing the node and outside of it"},
+        [UCC_CL_HIER_ALLTOALLV_ALG_LAST] = {
+            .id = 0, .name = NULL, .desc = NULL}};
+
+static ucc_status_t ucc_cl_hier_alltoallv_start(ucc_coll_task_t *task)
+{
+    ucc_schedule_t *schedule = ucc_derived_of(task, ucc_schedule_t);
+
+    UCC_CL_HIER_PROFILE_REQUEST_EVENT(task, "cl_hier_alltoallv_start", 0);
+    return ucc_schedule_start(schedule);
+}
+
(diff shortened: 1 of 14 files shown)


src/components/cl/hier/cl_hier.c
```diff
@@ -38,6 +38,17 @@ static ucc_config_field_t ucc_cl_hier_lib_config_table[] = {
      ucc_offsetof(ucc_cl_hier_lib_config_t, sbgp_tls[UCC_HIER_SBGP_NET]),
      UCC_CONFIG_TYPE_STRING_ARRAY},
 
+    {"FULL_SBGP_TLS", "ucp",
+     "TLS to be used for FULL subgroup.\n"
+     "FULL subgroup contains all processes of the team",
+     ucc_offsetof(ucc_cl_hier_lib_config_t, sbgp_tls[UCC_HIER_SBGP_FULL]),
+     UCC_CONFIG_TYPE_STRING_ARRAY},
+
+    {"A2AV_NODE_THRESH", "0",
```
Sergei-Lebedev, line 47: i think we should keep existing naming scheme from TLs, i.e. COLLNAME_ALGNAME_PARAMNAME
for example ALLTOALLV_SPLIT_NODE_THRESH

Sergei-Lebedev (commented):
do you think we need similar algorithm for alltoall?

src/components/cl/hier/alltoallv/alltoallv.c
```diff
@@ -0,0 +1,240 @@
...
+
+    if (!SBGP_ENABLED(cl_team, NODE) || !SBGP_ENABLED(cl_team, FULL)) {
+        cl_debug(team->context->lib, "alltoallv requires NODE and FULL sbgps");
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    c64 = UCC_COLL_ARGS_COUNT64(&coll_args->args);
+    d64 = UCC_COLL_ARGS_DISPL64(&coll_args->args);
+
+    if ((c64 && !d64) || (!c64 && d64)) {
```
bureddy, line 129: `if (c64 ^ d64)` ?

bureddy (approved):

vspetrov:
> do you think we need similar algorithm for alltoall?

I think we can add if see a use case : does FB run pure a2a, not a2av ?

Sergei-Lebedev:
> > do you think we need similar algorithm for alltoall?
> 
> I think we can add if see a use case : does FB run pure a2a, not a2av ?

in pytorch we will run UCC alltoall in case of equal split https://github.com/facebookresearch/torch_ucc/blob/d29338636a968640228e6f2e6cbbf8eda288a07d/src/torch_ucc.cpp#L944.
IIRC param benchmark uses equal split for alltoall test

vspetrov:
> > > do you think we need similar algorithm for alltoall?
> > 
> > 
> > I think we can add if see a use case : does FB run pure a2a, not a2av ?
> 
> in pytorch we will run UCC alltoall in case of equal split https://github.com/facebookresearch/torch_ucc/blob/d29338636a968640228e6f2e6cbbf8eda288a07d/src/torch_ucc.cpp#L944. IIRC param benchmark uses equal split for alltoall test

@Sergei-Lebedev Added split node alltoall to the PR

Sergei-Lebedev (approved):

PR #361, opened 2021-12-06
Topic/tl ucp kn algs

vspetrov:
## What
_Describe what this PR is doing._ 

## Why ?
_Justification for the PR. If there is existing issue/bug please reference. For
bug fixes why and what can be merged in a single item._

## How ?
_It is optional but for complex PRs please provide information about the design,
architecture, approach, etc._

PR #413, opened 2022-02-01
TL/UCP: reduce_scatter_ring

vspetrov:
## What
Adds implementation of bi- uni- directional ring reduce_scatterin TL/UCP

## Why ?
Core API. Also used as part of 3-stage allreduce.

### DIFF
--- a/src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
+++ b/src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
@@ -12,6 +12,40 @@
 #include "utils/ucc_math.h"
 #include "utils/ucc_coll_utils.h"
 
+static inline void send_completion_common(void *request, ucs_status_t status,
+                                          void *user_data)
+{
+    ucc_tl_ucp_task_t *task = (ucc_tl_ucp_task_t *)user_data;
+
+    if (ucc_unlikely(UCS_OK != status)) {
+        tl_error(UCC_TASK_LIB(task), "failure in rs ring completion %s",
+                 ucs_status_string(status));
+        task->super.super.status = ucs_status_to_ucc_status(status);
+    }
+    task->send_completed++;
+    if (request) {
+        ucp_request_free(request);
+    }
+}
+
+static void send_completion_1(void *request, ucs_status_t status,
+                              void *user_data)
+{
+    ucc_tl_ucp_task_t *task = (ucc_tl_ucp_task_t *)user_data;
+
+    send_completion_common(request, status, user_data);
+    task->reduce_scatter_ring.s_scratch_busy[0] = 0;
+}
+
+static void send_completion_2(void *request, ucs_status_t status,
+                              void *user_data)
+{
+    ucc_tl_ucp_task_t *task = (ucc_tl_ucp_task_t *)user_data;
+
+    send_completion_common(request, status, user_data);
+    task->reduce_scatter_ring.s_scratch_busy[1] = 0;
+}
+
 static inline void ucc_ring_frag_count(ucc_tl_ucp_task_t *task, size_t count,
                                        ucc_rank_t block, size_t *frag_count)
 {
@@ -45,6 +79,7 @@ static inline void ucc_ring_frag_block_offset(ucc_tl_ucp_task_t *task,
     *block_offset = ucc_buffer_block_offset(count, size, block);
 }
 
+
 static inline ucc_status_t ucc_tl_ucp_test_ring(ucc_tl_ucp_task_t *task)
 {
     int polls = 0;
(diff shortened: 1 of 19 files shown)


vspetrov:
This is part of #361. It contains only reduce_scatter ring changes (and some utils that are required). @Sergei-Lebedev  @bureddy @manjugv plz review.

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -0,0 +1,364 @@
...
+                                    ucc_coll_task_t **    task_h)
+{
+
+    ucc_tl_ucp_team_t *tl_team  = ucc_derived_of(team, ucc_tl_ucp_team_t);
+    ucc_rank_t         size     = UCC_TL_TEAM_SIZE(tl_team);
+    size_t             count    = coll_args->args.dst.info.count * size;
+    ucc_datatype_t     dt       = coll_args->args.dst.info.datatype;
+    size_t             dt_size  = ucc_dt_size(dt);
+    ucc_memory_type_t  mem_type = coll_args->args.dst.info.mem_type;
+    int                bidir =
```
bureddy, line 300: align =

src/utils/ucc_coll_utils.c
```diff
@@ -291,3 +291,62 @@ void ucc_coll_str(const ucc_coll_task_t *task, char *str, size_t len)
...
+}
+
+ucc_status_t ucc_ep_map_create_inverse(ucc_ep_map_t map, ucc_ep_map_t *inv_map)
+{
+    ucc_ep_map_t inv;
+    ucc_rank_t   i, r;
+    ucc_rank_t   max_rank;
+
+    if (ucc_ep_map_is_reverse(&map)) {
+        inv = map;
```
bureddy, line 318: the function intention is to inverse the given map and here  your seting the same map to inv.  isn't possible to call inverse only if the map is not reverse?

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -0,0 +1,364 @@
...
+    block_count   = ucc_buffer_block_count(count, size, block);
+    *frag_offset  = ucc_buffer_block_offset(block_count, n_frags, frag);
+    *block_offset = ucc_buffer_block_offset(count, size, block);
+}
+
+static inline ucc_status_t ucc_tl_ucp_test_ring(ucc_tl_ucp_task_t *task)
+{
+    int polls = 0;
+    while (polls++ < task->n_polls) {
+        if (task->send_posted - task->send_completed <= 1 &&
```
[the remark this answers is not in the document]
vspetrov: To reduce sync/delays the algorithm uses 2 scratch buffers for incoming data (and to do reduction). They alternate each time. I don't care about send completions at all, i only need to make sure i have at least 1 free scratch to post next recv, ie number of outstanding sends must be < 2 (0 - both scratches free, 1 - only one).

src/utils/ucc_coll_utils.c
vspetrov: THe intent was to keep the client (tl/ucp) code clean. In case of bidirectional alg, tl/ucp ring always allocates just 2 rings 1 default + 1 reverse. I need to have inverse map for all rings to implement the alg. In this case when my ring are straight forward i could just  omit call to inverse in tl/ucp since it evaluates to the original ring. However, when we had custom rings used in the same alg (example was HCM, where rings would be topo dependent) then inverse compute is necessary. But the code stays the same, we can just pass a new "map" to the alg.

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -0,0 +1,364 @@
...
+    if (UCC_TL_UCP_TEAM_LIB(tl_team)->cfg.reduce_avg_pre_op &&
+        coll_args->args.op == UCC_OP_AVG) {
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    tl_schedule  = ucc_tl_ucp_get_schedule(tl_team, coll_args);
+    schedule     = &tl_schedule->super.super;
+    /* if count == size then we have 1 elem per rank, not enough
+       to split into 2 sets */
+    n_subsets    = (bidir && (count > size)) ? 2 : 1;
```
Sergei-Lebedev, line 319: does this condition work for inplace? i think should divide by size first in this case

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -0,0 +1,364 @@
...
+            return status;
+        }
+        if (task->recv_completed == size - 1) {
+            task->recv_posted = task->recv_completed = 0;
+            break;
+        }
+        ucc_assert(task->send_posted - task->send_completed <= 1);
+        ucc_assert(task->send_posted < size);
+
+        UCPCHECK_GOTO(ucc_tl_ucp_send_nb(s_scratch[(task->send_posted - 1) % 2],
```
Sergei-Lebedev, line 137: minor, instead of s_scratch[(task->send_posted - 1) % 2 just reduce_target

bureddy (approved):

vspetrov:
@Sergei-Lebedev @bureddy fixed scratch race by using custom callbacks as Sergey recommended in WG.

bureddy (approved):

Sergei-Lebedev (approved):

PR #414, opened 2022-02-02
UTIL: additions needed for tl/shm

shimmybalsam:
## What
Additions within util, needed to support tl/shm.

### DIFF
--- a/src/utils/ucc_coll_utils.c
+++ b/src/utils/ucc_coll_utils.c
@@ -229,6 +229,47 @@ ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
     return map;
 }
 
+ucc_ep_map_t ucc_ep_map_from_array_64(uint64_t **array, ucc_rank_t size,
+                                   ucc_rank_t full_size, int need_free)
+{
+    int          is_const_stride = 0;
+    ucc_ep_map_t map;
+    int64_t      stride;
+    ucc_rank_t   i;
+
+    map.ep_num = size;
+    if (size > 1) {
+        /* try to detect strided pattern */
+        stride          = (int64_t)(*array)[1] - (int64_t)(*array)[0];
+        is_const_stride = 1;
+        for (i = 2; i < size; i++) {
+            if (((int64_t)(*array)[i] - (int64_t)(*array)[i - 1]) != stride) {
+                is_const_stride = 0;
+                break;
+            }
+        }
+    }
+    if (is_const_stride) {
+        if ((stride == 1) && (size == full_size)) {
+            map.type = UCC_EP_MAP_FULL;
+        }
+        else {
+            map.type           = UCC_EP_MAP_STRIDED;
+            map.strided.start  = (uint64_t)(*array)[0];
+            map.strided.stride = stride;
+        }
+        if (need_free) {
+            ucc_free(*array);
+            *array = NULL;
+        }
+    } else {
+        map.type            = UCC_EP_MAP_ARRAY;
+        map.array.map       = (void *)(*array);
+        map.array.elem_size = sizeof(uint64_t);
+    }
+    return map;
+}
+
 static inline int
 ucc_coll_args_is_rooted(const ucc_base_coll_args_t *bargs)
 {
--- a/src/utils/ucc_coll_utils.c
+++ b/src/utils/ucc_coll_utils.c
@@ -188,8 +188,19 @@ size_t ucc_coll_args_msgsize(const ucc_base_coll_args_t *bargs)
     return 0;
 }
 
-ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
-                                   ucc_rank_t full_size, int need_free)
+static int64_t ucc_ep_map_get_elem(void **array, int i) {
+	return (int64_t) (*(ucc_rank_t **)(array))[i];
+}
+
+static int64_t ucc_ep_map_get_elem_64(void **array, int i) {
+	return (int64_t) (*(uint64_t **)(array))[i];
+}
+
+static ucc_ep_map_t
+ucc_ep_map_from_array_generic(void **array, ucc_rank_t size,
+                              ucc_rank_t full_size, int need_free,
+                              size_t elem_size,
+                              int64_t (*get_elem)(void**, int))
 {
     int          is_const_stride = 0;
     ucc_ep_map_t map;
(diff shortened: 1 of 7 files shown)


src/utils/ucc_coll_utils.c
```diff
@@ -229,6 +229,47 @@ ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
     return map;
 }
 
+ucc_ep_map_t ucc_ep_map_from_array_64(uint64_t **array, ucc_rank_t size,
```
vspetrov, line 232: this fn 99% code duplication of ucc_ep_map_from_array. Lets unify them. Add static ucc_ep_map_from_array_generic that takes void *array and size_t elem_size. Then
ucc_ep_map_from_array returns ucc_ep_map_from_array_generic(array, 4, ...), and the new one will call same with (array, 8, ...)

src/utils/ucc_sys.c
```diff
@@ -62,3 +62,38 @@ ucc_status_t ucc_sysv_free(void *addr)
...
+/*
+ * If a certain system constant (name) is undefined on the underlying system
+ * the sysconf routine returns -1.  ucs_sysconf return the negative value
+ * a user and the user is responsible to define default value or abort.
+ *
+ * If an error occurs sysconf modified errno and ucs_sysconf aborts.
+ *
+ * Otherwise, a non-negative values is returned.
+ */
+static long ucc_sysconf(int name)
```
vspetrov, line 75: why do we need this wrapper? looks like it does nothing

vspetrov (changes requested):

src/utils/ucc_sys.c
```diff
@@ -62,3 +62,38 @@ ucc_status_t ucc_sysv_free(void *addr)
 
     return UCC_OK;
 }
+
+/*
+ * If a certain system constant (name) is undefined on the underlying system
+ * the sysconf routine returns -1.  ucs_sysconf return the negative value
```
Sergei-Lebedev, line 68: ucs_sysconf -> ucc_sysconf

src/utils/ucc_coll_utils.c
shimmybalsam: Added generic function

src/utils/ucc_sys.c
shimmybalsam: removed

src/utils/ucc_coll_utils.c
```diff
@@ -224,11 +235,29 @@ ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
...
     return map;
 }
 
+ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
+                                   ucc_rank_t full_size, int need_free)
+{
+	size_t elem_size = sizeof(ucc_rank_t);
+    return ucc_ep_map_from_array_generic((void **) array, size, full_size,
+                                         need_free, elem_size,
+                                         &ucc_ep_map_get_elem);
```
vspetrov, line 249: @shimmybalsam i worry that doing that with fn pointer adds a high cost of fn call. This function (get_elem) is called O(team_size) times, ie looping through all the ranks in a team. I think having a branch is much cheaper in this case. So, i would declare 1 ucc_ep_map_get_elem function wich takes a flag is64:
static inline int64 ucc_ep_map_get_elem_function(void **array, int i, int is64)
{
    if (!is64) {
        return (int64_t) (*(ucc_rank_t **)(array))[i];
    } else {
        return (int64_t) (*(uint64_t **)(array))[i];
    }
}

Then ucc_ep_map_from_array_generic would also have is64 argument instead of fn pointer.

vspetrov (changes requested):

src/utils/ucc_coll_utils.c
shimmybalsam: done

src/utils/ucc_coll_utils.c
```diff
@@ -188,8 +188,19 @@ size_t ucc_coll_args_msgsize(const ucc_base_coll_args_t *bargs)
     return 0;
 }
 
-ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
-                                   ucc_rank_t full_size, int need_free)
+static int64_t ucc_ep_map_get_elem(void **array, int i, int is64)
```
vspetrov, line 191: inline

src/utils/ucc_coll_utils.c
```diff
@@ -188,8 +188,19 @@ size_t ucc_coll_args_msgsize(const ucc_base_coll_args_t *bargs)
...
+{
+    if (is64) {
+        return (int64_t) (*(uint64_t **)(array))[i];
+    } else {
+        return (int64_t) (*(ucc_rank_t **)(array))[i];
+    }
+}
+
+static ucc_ep_map_t
+ucc_ep_map_from_array_generic(void **array, ucc_rank_t size,
```
vspetrov, line 201: inline

src/utils/ucc_coll_utils.c
```diff
@@ -224,11 +238,27 @@ ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
...
-        map.array.elem_size = sizeof(ucc_rank_t);
+        map.array.elem_size = elem_size;
     }
     return map;
 }
 
+ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
+                                   ucc_rank_t full_size, int need_free)
+{
+	size_t elem_size = sizeof(ucc_rank_t);
```
vspetrov, line 249: alignment

src/utils/ucc_coll_utils.c
```diff
@@ -224,11 +238,27 @@ ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
...
+    return ucc_ep_map_from_array_generic((void **) array, size, full_size,
+                                         need_free, elem_size, 0);
+}
+
+ucc_ep_map_t ucc_ep_map_from_array_64(uint64_t **array, ucc_rank_t size,
+                                      ucc_rank_t full_size, int need_free)
+{
+	size_t elem_size = sizeof(uint64_t);
```
vspetrov, line 257: alignment

src/utils/ucc_coll_utils.c
```diff
@@ -224,11 +238,27 @@ ucc_ep_map_t ucc_ep_map_from_array(ucc_rank_t **array, ucc_rank_t size,
     } else {
         map.type            = UCC_EP_MAP_ARRAY;
         map.array.map       = (void *)(*array);
-        map.array.elem_size = sizeof(ucc_rank_t);
+        map.array.elem_size = elem_size;
```
Sergei-Lebedev, line 241: do you  need to pass elem_size to that function?
why not
```
map.array.elem_size = is64 ? sizeof(uint64_t): sizeof(ucc_rank_t)
```

shimmybalsam: you are correct. changed.

Sergei-Lebedev (approved):

PR #418, opened 2022-02-07
TL/CUDA: allgather ring

Sergei-Lebedev:
## What
Add NVLink rings detection and implement ring allgather in TL CUDA


Sergei-Lebedev:
replaced by https://github.com/openucx/ucc/pull/437

PR #419, opened 2022-02-07
BUILD: tl profiling fix

vspetrov:
## What
    TLs are autodetected, so they should add their profiling macros
    automatically as well.


bureddy (approved):

src/components/tl/sharp/configure.m4
```diff
@@ -11,6 +11,12 @@ AS_IF([test "$CHECKED_TL_REQUIRED" = "y"],
     if test $sharp_happy = "yes"; then
        tl_modules="${tl_modules}:sharp"
        tl_sharp_enabled=y
+       CHECK_NEED_TL_PROFILING(["tl_sharp"])
+       AS_IF([test "$TL_PROFILING_REQUIRED" = "y"],
+       [
+       AC_DEFINE([HAVE_PROFILING_TL_SHARP], [1], [Enable profiling for TL SHARP])
```
Sergei-Lebedev, line 17: alignment

PR #423, opened 2022-02-09
TL/CUDA: query nvml dev type in cuda11.0

Sergei-Lebedev:
## What
Support for device type detection in TL CUDA topo for older CUDA versions

## Why ?
Allows to use TL CUDA with CUDA 11.0 and above, currently it requires CUDA 11.4

## How ?
If nvmlDeviceGetNvLinkRemoteDeviceType function is not available do the following:
If remote device doesn't have PCI id it is not supported by CUDA TL return UCC_TL_CUDA_TOPO_DEV_TYPE_LAST, otherwise it's either NVSwitch or GPU
If can get remote device handle it's GPU otherwise it's NVSwtich


Sergei-Lebedev:
bot:retest

bureddy (approved):

vspetrov (approved):

PR #426, opened 2022-02-10
CORE: score fix on range boundary

vspetrov:
## What
Resolve boundary between neighbour ranges:
if ranges share a msg size as boundary leave that msgsize to the range wit higher score. That way components that report higher scores do not get overwritten at range boundary.

## Why ?
Example: TL/SHM can support bcast up to segment_size (say 4096), it enables the corresponding msg_range [0-4096]. However, w/o the fix msgsize=4096 would go to tl/ucp still.

### DIFF
--- a/src/coll_score/ucc_coll_score_map.c
+++ b/src/coll_score/ucc_coll_score_map.c
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * Copyright (C) Mellanox Technologies Ltd. 2021-2022.  ALL RIGHTS RESERVED.
  *
  * See file LICENSE for terms.
  */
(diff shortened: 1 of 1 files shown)


src/coll_score/ucc_coll_score_map.c
```diff
@@ -16,11 +16,40 @@ ucc_status_t ucc_coll_score_build_map(ucc_coll_score_t *score,
...
+
     map = ucc_malloc(sizeof(*map), "ucc_score_map");
     if (!map) {
         ucc_error("failed to allocate %zd bytes for score map", sizeof(*map));
         return UCC_ERR_NO_MEMORY;
     }
+
+    /* Resolve boundary between neighbour ranges:
+       if ranges share a msg size as boundary leave that msgsize
+       to the range wit higher score. That way components that report
```
Sergei-Lebedev, line 31: wit->with

Sergei-Lebedev (approved):

bureddy (approved):

PR #427, opened 2022-02-16
TL/UCP: reduce_scatterV ring algorithm

vspetrov:
## What
Implements ReduceScatterV ring algorithm in TL/UCP
Adds RSV to test/mpi and gtest

PR #432, opened 2022-02-25
TL/CUDA: Alltoallv implementation

kingchc:
## What
Implement TL CUDA alltoallv

## Why ?
Missing feature

## How ?
This PR is based on existing the TL CUDA Alltoall implementation and has the following differences to support Alltoallv.
- use info_v to get correct buffer ptr, counts, displacements, etc...
- extend shared memory region to store additional info for Alltoallv, e.g., counts, displacements, for proxies to access if applicable

### DIFF
--- a/src/components/tl/cuda/alltoallv/alltoallv_ce.c
+++ b/src/components/tl/cuda/alltoallv/alltoallv_ce.c
@@ -174,14 +174,14 @@ ucc_tl_cuda_alltoallv_ce_post_copies(ucc_tl_cuda_task_t *task)
                              args, peer_sync->src_displ, pdst) *
                          sdt_size;
             src       = PTR_OFFSET(task->alltoallv_ce.peer_map_addr_src[psrc],
-                             peer_sync->mem_info_src.offset);
+                                   peer_sync->mem_info_src.offset);
             src       = PTR_OFFSET(src, data_displ);
             peer_sync = TASK_SYNC(task, pdst);
             ucc_assert(
                 (rdt_size * ucc_coll_args_get_count(args, peer_sync->dst_cnts,
                                                     psrc)) == data_size);
             dst        = PTR_OFFSET(task->alltoallv_ce.peer_map_addr_dst[pdst],
-                             peer_sync->mem_info_dst.offset);
+                                    peer_sync->mem_info_dst.offset);
             data_displ = ucc_coll_args_get_displacement(
                              args, peer_sync->dst_displ, psrc) *
                          rdt_size;
--- a/src/components/tl/cuda/alltoallv/alltoallv_ce.c
+++ b/src/components/tl/cuda/alltoallv/alltoallv_ce.c
@@ -37,18 +37,25 @@ ucc_status_t ucc_tl_cuda_alltoallv_setup_start(ucc_tl_cuda_task_t *task)
     ucc_status_t        status;
     ucc_coll_args_t    *args = &TASK_ARGS(task);
 
-    // for A2av: copy counts and displ. for proxies (if any) to access
-    if (team->topo->num_proxies > 0 &&
-        UCC_COLL_TYPE_ALLTOALLV == args->coll_type) {
-        size_t elem_size = UCC_COLL_ARGS_COUNT64(args) ? 8 : 4;
-        memcpy(sync->src_cnts, args->src.info_v.counts,
-               elem_size * UCC_TL_TEAM_SIZE(team));
-        memcpy(sync->dst_cnts, args->dst.info_v.counts,
-               elem_size * UCC_TL_TEAM_SIZE(team));
-        memcpy(sync->src_displ, args->src.info_v.displacements,
-               elem_size * UCC_TL_TEAM_SIZE(team));
-        memcpy(sync->dst_displ, args->src.info_v.displacements,
-               elem_size * UCC_TL_TEAM_SIZE(team));
+    // For Alltoallv: copy counts and displ. to SHM for remote GPUs to access (if required)
+    if (UCC_COLL_TYPE_ALLTOALLV == args->coll_type) {
+        int    i;
+        size_t rdt_size = ucc_dt_size(task->alltoallv_ce.rdt);
+        size_t sdt_size = ucc_dt_size(task->alltoallv_ce.sdt);
+        for (i = 0; i < UCC_TL_TEAM_SIZE(team); ++i) {
+            sync->alltoallv_ce.sbytes[i] =
+                sdt_size * (size_t)ucc_coll_args_get_count(
+                               args, task->alltoallv_ce.scnts, i);
+            sync->alltoallv_ce.rbytes[i] =
+                rdt_size * (size_t)ucc_coll_args_get_count(
+                               args, task->alltoallv_ce.rcnts, i);
+            sync->alltoallv_ce.sdispl_bytes[i] =
+                sdt_size * (size_t)ucc_coll_args_get_displacement(
+                               args, task->alltoallv_ce.sdispl, i);
+            sync->alltoallv_ce.rdispl_bytes[i] =
+                rdt_size * (size_t)ucc_coll_args_get_displacement(
+                               args, task->alltoallv_ce.rdispl, i);
+        }
     }
     memcpy(&sync->mem_info_src, &task->alltoallv_ce.mem_info_src,
            sizeof(ucc_tl_cuda_mem_info_t));
@@ -126,8 +133,6 @@ ucc_status_t ucc_tl_cuda_alltoallv_ce_post_copies(ucc_tl_cuda_task_t *task)
     void                       *src, *dst;
     ucc_ee_executor_task_t    **exec_task;
     size_t                      data_size, data_displ;
-    size_t                      rdt_size = ucc_dt_size(task->alltoallv_ce.rdt);
-    size_t                      sdt_size = ucc_dt_size(task->alltoallv_ce.sdt);
     ucc_rank_t                  i, peer, psrc, pdst;
     ucc_status_t                status;
     ucc_ee_executor_task_args_t exec_args;
@@ -155,13 +160,13 @@ ucc_status_t ucc_tl_cuda_alltoallv_ce_post_copies(ucc_tl_cuda_task_t *task)
                                     sync->data[peer].ipc_event_remote, 0),
                 exit, status);
         }
-        data_size = sdt_size * task->alltoallv_ce.get_count(
-                                   task, task->alltoallv_ce.scnts, peer);
-        data_displ = sdt_size * task->alltoallv_ce.get_offset(
-                                    task, task->alltoallv_ce.sdispl, peer);
+        data_size =
+            task->alltoallv_ce.get_size(task, sync->alltoallv_ce.sbytes, peer);
+        data_displ = task->alltoallv_ce.get_offset(
+            task, sync->alltoallv_ce.sdispl_bytes, peer);
         src        = PTR_OFFSET(src, data_displ);
-        data_displ = rdt_size * task->alltoallv_ce.get_offset(
-                                    task, task->alltoallv_ce.rdispl, peer);
+        data_displ = task->alltoallv_ce.get_offset(
+            task, sync->alltoallv_ce.rdispl_bytes, peer);
         dst = PTR_OFFSET(task->alltoallv_ce.rbuf, data_displ);
 
         exec_args.task_type = UCC_EE_EXECUTOR_TASK_TYPE_COPY;
@@ -183,18 +188,18 @@ ucc_status_t ucc_tl_cuda_alltoallv_ce_post_copies(ucc_tl_cuda_task_t *task)
             pdst      = team->topo->proxies[i].dst;
             peer_sync = TASK_SYNC(task, psrc);
             // FIXME: assume the same datatype
-            data_size = sdt_size * task->alltoallv_ce.get_count(
-                                       task, peer_sync->src_cnts, pdst);
-            data_displ = sdt_size * task->alltoallv_ce.get_offset(
-                                        task, peer_sync->src_displ, pdst);
+            data_size = task->alltoallv_ce.get_size(
+                task, peer_sync->alltoallv_ce.sbytes, pdst);
+            data_displ = task->alltoallv_ce.get_offset(
+                task, peer_sync->alltoallv_ce.sdispl_bytes, pdst);
             src        = PTR_OFFSET(task->alltoallv_ce.peer_map_addr_src[psrc],
                                     peer_sync->mem_info_src.offset);
             src        = PTR_OFFSET(src, data_displ);
             peer_sync  = TASK_SYNC(task, pdst);
             dst        = PTR_OFFSET(task->alltoallv_ce.peer_map_addr_dst[pdst],
                                     peer_sync->mem_info_dst.offset);
-            data_displ = rdt_size * task->alltoallv_ce.get_offset(
-                                        task, peer_sync->dst_displ, psrc);
+            data_displ = task->alltoallv_ce.get_offset(
+                task, peer_sync->alltoallv_ce.rdispl_bytes, psrc);
             dst                 = PTR_OFFSET(dst, data_displ);
             exec_args.task_type = UCC_EE_EXECUTOR_TASK_TYPE_COPY;
             exec_args.bufs[0]   = dst;
@@ -326,18 +331,16 @@ ucc_tl_cuda_alltoallv_ce_triggered_post_setup(ucc_coll_task_t *coll_task)
 }
 
 //NOLINTNEXTLINE
-size_t ucc_tl_cuda_alltoallv_get_count(const ucc_tl_cuda_task_t *task,
-                                       ucc_count_t *cnts, ucc_rank_t block)
+size_t ucc_tl_cuda_alltoallv_get_size(const ucc_tl_cuda_task_t *task,
+                                      size_t *sizes, ucc_rank_t block)
 {
-    const ucc_coll_args_t *args = &TASK_ARGS(task);
-    return ucc_coll_args_get_count(args, cnts, block);
+    return sizes[block];
 }
 
 size_t ucc_tl_cuda_alltoallv_get_offset(const ucc_tl_cuda_task_t *task,
                                         size_t *displ, ucc_rank_t block)
 {
-    const ucc_coll_args_t *args = &TASK_ARGS(task);
-    return ucc_coll_args_get_displacement(args, displ, block);
+    return displ[block];
 }
 
 ucc_status_t ucc_tl_cuda_alltoallv_ce_init(ucc_tl_cuda_task_t *task)
@@ -358,7 +361,7 @@ ucc_status_t ucc_tl_cuda_alltoallv_ce_init(ucc_tl_cuda_task_t *task)
         return status;
     }
 
-    task->alltoallv_ce.get_count  = ucc_tl_cuda_alltoallv_get_count;
+    task->alltoallv_ce.get_size   = ucc_tl_cuda_alltoallv_get_size;
     task->alltoallv_ce.get_offset = ucc_tl_cuda_alltoallv_get_offset;
     task->alltoallv_ce.sdt        = args->src.info_v.datatype;
     task->alltoallv_ce.rdt        = args->dst.info_v.datatype;
@@ -386,6 +389,7 @@ ucc_status_t ucc_tl_cuda_alltoallv_ce_init(ucc_tl_cuda_task_t *task)
         goto exit_err;
     }
 
+    task->super.flags |= UCC_COLL_TASK_FLAG_EXECUTOR;
     task->super.post           = ucc_tl_cuda_alltoallv_ce_start;
     task->super.triggered_post = ucc_triggered_post;
     task->super.triggered_post_setup =
--- a/src/components/tl/cuda/alltoallv/alltoallv_ce.c
+++ b/src/components/tl/cuda/alltoallv/alltoallv_ce.c
@@ -11,8 +11,7 @@
 #include "utils/arch/cpu.h"
 #include "utils/arch/cuda_def.h"
 
-enum
-{
+enum {
     ALLTOALL_CE_STAGE_SYNC,  /*< Wait for free SYNC segment */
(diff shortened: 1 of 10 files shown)


kingchc:
cc @Sergei-Lebedev @vspetrov @bureddy

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,316 @@
...
+    status =
+        ucc_ec_create_event(&task->alltoall_ce.copy_done, UCC_EE_CUDA_STREAM);
+    if (ucc_unlikely(status != UCC_OK)) {
+        return status;
+    }
+
+    data_len = ucc_dt_size(args->src.info_v.datatype) *
+               ucc_coll_args_get_total_count(args, args->src.info_v.counts,
+                                             UCC_TL_TEAM_SIZE(team));
+    status = ucc_tl_cuda_mem_info_get(args->src.info_v.buffer, data_len, team,
```
Sergei-Lebedev, line 290: src and dst buffers are not always contig in alltoallv, so mem info get might fail here.
As a quick solution pls add check for UCC_COLL_ARGS_FLAG_CONTIG_SRC_BUFFER and UCC_COLL_ARGS_FLAG_CONTIG_DST_BUFFER and similarly to inplace flag return UCC_ERR_NOT_SUPPORTED

src/components/tl/cuda/tl_cuda.h
```diff
@@ -151,12 +156,12 @@ typedef struct ucc_tl_cuda_task {
...
-            void                   *peer_map_addr_src[UCC_TL_CUDA_MAX_PEERS];
-            void                   *peer_map_addr_dst[UCC_TL_CUDA_MAX_PEERS];
-            void                   *copy_done;
+            int                    stage;
+            ucc_tl_cuda_mem_info_t mem_info_src;
+            ucc_tl_cuda_mem_info_t mem_info_dst;
+            void                  *peer_map_addr_src[UCC_TL_CUDA_MAX_PEERS];
+            void                  *peer_map_addr_dst[UCC_TL_CUDA_MAX_PEERS];
+            void                  *copy_done;
         } alltoall_ce;
```
Sergei-Lebedev, line 165: minor, i understand that alltoall_ce and alltoallv_ce algorithms will probably be identical, but anyway i think it's better to have a different struct for alltoallv

Sergei-Lebedev:
@kingchc thanks for the PR! overall looks good to me, let me do some testing internally

Sergei-Lebedev:
ok to test

src/components/tl/cuda/tl_cuda.h
```diff
@@ -113,19 +114,23 @@ typedef struct ucc_tl_cuda_sync_data {
...
-    ucc_tl_cuda_mem_info_t   mem_info_dst;
-    cudaEvent_t              ipc_event_local;
-    cudaIpcEventHandle_t     ev_handle;
-    ucc_tl_cuda_sync_data_t  data[1];
+    uint32_t                seq_num[2];
+    ucc_tl_cuda_mem_info_t  mem_info_src;
+    ucc_tl_cuda_mem_info_t  mem_info_dst;
+    cudaEvent_t             ipc_event_local;
+    cudaIpcEventHandle_t    ev_handle;
+    ucc_tl_cuda_sync_data_t data[UCC_TL_CUDA_MAX_PEERS];
```
kingchc, line 129: @Sergei-Lebedev - I have changed this from `data[1];` to `data[UCC_TL_CUDA_MAX_PEERS];` to avoid index-out-of-range error when running alltoallv (when perfoming cudaStreamWaitEvent). I'm surprised it works fine for alltoall, please let me know if this change looks okay to you.

src/components/tl/cuda/tl_cuda.h
kingchc: done

src/components/tl/cuda/alltoallv/alltoallv_ce.c
kingchc: Added a check in the beginning of `ucc_tl_cuda_alltoallv_ce_init`.

src/components/tl/cuda/tl_cuda.h
kingchc: this is fixed by moving around the data structure of `ucc_tl_cuda_sync_t` as this PR adds new fields like `src_cnts`.

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,319 @@
...
+}
+
+ucc_status_t ucc_tl_cuda_alltoallv_ce_init(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
+    ucc_status_t        status;
+    size_t              data_len;
+
+    if (!(args->flags & (UCC_COLL_ARGS_FLAG_CONTIG_SRC_BUFFER |
```
vspetrov, line 278: i think the check is not correct:
1. you also need to check if args->mask & UCC_COLL_ARGS_FIELD_FLAGS is there (otherwise flags can have garbage)
2. you check would pass if either src or dst contig is set, while you need both

I would add helper macros to ucc_Coll_utils:
UCC_ARGS_SRC_CONTIG(args) (args.mask & UCC_COLL_ARGS_FIELD_FLAGS) && (args.flags & UCC_COLL_ARGS_FLAG_CONTIG_SRC_BUFFER)
similar for dst
and
for both:
UCC_COLL_ARGS_CONTIG_BUFFER(args) UCC_COLL_ARGS_SRC_CONTIG(args) && UCC_COLL_ARGS_DST_CONTIG(args)

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,319 @@
...
+ucc_status_t ucc_tl_cuda_alltoallv_setup_start(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, UCC_TL_TEAM_RANK(team));
+    ucc_status_t        status;
+
+    // copy a2av counts and displ. for proxies (if any) to access
+    if (team->topo->num_proxies > 0) {
+        ucc_coll_args_t *args = &TASK_ARGS(task);
+        // TODO: would team size match the size in args when using HIER CL?
```
vspetrov, line 41: yes, CL/HIER sets up counts/displs appropriately

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,319 @@
...
+        memcpy(sync->src_cnts, args->src.info_v.counts,
```
vspetrov, line 42: the problem with these memcpy is that counts/displs may be either 32 or 64 bits wide, so you can't use sizeof(ucc_count_t). Need to do smth like:
elem_size = UCC_COLL_ARGS_COUNT64(args) ? 8 : 4;
memcpy(..., elem_size * team_size)

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,319 @@
...
+        goto exit_err;
+    }
+
+    return UCC_OK;
+
+exit_err:
+    return status;
+}
+
+ucc_status_t ucc_tl_cuda_alltoallv_setup_test(ucc_tl_cuda_task_t *task)
```
vspetrov, line 69: this fn looks exactly same as alltoall_setup_test - can we unify them?

vspetrov (commented):
In general alltoallv start/progress looks very similar to alltoall. The only main difference (except for setup) is get_count/get_displ. Probably we could abstract them (adding fn pointer to task) and then re-use the same code for both algs. Similar thing is done by @Sergei-Lebedev in #437 (e.g. tl/cuda/allgatherv/allgatherv_ring.c, in tl_cuda_allgatherv_ring_init, look for ucc_tl_cuda_allgatherv_get_count)

kingchc:
@Sergei-Lebedev @vspetrov @bureddy - This PR is ready for review again, major changes.
- rebased on top of latest master
- unify alltoall and alltoallv similar to #427 
- addressed above comments

Please let me know if you have any further comments

Sergei-Lebedev:
CI error
```
/opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c: In function 'ucc_tl_cuda_alltoallv_ce_post_copies':
/opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c:122:33: error: unused variable 'args' [-Werror=unused-variable]
02:14:31       ucc_coll_args_t            *args = &TASK_ARGS(task);```

**Comment by kingchc** [comment]
> CI error
> 
> ```
> /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c: In function 'ucc_tl_cuda_alltoallv_ce_post_copies':
> /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c:122:33: error: unused variable 'args' [-Werror=unused-variable]
> 02:14:31       ucc_coll_args_t            *args = &TASK_ARGS(task);```
> ```

Thanks for catching this. Fix pushed.

File: src/components/tl/cuda/alltoallv/alltoallv_ce.c [hunk 8]
```diff
@@ -0,0 +1,402 @@
...
+            src = task->alltoallv_ce.sbuf;
+        } else {
+            src = PTR_OFFSET(task->alltoallv_ce.peer_map_addr_src[peer],
+                             peer_sync->mem_info_src.offset);
+            CUDA_CHECK_GOTO(
+                cudaStreamWaitEvent(team->stream,
+                                    sync->data[peer].ipc_event_remote, 0),
+                exit, status);
+        }
+        data_size = sdt_size * task->alltoallv_ce.get_count(
```
Sergei-Lebedev, line 158: for alltoall, get_count function returns number of elements in dst buffer which might be different from number of elements in src buffer, so need to multiple by rdt_size

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,402 @@
...
+                                   task, task->alltoallv_ce.scnts, peer);
+        data_displ = sdt_size * task->alltoallv_ce.get_offset(
```
Sergei-Lebedev, line 160: since you calculate displacement in remote buffer you need to use offset from shared memory i.e.
```
data_displ = sdt_size * task->alltoallv_ce.get_offset(task, peer_sync->src_displ, rank);
```

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,402 @@
...
+        tl_error(UCC_TL_TEAM_LIB(team), "Do not support non-contiguous buffer");
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    status =
+        ucc_ec_create_event(&task->alltoallv_ce.copy_done, UCC_EE_CUDA_STREAM);
+    if (ucc_unlikely(status != UCC_OK)) {
+        return status;
+    }
+
```
Sergei-Lebedev, line 360: need to add task->super.flags |= UCC_COLL_TASK_FLAG_EXECUTOR to tell upper layer to create executor for regular post

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,402 @@
...
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, UCC_TL_TEAM_RANK(team));
+    ucc_status_t        status;
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
+
+    // for A2av: copy counts and displ. for proxies (if any) to access
+    if (team->topo->num_proxies > 0 &&
+        UCC_COLL_TYPE_ALLTOALLV == args->coll_type) {
+        size_t elem_size = UCC_COLL_ARGS_COUNT64(args) ? 8 : 4;
+        memcpy(sync->src_cnts, args->src.info_v.counts,
```
Sergei-Lebedev, line 44: it would be easier if we save offsets and counts in bytes rather than in elements, it will solver sdt_size and rdt_size problem and also enable support for the case when source datatype is not the same as recv datatype

Sergei-Lebedev (changes requested):

src/components/tl/cuda/alltoallv/alltoallv_ce.c
kingchc: hmm...in the case of a remote process uses different datatype, I think both rdt_size and sdt_size could be wrong. I will take your suggestion below to save size and offset in bytes to solve this problem.

kingchc: is it correct? If the src is rank itself (i.e., `peer==rank`), then it should be `sdt_size`. If src is from remote GPU, it should be remote's `sdt_size` which might not be the same as the local `sdt_size`/`rdt_size`, if I understand the semantic correctly. I like the idea you mentioned below of saving number of bytes instead elements to make things easier.

Sergei-Lebedev: right, let's just save number of bytes in shm segment 😊

kingchc:
Addressed Sergey's comments. Major changes:
- rebased
- cache A2Av sizes and displacement **in bytes** on SHM, e.g., `get_count`->`get_size`, to make remote access easier without explicitly handling about src/dst datatype

kingchc:
There seems to be some tests failed, but I could not see the details. @Sergei-Lebedev - plz let me know if you see any runtime issue, and I can fix them.

Sergei-Lebedev:
> There seems to be some tests failed, but I could not see the details. @Sergei-Lebedev - plz let me know if you see any runtime issue, and I can fix them.

it's compilation issue
```
00:27:54  /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c: In function 'ucc_tl_cuda_alltoallv_setup_start':
00:27:54  /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c:66:5: error: implicit declaration of function 'ucc_memory_bus_fence'; did you mean 'ucc_memory_cpu_fence'? [-Werror=implicit-function-declaration]
00:27:54       ucc_memory_bus_fence();
```
if https://github.com/openucx/ucc/pull/442 ucc_memory_bus_fence was removed, now it should be ucc_memory_cpu_store_fence();

kingchc:
> > There seems to be some tests failed, but I could not see the details. @Sergei-Lebedev - plz let me know if you see any runtime issue, and I can fix them.
> 
> it's compilation issue
> 
> ```
> 00:27:54  /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c: In function 'ucc_tl_cuda_alltoallv_setup_start':
> 00:27:54  /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c:66:5: error: implicit declaration of function 'ucc_memory_bus_fence'; did you mean 'ucc_memory_cpu_fence'? [-Werror=implicit-function-declaration]
> 00:27:54       ucc_memory_bus_fence();
> ```
> 
> if #442 ucc_memory_bus_fence was removed, now it should be ucc_memory_cpu_store_fence();

Thanks for pointing it out! looks like I tested a wrong version locally and did not catch this. Fixes pushed.

kingchc:
> > There seems to be some tests failed, but I could not see the details. @Sergei-Lebedev - plz let me know if you see any runtime issue, and I can fix them.
> 
> it's compilation issue
> 
> ```
> 00:27:54  /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c: In function 'ucc_tl_cuda_alltoallv_setup_start':
> 00:27:54  /opt/nvidia/src/ucc/src/components/tl/cuda/alltoallv/alltoallv_ce.c:66:5: error: implicit declaration of function 'ucc_memory_bus_fence'; did you mean 'ucc_memory_cpu_fence'? [-Werror=implicit-function-declaration]
> 00:27:54       ucc_memory_bus_fence();
> ```
> 
> if #442 ucc_memory_bus_fence was removed, now it should be ucc_memory_cpu_store_fence();

Thanks for pointing it out! looks like I tested a wrong version locally and did not catch this.

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,405 @@
...
+{
+    ucc_tl_cuda_task_t *task = ucc_derived_of(coll_task, ucc_tl_cuda_task_t);
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_status_t        status;
+    int                 i;
+
+    switch (task->alltoallv_ce.stage) {
+    case ALLTOALL_CE_STAGE_SYNC:
+        if (ucc_tl_cuda_get_sync(task) != UCC_OK) {
+            task->super.super.status = UCC_INPROGRESS;
```
Sergei-Lebedev, line 233: need to replace all task->super.super.status = something with task->super.status = something, because of this change https://github.com/openucx/ucc/pull/441

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,405 @@
...
+    if (task->alltoallv_ce.stage != ALLTOALL_CE_STAGE_POST_COPIES) {
+        task->alltoallv_ce.stage = ALLTOALL_CE_STAGE_SYNC;
+    }
+
+    ucc_tl_cuda_alltoallv_ce_progress(coll_task);
+    if (task->super.super.status == UCC_INPROGRESS) {
+        ucc_progress_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+        return UCC_OK;
+    }
+    return ucc_task_complete(coll_task);
```
Sergei-Lebedev, line 298: pls replace with, 
```suggestion
return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
```
ucc_progress_queue_enqueue will do all the necessary setup for task,

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,405 @@
...
+                             peer_sync->mem_info_src.offset);
+            CUDA_CHECK_GOTO(
+                cudaStreamWaitEvent(team->stream,
+                                    sync->data[peer].ipc_event_remote, 0),
+                exit, status);
+        }
+        data_size =
+            task->alltoallv_ce.get_size(task, sync->alltoallv_ce.sbytes, peer);
+        data_displ = task->alltoallv_ce.get_offset(
+            task, sync->alltoallv_ce.sdispl_bytes, peer);
```
Sergei-Lebedev, line 166: data_displ is not correct
```suggestion
        data_size =
            task->alltoallv_ce.get_size(task, peer_sync->alltoallv_ce.sbytes, rank);
        data_displ = task->alltoallv_ce.get_offset(
            task, peer_sync->alltoallv_ce.sdispl_bytes, rank);
```

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,405 @@
...
+                &task->alltoallv_ce.exec_task[task->alltoallv_ce.num_posted];
+            status = ucc_ee_executor_task_post(exec, &exec_args, exec_task);
+            if (ucc_unlikely(status != UCC_OK)) {
+                goto exit;
+            }
+            task->alltoallv_ce.num_posted++;
+        }
+    }
+    status = ucc_ec_event_post(team->stream, task->alltoallv_ce.copy_done,
+                               UCC_EE_CUDA_STREAM);
```
kingchc, line 218: @Sergei-Lebedev - I just realized this will result in calling `cudaEventRecord` and may cause deadlock for multi-thread case (e.g., PyTorch). It seems to me we don't really need this now since we introduce the executor. 

I plan to remove this and all `task->alltoallv_ce.copy_done`, please let me know if you think otherwise.

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -0,0 +1,389 @@
...
+
+ucc_status_t ucc_tl_cuda_alltoallv_ce_init(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
+    ucc_status_t        status;
+    size_t              data_len;
+
+    if (!UCC_COLL_ARGS_CONTIG_BUFFER(args)) {
+        tl_error(UCC_TL_TEAM_LIB(team), "Do not support non-contiguous buffer");
```
Sergei-Lebedev, line 344: I think it's better to use tl_debug here since we will fallback to other TL. I see a lot of errors in internal tests log but test passes anyway
```
[1649680880.572873]     alltoallv_ce.c:344  TL_CUDA ERROR Do not support non-contiguous buffer
```

Sergei-Lebedev (approved):
PR looks good to me, thanks @kingchc!

src/components/tl/cuda/alltoallv/alltoallv_ce.c
kingchc: I'm concerned that users might make wrong assumptions if we fall back silently. Perhaps we can just print the error the first time?

vspetrov: @kingchc the behavior we are trying to keep in UCC is: be functional as much as possible (means, if smth is not available/not implement try to fallback to other components) and be "silent" by default unless fatal error (like NO_MEM). User can always check with proper LOG_LEVEL setting

kingchc:
rebased and addressed review comments. (cc @vspetrov @Sergei-Lebedev)

PR #434, opened 2022-03-01
UTIL: cpu model and vendor detection

shimmybalsam:
## What
Adding CPU model & vendor detection

## Why ?
Needed for TL SHM implementation

## How ?
Copy of needed functions from UCS

### DIFF
--- a/src/utils/arch/aarch64/cpu.h
+++ b/src/utils/arch/aarch64/cpu.h
@@ -35,4 +35,44 @@
 #define ucc_memory_cpu_store_fence()  ucc_aarch64_dmb(ishst)
 #define ucc_memory_cpu_load_fence()   ucc_aarch64_dmb(ishld)
 
+typedef struct ucc_aarch64_cpuid {
+    int       implementer;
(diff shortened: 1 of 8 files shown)


src/utils/arch/aarch64/cpu.h
```diff
@@ -35,4 +35,44 @@
...
+    int       revision;
+} ucc_aarch64_cpuid_t;
+
+static inline void ucc_aarch64_cpuid(ucc_aarch64_cpuid_t *cpuid);
+{
+    static ucc_aarch64_cpuid_t cached_cpuid;
+    static int initialized = 0;
+
+    if (!initialized) {
+        ucc_aarch64_cpuid_from_proc(&cached_cpuid);
```
Sergei-Lebedev, line 52: this functions is undefined, pls test on arm and ppc to make sure it works

shimmybalsam: added and tested.

Sergei-Lebedev (approved):

PR #441, opened 2022-03-21
CORE: add ucc task internal status

Sergei-Lebedev:
## What
Fix mt race condition with cb.

## How ?
Added ucc task internal status. Each TL/CL use internal status to represent collective state, while user visible status is updated only after all work is done and request ownership is moved to user code.

Also some other small fixes:
1. Double free in onesided alltoall
2. Persistent barrier task reset
3. Incorrect TL UCP status after ucc_collective_initialized

### DIFF
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -1981,6 +1981,9 @@ static inline ucc_status_t ucc_collective_test(ucc_coll_req_h request)
  *
  *  @ref ucc_collective_finalize operation releases all resources
  *  associated with the collective operation represented by the request handle.
+ *  In UCC_THREAD_MULTIPLE mode, the user is responsible for ensuring that
+ *  @ref ucc_collective_finalize is called after the status is UCC_OK and after
+ *  completing the execution of any callback registered with @ref ucc_coll_args_t.
  *
  *  @endparblock
  *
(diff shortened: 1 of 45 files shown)


Sergei-Lebedev:
bot:retest

manjugv:
@Sergei-Lebedev Will you create a separate port for v1.0 branch?

vspetrov (approved):

manjugv (approved):

bureddy (approved):

src/ucc/api/ucc.h
```diff
@@ -1981,6 +1981,9 @@ static inline ucc_status_t ucc_collective_test(ucc_coll_req_h request)
  *
  *  @ref ucc_collective_finalize operation releases all resources
  *  associated with the collective operation represented by the request handle.
+ *  In UCC_THREAD_MULTIPLE mode, the user is responsible for ensuring that
+ *  @ref ucc_collective_finalize is called after the status is UCC_OK and after
+ *  completing the execution of any callback registered with @ref ucc_coll_args_t.
  *
```
bureddy, line 1987: how about returning status from callback to know if finalize() is called request is released so that we dont set externel status inside UCC.  so this way we can set external status after callback and i think we can remove this restriction

PR #446, opened 2022-03-24
CL/HIER: enable a2a(v)

kingchc:
## What
Fully enable CL HIER Alltoall and Alltoallv

## Why ?
Enable HIER Alltoall(v) such as TL/UCP and TL/CUDA

## How ?
Enabled in the code and added triggered_post_setup function


kingchc:
cc @vspetrov @Sergei-Lebedev

PR #447, opened 2022-03-28
TL/UCP: correct RS knomial scratch size

vspetrov:
## What
Fixes memory corruption caused by tl_ucp_reduce_scatter_knomial algorithm. It can be reproduced when SRA allreduce is enabled for very small msg sizes (e.g. 4 bytes AR) - this is not default. In those cases dst buffer can not be used as scratch (amount of recv data can overrun it)

### DIFF
--- a/src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
+++ b/src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
@@ -190,15 +190,13 @@ ucc_status_t ucc_tl_ucp_reduce_scatter_knomial_start(ucc_coll_task_t *coll_task)
     ucc_tl_ucp_team_t *team = TASK_TEAM(task);
     ucc_rank_t         rank = UCC_TL_TEAM_RANK(team);
     ucc_rank_t         size = UCC_TL_TEAM_SIZE(team);
-    uint8_t            node_type;
 
     UCC_TL_UCP_PROFILE_REQUEST_EVENT(coll_task, "ucp_reduce_scatter_kn_start",
                                      0);
     ucc_tl_ucp_task_reset(task, UCC_INPROGRESS);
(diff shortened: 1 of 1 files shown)


src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
```diff
@@ -248,11 +247,22 @@ ucc_status_t ucc_tl_ucp_reduce_scatter_knomial_init_r(
...
+    } else {
+        step_radix =
+            ucc_sra_kn_compute_step_radix(rank, size,
+                                          &task->reduce_scatter_kn.p);
+        max_recv_size = ucc_sra_kn_compute_seg_size(count, step_radix, 0) *
+            step_radix * dt_size;
+    }
     if (UCC_IS_INPLACE(coll_args->args) ||
-        (KN_NODE_PROXY == task->reduce_scatter_kn.p.node_type)) {
+        (KN_NODE_PROXY == task->reduce_scatter_kn.p.node_type) ||
```
Sergei-Lebedev, line 262: just two questions
1. Why condition is (UCC_IS_INPLACE or node_type == PROXY or max_recv_size > data_size), should it be 
((UCC_IS_INPLACE and node_type == PROXY) or max_recv_size > data_size)
2. EXTRA rank doesn't need any additional space right? Maybe add it to condition

vspetrov: Regarding (2) - argee. Extra never needs scratch. Will fix.
Regarding (1) - no. PROXY always needs scratch. If UCC_IN_PLACE both PROXY and BASE need scratch. If rbuf is too small (max_recv_size > data_size) also always need bigger scratch. In other words: we can avoid scratch only if: (NON_INPLACE && max_recv_size <= data_isze && NOT_PROXY) || EXTRA.

Sergei-Lebedev (approved):

bureddy (approved):

PR #449, opened 2022-03-29
TL/UCP: reduce_kn avg pre op fix

shimmybalsam:
## What
Fixing reduce knomial avg pre op.

## Why ?
Correctness.

## How ?
Adding pre op avg division, in the case where single process doesnt take part in first iteration reduction.

### DIFF
--- a/src/components/tl/ucp/reduce/reduce_knomial.c
+++ b/src/components/tl/ucp/reduce/reduce_knomial.c
@@ -101,6 +101,21 @@ UCC_REDUCE_KN_PHASE_MULTI:
                         task->super.status = status;
                         return;
                     }
+                } else if (args->op == UCC_OP_AVG && avg_pre_op &&
+                           task->reduce_kn.dist == 1) {
+                    ucc_mc_memcpy(rbuf, args->src.info.buffer, data_size,
+                                  mtype, args->src.info.mem_type);
+                    status = ucc_dt_reduce_multi_alpha(args->src.info.buffer,
+                        rbuf, rbuf, 1, count, data_size, dt, UCC_OP_PROD,
+                        (double)1 /
+                        (double)(UCC_TL_TEAM_SIZE(TASK_TEAM(task)) * 2),
+                        mtype, args);
+                    if (ucc_unlikely(UCC_OK != status)) {
+                        tl_error(UCC_TASK_LIB(task),
+                                 "failed to perform dt reduction");
+                        task->super.super.status = status;
+                        return status;
+                    }
                 }
             } else {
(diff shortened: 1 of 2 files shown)


src/components/tl/ucp/reduce/reduce_knomial.c
```diff
@@ -101,6 +101,21 @@ ucc_status_t ucc_tl_ucp_reduce_knomial_progress(ucc_coll_task_t *coll_task)
                         task->super.super.status = status;
                         return status;
                     }
+                } else if (args->op == UCC_OP_AVG && avg_pre_op &&
+                           task->reduce_kn.dist == 1) {
+                    ucc_mc_memcpy(rbuf, args->src.info.buffer, data_size,
+                                  mtype, args->src.info.mem_type);
+                    status = ucc_dt_reduce_multi_alpha(args->src.info.buffer,
```
vspetrov, line 108: i think we can avoid memcpy by calling reduce_multi(src, src, scratch)

vspetrov: also can we move this to "init" function?

src/components/tl/ucp/reduce/reduce_knomial.c
```diff
@@ -141,7 +147,7 @@ ucc_status_t ucc_tl_ucp_reduce_knomial_start(ucc_coll_task_t *coll_task)
         args->src.info.buffer = args->dst.info.buffer;
     }
 
-    if (isleaf) {
+    if (isleaf && !self_avg) {
```
vspetrov, line 150: maybe we move that also to init, this way we just set the scratch once.

vspetrov (approved):

src/components/tl/ucp/reduce/reduce_knomial.c
Sergei-Lebedev: > i think we can avoid memcpy by calling reduce_multi(src, src, scratch)

@vspetrov @shimmybalsam 
we should not move it to collective init, otherwise persistent mode won't work. Let's better move it to start function

src/components/tl/ucp/reduce/reduce_knomial.c
shimmybalsam: in the condition above dst is assigned to src and here src is assigned to scratch, so IMHO it is more readable to keep this here.

src/components/tl/ucp/reduce/reduce_knomial.c
shimmybalsam: @Sergei-Lebedev moved to start

PR #457, opened 2022-04-05
TEST: mpi barrier progress

vspetrov:
Need to call mpi_progress otherwise a hang is possible in MPI_Ssend when root will be stuck in ucc collective.

### DIFF
--- a/test/mpi/test_barrier.cc
+++ b/test/mpi/test_barrier.cc
@@ -73,9 +73,11 @@ void TestBarrier::run()
             if (0 == rank && !completed) {
                 MPI_Test(&rreq, &completed, MPI_STATUS_IGNORE);
             }
+            mpi_progress();
         } while(UCC_OK != status);
-        if  (0 == rank && !completed) {
-            MPI_Wait(&rreq, MPI_STATUS_IGNORE);
+        while (0 == rank && !completed) {
+            MPI_Test(&rreq, &completed, MPI_STATUS_IGNORE);
+            mpi_progress();
         }
         if (i < size - 1) {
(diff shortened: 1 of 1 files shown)


test/mpi/test_barrier.cc
```diff
@@ -73,9 +73,11 @@ void TestBarrier::run()
             if (0 == rank && !completed) {
                 MPI_Test(&rreq, &completed, MPI_STATUS_IGNORE);
             }
+            mpi_progress();
         } while(UCC_OK != status);
-        if  (0 == rank && !completed) {
-            MPI_Wait(&rreq, MPI_STATUS_IGNORE);
+        while (0 == rank && !completed) {
+            MPI_Test(&rreq, &completed, MPI_STATUS_IGNORE);
```
Sergei-Lebedev, line 79: need to progress mpi for rank i as well?

vspetrov: i think we have enough progress. all other ranks have no place to block and they call mpi_progress from line 76 otherwise

manjugv (approved):

test/mpi/test_barrier.cc
Sergei-Lebedev: ok, I just thought that for the last barrier we need to progress send rank because need to progress ucp_worker

Sergei-Lebedev (approved):

test/mpi/test_barrier.cc
vspetrov: the rank that returns from Ssend goes either to the while loop of collective_test (with mpi_progress at line 76) or it would go to next iteration, or to MPI_Barrier which also progresses MPI.

PR #459, opened 2022-04-05
BUILD: add cuda arch flag

Sergei-Lebedev:
## What
Adding build flag to set CUDA virtual arch and PTX generation

## Why ?
- compile for GPU arch not enabled by default
- reduce binary size
- reduce compilation time

## How ?
Example cmd line
```
../configure --with-cuda --with-nvcc-gencode="-gencode=arch=compute_35,code=sm_35"
```

### DIFF
--- a/config/m4/cuda.m4
+++ b/config/m4/cuda.m4
@@ -22,6 +22,11 @@ AS_IF([test "x$cuda_checked" != "xyes"],
     AC_ARG_WITH([cuda],
                 [AS_HELP_STRING([--with-cuda=(DIR)], [Enable the use of CUDA (default is guess).])],
                 [], [with_cuda=guess])
+    AC_ARG_WITH([nvcc-gencode],
+                [AS_HELP_STRING([--with-nvcc-gencode=arch,code],
+                                [Defines target GPU architecture,
+                                 see nvcc -gencode option for details])],
+                [], [with_nvcc_gencode=default])
     AS_IF([test "x$with_cuda" = "xno"],
         [
          cuda_happy=no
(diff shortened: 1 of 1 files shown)


config/m4/cuda.m4
```diff
@@ -96,9 +101,12 @@ AS_IF([test "x$cuda_checked" != "xyes"],
                [NVCC_CFLAGS="$NVCC_CFLAGS -O0 -g"],
                [NVCC_CFLAGS="$NVCC_CFLAGS -O3 -g -DNDEBUG"])
          AS_IF([test "x$cuda_happy" = "xyes"],
-               [AS_IF([test $CUDA_MAJOR_VERSION -eq 11],
-                      [NVCC_ARCH="${ARCH8} ${ARCH9} ${ARCH10} ${ARCH11}"])
-                AC_SUBST([NVCC_ARCH], ["$NVCC_ARCH"])])
+               [AS_IF([test "x$with_nvcc_gencode" = "xdefault"],
```
vspetrov, line 104: i would simplify user experience a bit:
we can add a helper fn on top of this file:
```
  AC_DEFUN([CUDA_ARCH_LIST], [
      arch_config_str=$1
      NVCC_ARCH=""
      for arch in $(echo ${arch_config_str} | tr "," " "); do
           NVCC_ARCH="${NVCC_ARCH} -gencode=arch=compute_${arch},code=sm_${arch}"
      done
   ])
```
then if with_nvcc_gencode != default:
CUDA_ARCH_LIST([$with_nvcc_gencode]) 
which will set NVCC_ARCH.

in that case user can set --with-nvcc-gencode=50,60,70
wdyt?

manjugv:
@Sergei-Lebedev How do we ensure we support PASCAL and above, not something below? IIRC, with CUDA 11, there is no support for MAXWELL, right?

config/m4/cuda.m4
Sergei-Lebedev: i thought about it, the drawback here is that by setting ucc configure arg you explicitly compile for a particular GPU i.e. generating particular PTX and SAAS. What if user wants set only virtual arch?

Sergei-Lebedev:
> @Sergei-Lebedev How do we ensure we support PASCAL and above, not something below? IIRC, with CUDA 11, there is no support for MAXWELL, right?

If we use something architecture specific we add conditional compilation, for example here to support fp16 https://github.com/openucx/ucc/blob/f32f78ee53523b66ba60c23099c02c0d2a120d02/src/components/mc/cuda/kernel/mc_cuda_reduce_ops.h#L28

In CUDA 11 even Kepler is supported, it's deprecated but supported. That's why we don't compile for Kepler by default, but if user wants it's now possible with the flag added in this PR.

config/m4/cuda.m4
Sergei-Lebedev: yes, i think so, if you want to build ucc binary today but support future GPU architectures you will need to add ptx

vspetrov (approved):

bureddy (approved):

PR #462, opened 2022-04-07
EC/CUDA: use multiple streams

Sergei-Lebedev:
## What
Use multiple streams in EC CUDA executor.

## Why ?
Algorithms in TL CUDA use executor for communication. While executor api supports nonblocking operations executor tasks are synchronized because interruptible executor submits them to the same stream. This PR adds multiple streams to EC CUDA and executor assigns tasks to different streams in round robin manner.

### DIFF
--- a/src/components/ec/cuda/ec_cuda.c
+++ b/src/components/ec/cuda/ec_cuda.c
@@ -216,8 +216,9 @@ static ucc_status_t ucc_ec_cuda_init(const ucc_ec_params_t *ec_params)
     cudaError_t           cuda_st;
     const char           *cu_err_st_str;
 
-    ucc_ec_cuda.stream             = NULL;
-    ucc_ec_cuda.stream_initialized = 0;
+    ucc_ec_cuda.stream                   = NULL;
+    ucc_ec_cuda.stream_initialized       = 0;
+    ucc_ec_cuda.exec_streams_initialized = 0;
     ucc_strncpy_safe(ucc_ec_cuda.super.config->log_component.name,
                      ucc_ec_cuda.super.super.name,
                      sizeof(ucc_ec_cuda.super.config->log_component.name));
@@ -442,15 +443,16 @@ static ucc_status_t ucc_ec_cuda_finalize()
 {
     int i;
 
-    if (ucc_ec_cuda.stream != NULL) {
-        CUDA_CHECK(cudaStreamDestroy(ucc_ec_cuda.stream));
-        ucc_ec_cuda.stream = NULL;
+    if (ucc_ec_cuda.stream_initialized) {
+        CUDA_FUNC(cudaStreamDestroy(ucc_ec_cuda.stream));
+        ucc_ec_cuda.stream_initialized = 0;
     }
 
-    if (ucc_ec_cuda.exec_streams[0] != NULL) {
+    if (ucc_ec_cuda.exec_streams_initialized) {
         for (i = 0; i < EC_CUDA_CONFIG->exec_num_streams; i++) {
-            cudaStreamDestroy(ucc_ec_cuda.exec_streams[i]);
+            CUDA_FUNC(cudaStreamDestroy(ucc_ec_cuda.exec_streams[i]));
         }
+        ucc_ec_cuda.exec_streams_initialized = 0;
     }
 
     ucc_mpool_cleanup(&ucc_ec_cuda.events, 1);
--- a/src/components/ec/cuda/ec_cuda.c
+++ b/src/components/ec/cuda/ec_cuda.c
@@ -63,6 +63,11 @@ static ucc_config_field_t ucc_ec_cuda_config_table[] = {
      ucc_offsetof(ucc_ec_cuda_config_t, exec_max_tasks),
      UCC_CONFIG_TYPE_ULUNITS},
 
+    {"EXEC_NUM_STREAMS", "16",
+     "Number of streams used by interruptible executor",
+     ucc_offsetof(ucc_ec_cuda_config_t, exec_num_streams),
+     UCC_CONFIG_TYPE_ULUNITS},
+
     {NULL}
 
 };
@@ -224,6 +229,13 @@ static ucc_status_t ucc_ec_cuda_init(const ucc_ec_params_t *ec_params)
     }
     CUDA_CHECK(cudaGetDevice(&device));
     /*create event pool */
+    ucc_ec_cuda.exec_streams = ucc_calloc(cfg->exec_num_streams,
+                                          sizeof(cudaStream_t),
+                                          "ec cuda streams");
+    if (!ucc_ec_cuda.exec_streams) {
+        ec_error(&ucc_ec_cuda.super, "failed to allocate streams array");
+        return UCC_ERR_NO_MEMORY;
+    }
     status = ucc_mpool_init(&ucc_ec_cuda.events, 0, sizeof(ucc_ec_cuda_event_t),
                             0, UCC_CACHE_LINE_SIZE, 16, UINT_MAX,
                             &ucc_ec_cuda_event_mpool_ops, UCC_THREAD_MULTIPLE,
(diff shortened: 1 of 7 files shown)


src/components/ec/cuda/ec_cuda_executor_interruptible.c
```diff
@@ -0,0 +1,157 @@
...
+#include "ec_cuda_executor.h"
+#include "components/mc/ucc_mc.h"
+
+ucc_status_t ucc_cuda_executor_interruptible_get_stream(cudaStream_t *stream)
+{
+    static int last_used = 0;
+    cudaError_t cuda_st;
+    int i, j;
+
+    ucc_spin_lock(&ucc_ec_cuda.init_spinlock);
```
vspetrov, line 16: why we need lock always? do we need to check the thread mode?
Can we implement this "get_stream" using ucc_mpool_get? E.g. we can allocate a pool of ptrs with "init" function that actually creates a stream. Then it would be called when required. wdyt?

Sergei-Lebedev: I intentionally didn't use mpool because the idea here is to start reuse streams when we reach exec_num_streams. Alternatively we can use mpool, but then in some cases it will grow allocating a lot of streams, I wanted to avoid this. Do you think it's better not to limit number of streams?
cc @bureddy

vspetrov: right, this makes sense. the question regarding locking still remains: we should probably use atomic_fadd for last_used in MT case and just regular "++" in ST. ?

bureddy: with mpool api have a parameter to set max limit.

vspetrov: @bureddy yes it has "max_limit" but not clear how to re-use round robin way. we don't want a stream being taken from pool until the op is done. we want them to be used round-robin no matter how many outstanding ops are there. E.g., with 8 ops being posted and 4 streams max, each would get 2 ops. If we use mpool, then a stream taken from there is not available until it is returned back.

src/components/ec/cuda/ec_cuda_executor_interruptible.c
```diff
@@ -0,0 +1,157 @@
...
+    if (ucc_unlikely(ucc_ec_cuda.exec_streams[0] == NULL)) {
+        for(i = 0; i < EC_CUDA_CONFIG->exec_num_streams; i++) {
+            cuda_st = cudaStreamCreateWithFlags(&ucc_ec_cuda.exec_streams[i],
```
bureddy, line 19: will it be optimal to create steam on demand at line:35 on first use?
I think cuda stream is a costly object. if I remember it consumes ~100 MB GPU memory.  can we check this again?

src/components/ec/cuda/ec_cuda_executor_interruptible.c
Sergei-Lebedev: @vspetrov, added atomic add instead of spinlock for main path. For stream init spinlock is still used, i think it's ok because we do it only once and lock/unlock overhead is small compared to stream create, also all ec/mc components initialized with UCC_THREAD_MULTIPLE and I don't think this will be fixed soon

src/components/ec/cuda/ec_cuda_executor_interruptible.c
Sergei-Lebedev: I don't think it makes sense, anyway after 16 ops all streams will be initialized because we use them one by one. Similar initialization scheme is used in cuda_ipc in UCX https://github.com/openucx/ucx/blob/b054f883c9f8912444f34a48873429c9041c4f03/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c#L451

src/components/ec/cuda/ec_cuda.c
```diff
@@ -427,13 +439,24 @@ ucc_status_t ucc_ec_cuda_event_test(void *event)
...
+    int i;
+
     if (ucc_ec_cuda.stream != NULL) {
         CUDA_CHECK(cudaStreamDestroy(ucc_ec_cuda.stream));
         ucc_ec_cuda.stream = NULL;
     }
+
+    if (ucc_ec_cuda.exec_streams[0] != NULL) {
+        for (i = 0; i < EC_CUDA_CONFIG->exec_num_streams; i++) {
+            cudaStreamDestroy(ucc_ec_cuda.exec_streams[i]);
```
vspetrov, line 451: CUDA_CHECK

src/components/ec/cuda/ec_cuda_executor_interruptible.c
```diff
@@ -0,0 +1,164 @@
...
+
+        for(i = 0; i < EC_CUDA_CONFIG->exec_num_streams; i++) {
+            cuda_st = cudaStreamCreateWithFlags(&ucc_ec_cuda.exec_streams[i],
+                                                cudaStreamNonBlocking);
+            if (ucc_unlikely(cuda_st != cudaSuccess)) {
+                ec_error(&ucc_ec_cuda.super,
+                         "failed to allocate cuda stream %d(%s)",
+                         cuda_st, cudaGetErrorString(cuda_st));
+                for (j = 0; j < i; j++) {
+                    cudaStreamDestroy(ucc_ec_cuda.exec_streams[j]);
```
vspetrov, line 32: CUDA_CHECK

src/components/ec/cuda/ec_cuda_executor_interruptible.c
```diff
@@ -0,0 +1,164 @@
...
+    uint32_t        id;
+
+    if (ucc_unlikely(ucc_ec_cuda.exec_streams[0] == NULL)) {
+        ucc_spin_lock(&ucc_ec_cuda.init_spinlock);
+        if (ucc_ec_cuda.exec_streams[0] != NULL) {
+            goto unlock;
+        }
+
+        for(i = 0; i < EC_CUDA_CONFIG->exec_num_streams; i++) {
+            cuda_st = cudaStreamCreateWithFlags(&ucc_ec_cuda.exec_streams[i],
```
vspetrov, line 25: @Sergei-Lebedev i've discovered a race in MC_INIT in master (Fix: https://github.com/openucx/ucc/pull/468/commits/fd1b5529ece38de433b039caf029be5a96aa01a9). Happens in MT case. Apparently, internal CUDA implementation only guarantees that the stream ptr can be safely re-used once the corresponding cudaStreamCreateWithFlags returns. The race is: one thread gets the lock and calls streamCreate: inside the call the shared stream ptr gets non Null value, but stream is not entirely initialized yet. At the same time another thread seeys stream != NULL ang goes to cudaMemcpy -> this causes an error from cuda: [1650316036.432440] [swx-clx02:15445:0]         mc_cuda.c:309  cuda mc ERROR failed to launch cudaMemcpyAsync,  dst 0x7f441bf00000, src 0x7f441c001240, len 4 cuda error 400(invalid resource handle)

The fix is to use local variable and atomic_cswap. In the case of array i think you can do the same but check the last ptr of the array for NULL (or loop backward)

vspetrov (changes requested):

src/components/ec/cuda/ec_cuda_executor_interruptible.c
Sergei-Lebedev: good catch! fixed

src/components/ec/cuda/ec_cuda_executor_interruptible.c
```diff
@@ -0,0 +1,168 @@
...
+    if (ucc_unlikely(ucc_ec_cuda.exec_streams[num_streams - 1] == NULL)) {
+        ucc_spin_lock(&ucc_ec_cuda.init_spinlock);
+        if (ucc_ec_cuda.exec_streams[num_streams - 1] != NULL) {
+            goto unlock;
+        }
+
+        for(i = 0; i < num_streams; i++) {
+            cuda_st = cudaStreamCreateWithFlags(&tmp_stream,
+                                                cudaStreamNonBlocking);
+            ucs_atomic_cswap64((uint64_t*)(uintptr_t)&ucc_ec_cuda.exec_streams[i],
```
vspetrov, line 29: @Sergei-Lebedev plz note we changed the fix in master. Per Devendars suggestion i've removed atomic cswap in favor of lobal ucc_ec_cuda.initialized in variable. This is done to make no assumption that cudaStram_t is 8 bytes (ptr). plz change accordingly

vspetrov (changes requested):

src/components/ec/cuda/ec_cuda_executor_interruptible.c
Sergei-Lebedev: done

vspetrov (approved):

bureddy (approved):

PR #468, opened 2022-04-13
Fix multi-threading of ucc_test_mpi

vspetrov:
## What
Fixes threading in ucc_test_mpi (incorrect thread_mode parsing/init order)
Verbose option (less output by default)
Run MT tests in jenkins

### DIFF
--- a/.ci/scripts/run_tests_ucc_mpi.sh
+++ b/.ci/scripts/run_tests_ucc_mpi.sh
@@ -24,99 +24,44 @@ export MASTER_ADDR=${HEAD_NODE}
 
 NP=$(wc --lines "$HOSTFILE" | awk '{print $1}')
 
-# shellcheck disable=SC2086
-mpirun \
-    -np $NP \
-    --hostfile ${HOSTFILE} \
-    --map-by node \
-    --allow-run-as-root \
-    --mca plm_rsh_args '-p 12345' \
-    -x PATH \
-    -x LD_LIBRARY_PATH \
-    hostname
+mpi_params="-np $NP --hostfile ${HOSTFILE} --map-by node --allow-run-as-root -x PATH -x LD_LIBRARY_PATH --mca opal_common_ucx_opal_mem_hooks 1 --mca plm_rsh_args -p12345 "
 
 # shellcheck disable=SC2086
-mpirun \
-    -np $NP \
-    --hostfile ${HOSTFILE} \
-    --map-by node \
-    --allow-run-as-root \
-    --mca plm_rsh_args '-p 12345' \
-    -x PATH \
-    -x LD_LIBRARY_PATH \
-    cat /proc/1/cgroup
+mpirun $mpi_params hostname
+
+# shellcheck disable=SC2086
+mpirun $mpi_params cat /proc/1/cgroup
 
 echo "INFO: UCC MPI unit tests (CPU/GPU with NCCL) ..."
 # shellcheck disable=SC2086
-mpirun \
-    -np $NP \
-    --hostfile ${HOSTFILE} \
-    --map-by node \
-    --allow-run-as-root \
-    --mca plm_rsh_args '-p 12345' \
-    -x PATH \
-    -x LD_LIBRARY_PATH \
-    /opt/nvidia/src/ucc/build/test/mpi/ucc_test_mpi --mtypes host,cuda --inplace 2 --set_device 1 --root random:2 --count_bits 32,64 --displ_bits 32,64
+mpirun $mpi_params /opt/nvidia/src/ucc/build/test/mpi/ucc_test_mpi --mtypes host,cuda --inplace 2 --set_device 1\
+       --root random:2 --count_bits 32,64 --displ_bits 32,64
 echo "INFO: UCC MPI unit tests (CPU/GPU with NCCL) ... DONE"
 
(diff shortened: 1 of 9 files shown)


Sergei-Lebedev (approved):

.ci/scripts/run_tests_ucc_mpi.sh
```diff
@@ -24,99 +24,44 @@ export MASTER_ADDR=${HEAD_NODE}
...
-mpirun \
-    -np $NP \
-    --hostfile ${HOSTFILE} \
-    --map-by node \
-    --allow-run-as-root \
-    --mca plm_rsh_args '-p 12345' \
-    -x PATH \
-    -x LD_LIBRARY_PATH \
-    hostname
+mpirun --mca plm_rsh_args '-p 12345' $mpi_params hostname
```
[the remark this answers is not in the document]
vspetrov: i tried ... couldn't make bash work it either makes: --mca pml_rsh_args -p 12345 (w/o '') and then it does not work; OR it makes -mca pml_rsh_args \"' -p 1234 '\". bottom line: i don't know how to properly escape it: plz help if you know how.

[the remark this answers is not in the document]
Sergei-Lebedev: will it work if you omit quotes for args
```
mpi_params="-np $NP --hostfile ${HOSTFILE} --map-by node --allow-run-as-root -x PATH -x LD_LIBRARY_PATH --mca opal_common_ucx_opal_mem_hooks 1  -mca pml_rsh_args -p12345"
```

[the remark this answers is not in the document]
vspetrov: right should work, updating

.ci/scripts/run_tests_ucc_mpi.sh
```diff
@@ -24,99 +24,44 @@ export MASTER_ADDR=${HEAD_NODE}
 
 NP=$(wc --lines "$HOSTFILE" | awk '{print $1}')
 
+mpi_params="-np $NP --hostfile ${HOSTFILE} --map-by node --allow-run-as-root -x PATH -x LD_LIBRARY_PATH --mca opal_common_ucx_opal_mem_hooks 1 --mca plm_rsh_args -p12345 "
```
bureddy, line 27: why do we need opal_common_ucx_opal_mem_hooks ?

src/components/ec/cuda/ec_cuda.h
```diff
@@ -142,8 +143,11 @@ extern ucc_ec_cuda_t ucc_ec_cuda;
         cudaError_t cuda_st = cudaSuccess;                                     \
         ucc_spin_lock(&ucc_ec_cuda.init_spinlock);                             \
         if (ucc_ec_cuda.stream == NULL) {                                      \
-            cuda_st = cudaStreamCreateWithFlags(&ucc_ec_cuda.stream,           \
+            cudaStream_t _stream;                                              \
+            cuda_st = cudaStreamCreateWithFlags(&_stream,                      \
                                                 cudaStreamNonBlocking);        \
+            ucs_atomic_cswap64((uint64_t*)(uintptr_t)&ucc_ec_cuda.stream,      \
+                               (uint64_t)NULL, (uint64_t)_stream);             \
```
bureddy, line 150: why is this needed? seems there is already lock above

vspetrov: the lock is not enough, there is a race:
consider 2 threads. 1st thread grabs the lock above and goes to streamCreate. Inside that API the stream gets allocated and the ptr != NULL, yet the stream object is not entirely initialized (we should not use the stream ptr until the return from the cudaStreamCreate call). However, another thread reads non-NULL stream value and uses it, e.g. in ucc_mc_memcpy. I've reproduced that in that PR with modified MT ucc_test_mpi. 

I use local variable to init stream + atomic to write 8bytes of ptr to shared ucc_ec_cuda.stream atomically, so that other threads could not read incorrect ptr.

.ci/scripts/run_tests_ucc_mpi.sh
vspetrov: the way we build ompi (from master): it says
[swx-clx01:00208] common_ucx.c:164  Warning: UCX is unable to handle VM_UNMAP event. This may cause performance degradation or data corruption. Pls try adding --mca opal_common_ucx_opal_mem_hooks 1 to mpirun/oshrun command line to resolve this issue.

src/components/ec/cuda/ec_cuda.h
vspetrov: After discussion, change: using variable stream_initialized on ucc_mc/ec_cuda instead of atomic in order to avoid assumption sizeof(cudaStream_t) == 8.

src/components/ec/cuda/ec_cuda.h
```diff
@@ -11,6 +11,7 @@
 #include "components/ec/ucc_ec_log.h"
 #include "utils/arch/cuda_def.h"
 #include "utils/ucc_mpool.h"
+#include "utils/ucc_atomic.h"
```
bureddy, line 14: do we need this ?

bureddy (approved):

src/components/ec/cuda/ec_cuda.h
vspetrov: no

PR #479, opened 2022-04-18
API: Add pt2pt send/recv

wesbland:
## What
This PR proposes new interfaces for point-to-point send/recv using the existing collective infrastructure for simplicity.

## Why ?
There are many users of UCC that would benefit from simple pt2pt interfaces and indeed some of them even go around UCC to get them from UCX directly (e.g., [Torch-UCC](https://github.com/openucx/torch-ucc)). Some of the libraries underlying the TLs even support pt2pt operations, which makes implementing them in the TL themselves often relatively simple (e.g., NCCL and UCP).

## How ?
Rather than designing an entirely new API for the point-to-point operations which would largely duplicate the existing infrastructure, this work treats pt2pt send/recv calls as if they are simply another type of collective. These operations only require the participation of the two involved processes instead of the entire team and obey all of the usual semantics of point-to-point communication generally and UCC collectives specifically (e.g., using the same operation lifecycle).

The alternative implementation would have been to create new functions to initialize and start pt2pt operations, which would have required a significant amount of code duplication, both at the interface level and the underlying data structures. This seemed to be an unnecessary cost for something that otherwise had no need for new interfaces.

### DIFF
--- a/src/components/tl/nccl/tl_nccl_coll.c
+++ b/src/components/tl/nccl/tl_nccl_coll.c
@@ -1,6 +1,6 @@
 /**
  * Copyright (C) Mellanox Technologies Ltd. 2021-2022.  ALL RIGHTS RESERVED.
- * Copyright (c) Facebook, Inc. and its affiliates. 2021.
+ * Copyright (c) Meta Platforms, Inc. and affiliates. 2022.
  *
  * See file LICENSE for terms.
  */
@@ -728,6 +728,77 @@ ucc_status_t ucc_tl_nccl_gatherv_init(ucc_tl_nccl_task_t *task)
     }
 
     task->super.post = ucc_tl_nccl_gatherv_start;
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_nccl_send_start(ucc_coll_task_t *coll_task)
+{
+    ucc_status_t        status = UCC_OK;
+    ucc_tl_nccl_task_t *task   = ucc_derived_of(coll_task, ucc_tl_nccl_task_t);
+    ucc_coll_args_t    *args   = &TASK_ARGS(task);
+    ucc_tl_nccl_team_t *team   = TASK_TEAM(task);
+    ucc_ee_h            ee     = coll_task->ee;
+    void *              sbuf   = args->src.info.buffer;
+    uint64_t            peer   = args->root;
+    cudaStream_t        stream = (ee) ? (cudaStream_t) ee->ee_context : team->stream;
+    ncclDataType_t      nccl_dt;
+
+    task->super.status = UCC_INPROGRESS;
+
+    nccl_dt = ucc_to_nccl_dtype[UCC_DT_PREDEFINED_ID(TASK_ARGS(task).src.info.datatype)];
+    if (args->src.info.count == 0) {
+        task->super.status = UCC_OK;
+        return UCC_OK;
+    }
+    UCC_TL_NCCL_PROFILE_REQUEST_EVENT(coll_task, "nccl_send_start", 0);
+    NCCLCHECK_GOTO(ncclSend(sbuf, args->src.info.count, nccl_dt, peer,
+                            team->nccl_comm, stream),
+                   exit_coll, status, UCC_TL_TEAM_LIB(team));
+    status = ucc_tl_nccl_collective_sync(task, stream);
+
+exit_coll:
+    return status;
+}
+
+ucc_status_t ucc_tl_nccl_send_init(ucc_tl_nccl_task_t *task)
+{
+    task->super.post     = ucc_tl_nccl_send_start;
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_nccl_recv_start(ucc_coll_task_t *coll_task)
+{
+    ucc_status_t        status = UCC_OK;
+    ucc_tl_nccl_task_t *task   = ucc_derived_of(coll_task, ucc_tl_nccl_task_t);
+    ucc_coll_args_t    *args   = &TASK_ARGS(task);
+    ucc_tl_nccl_team_t *team   = TASK_TEAM(task);
+    ucc_ee_h            ee     = coll_task->ee;
+    void *              rbuf   = args->dst.info.buffer;
+    uint64_t            peer   = args->root;
+    cudaStream_t        stream = (ee) ? (cudaStream_t) ee->ee_context : team->stream;
+    size_t data_size;
+    ncclDataType_t      nccl_dt;
+
+    task->super.status = UCC_INPROGRESS;
+
+    nccl_dt = ucc_to_nccl_dtype[UCC_DT_PREDEFINED_ID(TASK_ARGS(task).dst.info.datatype)];
+    if (args->src.info.count == 0) {
+        task->super.status = UCC_OK;
+        return UCC_OK;
+    }
+    UCC_TL_NCCL_PROFILE_REQUEST_EVENT(coll_task, "nccl_recv_start", 0);
+    NCCLCHECK_GOTO(ncclRecv(rbuf, args->dst.info.count, nccl_dt, peer,
+                            team->nccl_comm, stream),
+                   exit_coll, status, UCC_TL_TEAM_LIB(team));
+    status = ucc_tl_nccl_collective_sync(task, stream);
+
+exit_coll:
+    return status;
+}
+
+ucc_status_t ucc_tl_nccl_recv_init(ucc_tl_nccl_task_t *task)
+{
+    task->super.post     = ucc_tl_nccl_recv_start;
     return UCC_OK;
 }
 
--- a/src/components/tl/ucp/recv/recv.c
+++ b/src/components/tl/ucp/recv/recv.c
@@ -1,0 +1,48 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. 2022.
+ *
+ * See file LICENSE for terms.
+ */
+#include "config.h"
+#include "tl_ucp.h"
+#include "tl_ucp_coll.h"
+#include "tl_ucp_sendrecv.h"
+#include "recv.h"
+
+ucc_status_t ucc_tl_ucp_recv_init(ucc_tl_ucp_task_t *task)
+{
+    task->super.post     = ucc_tl_ucp_recv_start;
+    task->super.progress = ucc_tl_ucp_recv_progress;
+
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_ucp_recv_start(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t   *task       = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t   *team       = TASK_TEAM(task);
+    ucc_coll_args_t     *args       = &TASK_ARGS(task);
+    uint64_t             peer       = args->root;
+    void                *rbuf       = args->dst.info.buffer;
+    ucc_memory_type_t    mem_type   = args->dst.info.mem_type;
+    size_t               count      = args->dst.info.count;
+    ucc_datatype_t       dt         = args->dst.info.datatype;
+    size_t               dt_size    = ucc_dt_size(dt);
+    size_t               data_size  = count * dt_size;
+
+    UCPCHECK_GOTO(ucc_tl_ucp_recv_nb(rbuf, data_size, mem_type, peer, team, task), task, out);
+
+    task->super.status = UCC_INPROGRESS;
+    ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+    return UCC_OK;
+
+out:
+    return task->super.status;
+}
+
+void ucc_tl_ucp_recv_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t   *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+
+    task->super.status = ucc_tl_ucp_test(task);
+}
--- a/src/components/tl/ucp/send/send.c
+++ b/src/components/tl/ucp/send/send.c
@@ -1,0 +1,48 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. 2022.
+ *
+ * See file LICENSE for terms.
+ */
+#include "config.h"
+#include "tl_ucp.h"
+#include "tl_ucp_coll.h"
+#include "tl_ucp_sendrecv.h"
+#include "send.h"
+
+ucc_status_t ucc_tl_ucp_send_init(ucc_tl_ucp_task_t *task)
+{
+    task->super.post     = ucc_tl_ucp_send_start;
+    task->super.progress = ucc_tl_ucp_send_progress;
+
+    return UCC_OK;
+}
+
+ucc_status_t ucc_tl_ucp_send_start(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t   *task       = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t   *team       = TASK_TEAM(task);
+    ucc_coll_args_t     *args       = &TASK_ARGS(task);
+    uint64_t             peer       = args->root;
+    void                *sbuf       = args->src.info.buffer;
+    ucc_memory_type_t    mem_type   = args->src.info.mem_type;
+    size_t               count      = args->src.info.count;
+    ucc_datatype_t       dt         = args->src.info.datatype;
+    size_t               dt_size    = ucc_dt_size(dt);
+    size_t               data_size  = count * dt_size;
+
+    UCPCHECK_GOTO(ucc_tl_ucp_send_nb(sbuf, data_size, mem_type, peer, team, task), task, out);
+
+    task->super.status = UCC_INPROGRESS;
+    ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+    return UCC_OK;
+
+out:
+    return task->super.status;
+}
+
+void ucc_tl_ucp_send_progress(ucc_coll_task_t *coll_task)
+{
+    ucc_tl_ucp_task_t   *task       = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+
+    task->super.status = ucc_tl_ucp_test(task);
+}
--- a/src/components/tl/ucp/tl_ucp_coll.h
+++ b/src/components/tl/ucp/tl_ucp_coll.h
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) Mellanox Technologies Ltd. 2021.  ALL RIGHTS RESERVED.
+ * Copyright (c) Meta Platforms, Inc. and affiliates. 2022.
  *
  * See file LICENSE for terms.
  */
@@ -187,8 +188,17 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
     ucc_tl_ucp_task_t *task    = ucc_tl_ucp_get_task(tl_team);
 
     ucc_coll_task_init(&task->super, coll_args, team);
-    tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
-    task->tagged.tag           = tl_team->seq_num;
+    if (coll_args->mask & UCC_COLL_ARGS_FIELD_TAG) {
+        task->tagged.tag           = task->tagged.tag;
+    } else if (coll_args->args.coll_type == UCC_COLL_TYPE_SEND ||
+               coll_args->args.coll_type == UCC_COLL_TYPE_RECV) {
+        /* Use the maximum tag with the point-to-point bit set as a reserved tag
+         * for point-to-point messages that do not otherwies have a tag set. */
+        task->tagged.tag           = UCC_TL_UCP_MAX_TAG;
+    } else {
+        tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
+        task->tagged.tag           = tl_team->seq_num;
+    }
     task->super.finalize       = ucc_tl_ucp_coll_finalize;
     task->super.triggered_post = ucc_triggered_post;
     return task;
--- a/src/ucc/api/ucc.h
+++ b/src/ucc/api/ucc.h
@@ -4,6 +4,7 @@
  * @copyright Copyright (C) Mellanox Technologies Ltd. 2020-2021.  ALL RIGHTS RESERVED.
  * @copyright Copyright (C) Huawei Technologies Co., Ltd. 2020.  ALL RIGHTS RESERVED.
  * @copyright Copyright (C) UChicago Argonne, LLC. 2022.  ALL RIGHTS RESERVED.
+ * @copyright Copyright (c) Meta Platforms, Inc. and affiliates. 2022.
  *
  * See file LICENSE for terms.
  */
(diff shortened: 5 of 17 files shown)


src/ucc/api/ucc.h
```diff
@@ -149,11 +149,13 @@ typedef enum {
     UCC_COLL_TYPE_FANOUT             = UCC_BIT(8),
     UCC_COLL_TYPE_GATHER             = UCC_BIT(9),
     UCC_COLL_TYPE_GATHERV            = UCC_BIT(10),
-    UCC_COLL_TYPE_REDUCE             = UCC_BIT(11),
-    UCC_COLL_TYPE_REDUCE_SCATTER     = UCC_BIT(12),
-    UCC_COLL_TYPE_REDUCE_SCATTERV    = UCC_BIT(13),
-    UCC_COLL_TYPE_SCATTER            = UCC_BIT(14),
-    UCC_COLL_TYPE_SCATTERV           = UCC_BIT(15),
+    UCC_COLL_TYPE_RECV               = UCC_BIT(11),
```
vspetrov, line 152: need to add new ops in the end of enum, otherwise we break ABI

src/components/tl/ucp/send/send.c
```diff
@@ -0,0 +1,54 @@
...
+        task->super.status = UCC_INPROGRESS;
+        ucc_progress_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+        return UCC_OK;
+    }
+
+out:
+    return task->super.status;
+}
+
+ucc_status_t ucc_tl_ucp_send_progress(ucc_coll_task_t *coll_task)
```
vspetrov, line 47: plz check new TL interface. progress fn is void, only status should be set.

src/components/tl/ucp/send/send.c
```diff
@@ -0,0 +1,54 @@
...
+    size_t               dt_size    = ucc_dt_size(dt);
+    size_t               data_size  = count * dt_size;
+
+    UCPCHECK_GOTO(ucc_tl_ucp_send_nb(sbuf, data_size, mem_type, peer, team, task), task, out);
+    if (task->send_posted == task->send_completed) {
+        task->super.status = UCC_OK;
+        return ucc_task_complete(coll_task);
+    } else {
+        task->super.status = UCC_INPROGRESS;
+        ucc_progress_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
```
vspetrov, line 39: use ucc_progress_queue_enqueue

jladd-mlnx:
@Sergei-Lebedev @souravzzz - can you guys take a look, please.

wesbland:
@vspetrov I've added an additional patch here that I believe addresses your concerns we discussed about tagged messages.

From what I could find, there wasn't actually anywhere in any TL where the user's tag was being honored by checking `UCC_COLL_ARGS_FIELD_TAG`. So I added a patch that honors that anywhere that it's used (in a collective or pt2pt) in the UCP TL. Let me know if you think that addresses your concern about overlapping tag space.

I also somewhat arbitrarily chose to take the high-order tag bit out of the tag space to indicate when we're using a user-specified tag. Another solution would be to shift the tag space around a bit to take the bit from somewhere else, but this has the least impact. I'm not sure if there's a big concern about the number of tags available or not given that I assume it only matters if you have more outstanding collectives than you do tags.

cc: @kingchc

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -644,6 +644,77 @@ ucc_status_t ucc_tl_nccl_gather_init(ucc_tl_nccl_task_t *task)
...
+}
+
+ucc_status_t ucc_tl_nccl_recv_start(ucc_coll_task_t *coll_task)
+{
+    ucc_status_t        status = UCC_OK;
+    ucc_tl_nccl_task_t *task   = ucc_derived_of(coll_task, ucc_tl_nccl_task_t);
+    ucc_coll_args_t    *args   = &TASK_ARGS(task);
+    ucc_tl_nccl_team_t *team   = TASK_TEAM(task);
+    ucc_ee_h            ee     = coll_task->ee;
+    void *              rbuf   = args->src.info.buffer;
```
Sergei-Lebedev, line 689: here rbuf is args->src.info.buffer but in tl ucp rbuf is args->dst.info.buffer

src/components/tl/ucp/recv/recv.c
```diff
@@ -0,0 +1,52 @@
...
+    uint64_t             peer       = args->root;
+    void                *rbuf       = args->dst.info.buffer;
+    ucc_memory_type_t    mem_type   = args->dst.info.mem_type;
+    size_t               count      = args->dst.info.count;
+    ucc_datatype_t       dt         = args->dst.info.datatype;
+    size_t               dt_size    = ucc_dt_size(dt);
+    size_t               data_size  = count * dt_size;
+
+    UCPCHECK_GOTO(ucc_tl_ucp_recv_nb(rbuf, data_size, mem_type, peer, team, task), task, out);
+    if (task->tagged.recv_posted == task->tagged.recv_completed) {
```
Sergei-Lebedev, line 34: I think this if statement is not needed, you can just do return ucc_progress_queue_enqueue(...)

Sergei-Lebedev:
I think this proposal makes sense and changes in UCC API are minimal which is good, but just for consideration here is another proposal. To emphasize that UCC is not generic point to point library we can add two new collectives UCC_COLL_TYPE_UNICAST and UCC_COLL_TYPE_MULTICAST. Altogether UCC interface defines

- BCAST: one to every team member communication
- MULTICAST: one to many communication, in contrast to broadcast some team ranks set coll_args.dst.info.count =0 indicating that they won't receive data
- UNICAST: one to one a.k.a p2p communication, special flags UCC_COLL_ARGS_FLAG_SENDER or UCC_COLL_ARGS_FLAG_RECEIVER should be set for sender or receiver correspondingly

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -188,7 +189,11 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
 
     ucc_coll_task_init(&task->super, coll_args, team);
     tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
```
kingchc, line 191: Come to thinking about one scenario that one calls pt2pt without setting the tag. So those two ranks who are involved in pt2pt will use and increase the `seq_num`, while other ranks won't. This seems to cause tag mismatch in next set of collectives or pt2pt from different pair.

Just a few thoughts for brainstorming:
1) move this increment to the `else` block below (i.e., only increase seq_num if tag is not used), this probably not enough
2) have two internal seq_num with different ranges, one for pt2pt and one for collective to avoid collision. And plus 1) to avoid mismatch
3) force/assume all ranks to be involved during init even for pt2pt (sounds unnecessary) 
4) force pt2pt to always provide tag (even a dummy one) and only use internal seq_num for collective

wesbland: @kingchc and I had a conversation offline about this, but to summarize here:

> 1. move this increment to the `else` block below (i.e., only increase seq_num if tag is not used), this probably not enough

That doesn't solve the problem unless you pair with 4) below and require the user to always provide a tag when using pt2pt (which doesn't necessarily make sense if you're using something like the NCCL TL where tags aren't supported anyway).

> 2. have two internal seq_num with different ranges, one for pt2pt and one for collective to avoid collision. And plus 1) to avoid mismatch

This would solve the problem, but only if you keep track of a sequence number per-pair, which gets pretty unwieldy. My preference would be to say that if the user doesn't specify a tag, we just use a reserved value and all messages get sent with that tag (after adding the high-order bit to indicate that it's a pt2pt message). That way it just acts like an untagged message (in the sense that it can be matched in any order), but it doesn't conflict with any of the collective messages.

> 3. force/assume all ranks to be involved during init even for pt2pt (sounds unnecessary)

😄 

> 4. force pt2pt to always provide tag (even a dummy one) and only use internal seq_num for collective

This is the second best option in my mind in terms of keeping a clean interface. However, it does add a burden on the user that might be a little weird (see above). My preference would be for 2 and I'll update the PR to reflect that.

manjugv:
I guess we need to understand the use case better.  If we add send/recv as proposed, intentionally or unintentionally, we are going towards a path of defining all p2p operations in UCC. My concern is that UCC cannot provide better (richer or better performant) p2p operations than UCX and we will end up defining UCX. 
  
Other items that need to be clarified and some of them I mentioned during the WG call: 
1. What is the semantic relation between collectives and p2p operation in the use case? 
     — Does it use the same EP ranking?
     — What is the ordering between p2p calls?
     — What is the ordering between p2p and collectives?
     — What is the ordering in presence of multiple threads?
     --  Can send/recv happen between teams?
2. Do we intend to define all p2p interfaces here? Why or why not? 
3. If we add p2p API to UCC, we need to support other users of UCC that can use p2p - MPI / PGAS. How can we support it?
4. In single-threaded mode and multi-threaded mode, what are the ordering requirements between p2p and collectives? 
5. When a team is created in UCC_COLLECTIVE_INIT_AND_POST_ORDERED mode, how do we establish order between the collectives if only a few eps call the collective? 
6. We have synchronous and asynchronous collectives. How do we define this for send/recv?
7. Collective init and destroy is a collective operation. Should collective init be invoked by both send/receive? Are we breaking backward compatibility by introducing non-collective invocation behavior?
8. What about the triggered post? Should we support it? Why or Why not?

manjugv:
I guess we need to understand the use case better.  If we add send/recv as proposed, intentionally or unintentionally, we are going towards a path of defining all p2p operations in UCC. My concern is that UCC cannot provide better (richer or better performant) p2p operations than UCX and we will end up defining UCX. 
  
Other items that need to be clarified and some of them I mentioned during the WG call: 
1. What is the semantic relation between collectives and p2p operation in the use case? 
     — Does it use the same EP ranking?
     — What is the ordering between p2p calls?
     — What is the ordering between p2p and collectives?
     — What is the ordering in presence of multiple threads?
     --  Can send/recv happen between teams?
2. Do we intend to define all p2p interfaces here? Why or why not? 
3. If we add p2p API to UCC, we need to support other users of UCC that can use p2p - MPI / PGAS. How can we support it?
4. In single-threaded mode and multi-threaded mode, what are the ordering requirements between p2p and collectives? 
5. When a team is created in UCC_COLLECTIVE_INIT_AND_POST_ORDERED mode, how do we establish order between the collectives if only a few eps call the collective? 
6. We have synchronous and asynchronous collectives. How do we define this for send/recv?
7. Collective init and destroy is a collective operation. Should collective init be invoked by both send/receive? Are we breaking backward compatibility by introducing non-collective invocation behavior?
8. What about the triggered post? Should we support it? Why or Why not?

manjugv:
Another approach we should explore is creating a team cheaply for p2p participants. That way it is a collective op rather than a p2p operation. Today in some cases, we don’t need to exchange any information during team creation (it is done during context). So, we should explore that approach here. UCC API is equipped to do this. 

To elaborate:
- Create a (shared) UCC context with OOB and EP list; so all information is exchanged (allgather is done here) during the context create
- Share the UCC context for collectives (coll_team) and p2p teams (p2p_teams) team create
- Call p2p team creation with no OOB (imitating Active Sets behavior in OpenSHMEM)
   - This then will be a local operation. Though all participants (in this case two) should call it, there is no collective operation or information exchanged
   - The overhead for such a team creation should be very minimal (few function calls) 
- If you have a fixed pattern for p2p ops, it is even better, since you can use persistent semantics.

(We might have to rethink team creation implementation to achieve this optimization but seems doable)

manjugv:
Another approach we should explore is creating a team cheaply for p2p participants. That way it is a collective op rather than a p2p operation. Today in some cases, we don’t need to exchange any information during team creation (it is done during context). So, we should explore that approach here. UCC API is equipped to do this. 

To elaborate:
- Create a (shared) UCC context with OOB and EP list; so all information is exchanged (allgather is done here) during the context create
- Share the UCC context for collectives (coll_team) and p2p teams (p2p_teams) team create
- Call p2p team creation with no OOB (imitating Active Sets behavior in OpenSHMEM)
   - This then will be a local operation. Though all participants (in this case two) should call it, there is no collective operation or information exchanged
   - The overhead for such a team creation should be very minimal (few function calls) 
- If you have a fixed pattern for p2p ops, it is even better, since you can use persistent semantics.

(We might have to rethink team creation implementation to achieve this optimization but seems doable)

manjugv:
> @vspetrov I've added an additional patch here that I believe addresses your concerns we discussed about tagged messages.
> 
> From what I could find, there wasn't actually anywhere in any TL where the user's tag was being honored by checking `UCC_COLL_ARGS_FIELD_TAG`. So I added a patch that honors that anywhere that it's used (in a collective or pt2pt) in the UCP TL. Let me know if you think that addresses your concern about overlapping tag space.
> 
> I also somewhat arbitrarily chose to take the high-order tag bit out of the tag space to indicate when we're using a user-specified tag. Another solution would be to shift the tag space around a bit to take the bit from somewhere else, but this has the least impact. I'm not sure if there's a big concern about the number of tags available or not given that I assume it only matters if you have more outstanding collectives than you do tags.
> 
> cc: @kingchc



> @vspetrov I've added an additional patch here that I believe addresses your concerns we discussed about tagged messages.
> 
> From what I could find, there wasn't actually anywhere in any TL where the user's tag was being honored by checking `UCC_COLL_ARGS_FIELD_TAG`. So I added a patch that honors that anywhere that it's used (in a collective or pt2pt) in the UCP TL. Let me know if you think that addresses your concern about overlapping tag space.
> 
> I also somewhat arbitrarily chose to take the high-order tag bit out of the tag space to indicate when we're using a user-specified tag. Another solution would be to shift the tag space around a bit to take the bit from somewhere else, but this has the least impact. I'm not sure if there's a big concern about the number of tags available or not given that I assume it only matters if you have more outstanding collectives than you do tags.
> 
> cc: @kingchc
(I don't have the exact context of offline conversation, so ignore it if this is redundant :))

I think this might not be clear from the documentation. So let me clarify. UCC_COLL_ARGS_FIELD_TAG makes sense only for teams created for unordered collectives. We have a few ordering models that we support (from an API perspective). We haven’t implemented unordered collectives yet. So, it is not surprising we have ignored checking it. 

In the case of unordered collectives, we expect the user to provide a tag and library to order it based on the tag. We cannot mix ordered and unordered collectives as defined today. Mixing them will be a nightmare for implementations.

manjugv:
> @vspetrov I've added an additional patch here that I believe addresses your concerns we discussed about tagged messages.
> 
> From what I could find, there wasn't actually anywhere in any TL where the user's tag was being honored by checking `UCC_COLL_ARGS_FIELD_TAG`. So I added a patch that honors that anywhere that it's used (in a collective or pt2pt) in the UCP TL. Let me know if you think that addresses your concern about overlapping tag space.
> 
> I also somewhat arbitrarily chose to take the high-order tag bit out of the tag space to indicate when we're using a user-specified tag. Another solution would be to shift the tag space around a bit to take the bit from somewhere else, but this has the least impact. I'm not sure if there's a big concern about the number of tags available or not given that I assume it only matters if you have more outstanding collectives than you do tags.
> 
> cc: @kingchc



> @vspetrov I've added an additional patch here that I believe addresses your concerns we discussed about tagged messages.
> 
> From what I could find, there wasn't actually anywhere in any TL where the user's tag was being honored by checking `UCC_COLL_ARGS_FIELD_TAG`. So I added a patch that honors that anywhere that it's used (in a collective or pt2pt) in the UCP TL. Let me know if you think that addresses your concern about overlapping tag space.
> 
> I also somewhat arbitrarily chose to take the high-order tag bit out of the tag space to indicate when we're using a user-specified tag. Another solution would be to shift the tag space around a bit to take the bit from somewhere else, but this has the least impact. I'm not sure if there's a big concern about the number of tags available or not given that I assume it only matters if you have more outstanding collectives than you do tags.
> 
> cc: @kingchc
(I don't have the exact context of offline conversation, so ignore it if this is redundant :))

I think this might not be clear from the documentation. So let me clarify. UCC_COLL_ARGS_FIELD_TAG makes sense only for teams created for unordered collectives. We have a few ordering models that we support (from an API perspective). We haven’t implemented unordered collectives yet. So, it is not surprising we have ignored checking it. 

In the case of unordered collectives, we expect the user to provide a tag and library to order it based on the tag. We cannot mix ordered and unordered collectives as defined today. Mixing them will be a nightmare for implementations.

wesbland:
> Another approach we should explore is creating a team cheaply for p2p participants. That way it is a collective op rather than a p2p operation. Today in some cases, we don’t need to exchange any information during team creation (it is done during context). So, we should explore that approach here. UCC API is equipped to do this.

The obvious counter-argument here is that anything using this API would have to create and maintain a cache of data structures on top of UCC in a way that every time it wants to make a pt2pt call, it needs to first check to see if a team already exists, then either create one or fetch it, then finally perform the pt2pt operation. This adds both latency and memory overhead in a way that seems somewhat unnecessary (not to mention the complexity necessary in the cache itself).

> To elaborate:
> 
> * Create a (shared) UCC context with OOB and EP list; so all information is exchanged (allgather is done here) during the context create
> * Share the UCC context for collectives (coll_team) and p2p teams (p2p_teams) team create
> * Call p2p team creation with no OOB (imitating Active Sets behavior in OpenSHMEM)
>   * This then will be a local operation. Though all participants (in this case two) should call it, there is no collective operation or information exchanged
>   * The overhead for such a team creation should be very minimal (few function calls)
> * If you have a fixed pattern for p2p ops, it is even better, since you can use persistent semantics.
> 
> (We might have to rethink team creation implementation to achieve this optimization but seems doable)

Assuming one accepts the premise above, this implementation sounds reasonable. I can see where some of these optimizations are missing today, but that's an implementation issue, not a spec issue. 👍 

----

I think the point we were trying to make by introducing these APIs is to expose functionality that already existed at the lower levels of the TLs to be able to do simple pt2pt for things like control messages without requiring users of UCC to also pull in another library and the associated software overheads (both in terms of performance and maintenance) that go with it.

It's possible that the "Unicast" idea floated by @Sergei-Lebedev above would work if the issue is naming. You get very similar semantics that way and under the hood, it would be implemented using UCC/NCCL pt2pt calls.

wesbland:
Thanks for the well thought out responses @manjugv. Sorry it took me a while to get all my thoughts together here. Hopefully my (very long) answer will summarize some of the thinking behind this proposal and we can keep iterating on the places that I hadn't thought through yet.

> I guess we need to understand the use case better. If we add send/recv as proposed, intentionally or unintentionally, we are going towards a path of defining all p2p operations in UCC. My concern is that UCC cannot provide better (richer or better performant) p2p operations than UCX and we will end up defining UCX.

I think we can summarize the use case by saying that we're trying to simplify the use of communication libraries in the case of codes like [Torch-UCC](https://github.com/openucx/torch-ucc), where we currently have to make assumptions about the underlying runtime for simplicity. For example, we assume we'll always use UCP/UCX for pt2pt when we could have used NCCL pt2pt and avoided creating additional runtime overheads by spinning up an instance of UCX separately from NCCL. By adding these APIs, we can [simplify the code](https://github.com/facebookresearch/torch_ucc/pull/72) to use UCC for everything and let UCC pick the appropriate runtime to avoid the extra overheads.

> Other items that need to be clarified and some of them I mentioned during the WG call:

These are all good points. I don't have answers to all of these of course, but I can put out a few thoughts here. I'm sure I'm going to get some of the UCC semantics wrong as I haven't fully grokked the API docs, so please correct me when I do.

> 1. What is the semantic relation between collectives and p2p operation in the use case?
>    — Does it use the same EP ranking?

In this proposal at least, the teams remain the same and we're essentially introducing the concept of "partial collectives" where not every member of the team needs to be involved in initializing a collective (if they won't be participating in said collective). Therefore, the ranking is the same.

>    — What is the ordering between p2p calls?

The intention here (and I think it might be incorrect based on your [later reply](https://github.com/openucx/ucc/pull/479#issuecomment-1104566814)) is that pt2pt operations would have the same ordering semantics as collectives. It seems to me from reading the code that collective ordering is provided by iterating a tag value (at least in the UCP TL) or the underlying blocking nature of the calls (in NCCL TL). For pt2pt operations, the tag is optional. So if the tag is provided, matching would be provided by the tag value. If two receive operations are issued simultaneously with the same tag, they could be matched in any order. If no tag is provided, they could be matched in any order.

>    — What is the ordering between p2p and collectives?

The ordering should be the same between collectives as it is between collectives and pt2pt. On the UCP side, I enforced this by taking the highest order bit from the user tag to indicate whether the communication or pt2pt.

>    — What is the ordering in presence of multiple threads?

This is a place where I can definitely use a better understanding as I'm not 100% clear on the intention of the UCC API after reading the [docs](https://openucx.github.io/ucc/api/v1.0/html/index.html). The intention here, again, is to be consistent with what is already there. If you're involved in pt2pt, it's the same as being involved in a collective.

I assume that the case you're concerned about is if you have some processes involved in a collective and others have a pt2pt in-between two other collectives? You want to make sure that everything matches up properly. Is that not already covered by the case where you could have processes in multiple teams?

>    --  Can send/recv happen between teams?

No.

> 2. Do we intend to define all p2p interfaces here? Why or why not?

By "all p2p interfaces", do you mean adding non-blocking communication? Additional datatypes? Anysource receive?

My intention at least is to only expose the APIs that are available from the TLs and not to re-implement all of MPI on top of UCC.

> 3. If we add p2p API to UCC, we need to support other users of UCC that can use p2p - MPI / PGAS. How can we support it?

As I understand it, the current users of UCC are a different set of apps/libraries than MPI libraries. Things like Torch-UCC only want a very small subset of what MPI would provide and therefore only want the runtime layers underneath them to support that subset.

MPI/PGAS has many more requirements than what UCC supports (e.g., one-sided communication, complex datatypes, etc.) and more than what at least some of the TLs underneath it could support (e.g., NCCL). For those use cases, the MPI libraries should do what they're already doing now: use UCX or OFI.

> 4. In single-threaded mode and multi-threaded mode, what are the ordering requirements between p2p and collectives?

Is this the same question as above?

> 5. When a team is created in UCC_COLLECTIVE_INIT_AND_POST_ORDERED mode, how do we establish order between the collectives if only a few eps call the collective?

That's a good question. I wasn't aware of that mode. I would think that we would need to define this mode as invalid for collectives that do not involve the entire team or expand the definition to say something like "When set to this value, the collective participants shall initialize and post the operation in the same order for operations in which they are involved."

> 6. We have synchronous and asynchronous collectives. How do we define this for send/recv?

This isn't something I covered in this PR, but it could be implemented with the same semantics as `MPI_SSEND`. It just requires additional OOB RTS/CTS calls before the main call.

> 7. Collective init and destroy is a collective operation. Should collective init be invoked by both send/receive? Are we breaking backward compatibility by introducing non-collective invocation behavior?

I would expect the sender and receiver to invoke their respective calls and that we would define send/recv to be collective with each other. If this is semantically weird, we could change the API to be something else like `UCC_COLL_TYPE_PT2PT` and have an additional flag somewhere that indicates whether a process is a sender or receiver (similar to the unicast proposal by @Sergei-Lebedev [above](https://github.com/openucx/ucc/pull/479#issuecomment-1103574777).

> 8. What about the triggered post? Should we support it? Why or Why not?

I don't see why not. Are there other APIs that are not allowed to be posted from a triggered op?

wesbland:
I should have posted this after our discussion last week, but a few of us got together on a call to discuss the way forward for this idea.

The end result was that we decided to add the notion of specifying that a collective would happen on a subset of the processes within a team (i.e., the "active sets" notion mentioned by @manjugv [above](https://github.com/openucx/ucc/pull/479#issuecomment-1104564412)). This would allow the caller to have a "two-process bcast" that would have essentially the same semantics as send/recv without explicitly being a pt2pt call.

This also opens up the opportunity for a more flexible bcast that has an number of participants up to the entire team. For now, this will probably not be implemented. It will also most likely have a performance penalty over creating a new team as it will require more setup costs when initializing the collective operation. However, there is an opportunity to provide this feature in the future and with good caching, it may be possible to avoid the performance penalties. For a two-process bcast, this penalty can be avoided because there is less need to negotiate between the participating processes for an internal team ID for matching.

@vspetrov agreed to create a PR to implement this idea and I will rebase this PR on top of that one when it is ready.

If I got any of that wrong, please correct me. Unfortunately I didn't come write these notes immediately after our discussion last week so I've slept a few times since then.

manjugv:
@wesbland Good summary. Thanks!

vspetrov:
@wesbland all is correct. I will work on the PR this week (we have short week due to RF holidays, hooefully ill push the PR by the end of it)

PR #482, opened 2022-04-19
Use UCS_ALLOW_LIST for TLS parameter

vspetrov:
## What
Changes the way UCC_TLS (and UCC_CL_BASIC/HIER_TLS and UCC_CL_HIER_<SBGP_TYPE>_TLS) is handled. It used to be ucc_config_names_array_t. Now it will be ucc_config_allow_list_t. The difference is that the new type supports NEGATE option. Ie, now user can set UCC_TLS=^sharp, which means use everything except for specified value. This is not only convenience. 
Before we had to specify the full list, e.g. UCC_TLS=cuda,ucp. However, this would be considered a list of TLs explicitly requested by user; if some TL fails to init it will cause an error output from UCC. New way allows disabling TL w/o forcing others.

PR also fixes config_names_array_merge implementation and clarifies TLs specification in CL/HIER.

### DIFF
--- a/src/components/cl/basic/cl_basic_lib.c
+++ b/src/components/cl/basic/cl_basic_lib.c
@@ -53,9 +53,9 @@ static inline ucc_status_t check_tl_lib_attr(const ucc_base_lib_t *lib,
 ucc_status_t ucc_cl_basic_get_lib_attr(const ucc_base_lib_t *lib,
                                        ucc_base_lib_attr_t  *base_attr)
 {
-    ucc_cl_lib_attr_t       *attr   = ucc_derived_of(base_attr, ucc_cl_lib_attr_t);
-    ucc_cl_basic_lib_t      *cl_lib = ucc_derived_of(lib, ucc_cl_basic_lib_t);
-    ucc_config_allow_list_t *tls    = &cl_lib->super.tls;
+    ucc_cl_lib_attr_t  *attr     = ucc_derived_of(base_attr, ucc_cl_lib_attr_t);
+    ucc_cl_basic_lib_t *cl_lib   = ucc_derived_of(lib, ucc_cl_basic_lib_t);
+    ucc_config_allow_list_t *tls = &cl_lib->super.tls;
     ucc_tl_iface_t          *tl_iface;
     int                      i;
     ucc_status_t             status;
@@ -79,10 +79,10 @@ ucc_status_t ucc_cl_basic_get_lib_attr(const ucc_base_lib_t *lib,
     for (i = 0; i < tls->array.count; i++) {
         /* Check TLs proveded in CL_BASIC_TLS. Not all of them could be
            available, check for NULL. */
-        tl_iface = ucc_derived_of(
-            ucc_get_component(&ucc_global_config.tl_framework,
-                              tls->array.names[i]),
-            ucc_tl_iface_t);
+        tl_iface =
+            ucc_derived_of(ucc_get_component(&ucc_global_config.tl_framework,
+                                             tls->array.names[i]),
+                           ucc_tl_iface_t);
         if (!tl_iface) {
             cl_warn(lib, "tl %s is not available", tls->array.names[i]);
             continue;
--- a/src/components/cl/basic/cl_basic_lib.c
+++ b/src/components/cl/basic/cl_basic_lib.c
@@ -24,7 +24,6 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_lib_t, const ucc_base_lib_params_t *params,
 UCC_CLASS_CLEANUP_FUNC(ucc_cl_basic_lib_t)
 {
     cl_info(&self->super, "finalizing lib object: %p", self);
-    ucc_config_names_array_free(&self->tls_forced);
 }
 
 UCC_CLASS_DEFINE(ucc_cl_basic_lib_t, ucc_cl_lib_t);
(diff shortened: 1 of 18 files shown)


src/utils/ucc_parser.c
```diff
@@ -101,3 +93,41 @@ int ucc_config_names_search(ucc_config_names_array_t *config_names,
...
+        out->array.names = ucc_malloc(sizeof(char *) * all->count, "names");
+        if (!out->array.names) {
+            ucc_error("failed to allocate %zd bytes for names array",
+                      sizeof(char *) * all->count);
+            return UCC_ERR_NO_MEMORY;
+        }
+
+        for (i = 0; i < all->count; i++) {
+            if (out->mode == UCC_CONFIG_ALLOW_LIST_ALLOW_ALL ||
+                (ucc_config_names_search(&list->array, all->names[i]) < 0)) {
```
Sergei-Lebedev, line 127: is out allow_list mode correct here? even if input list mode is NEGATE out list mode should be ALLOW anyway

src/utils/ucc_parser.c
```diff
@@ -101,3 +93,41 @@ int ucc_config_names_search(ucc_config_names_array_t *config_names,
...
+    out->mode        = list->mode;
+    out->array.names = NULL;
+
+    if (out->mode == UCC_CONFIG_ALLOW_LIST_ALLOW) {
+        status = ucc_config_names_array_dup(&out->array, &list->array);
+        if (UCC_OK != status) {
+            ucc_error("failed to dup config_names_array");
+            return status;
+        }
+    } else {
```
Sergei-Lebedev, line 113: i would rather use switch here, something like
```c
out->mode           = UCC_CONFIG_ALLOW_LIST_ALLOW;
out->array.names = NULL;
swtich (out->mode){
case UCC_CONFIG_ALLOW_LIST_ALLOW:
  ucc_config_names_array_dup(&out->array, &list->array);
case UCC_CONFIG_ALLOW_LIST_ALLOW_ALL:
  ucc_config_names_array_dup(&out->array, &all->array);
case UCC_CONFIG_ALLOW_LIST_NEGATE:
  for (i = 0; i < all->count; i++) {
            if (ucc_config_names_search(&list->array, all->names[i]) < 0)) {
                out->array.names[out->array.count++] = strdup(all->names[i]);
            }
default:
  ucc_error("not supported list mode");
  ucc_assert(0);
}
```

src/utils/ucc_parser.c
```diff
@@ -10,40 +10,32 @@
...
-                    if (0 != strcmp(src->names[i], dst->names[j])) {
-                        dst->names[dst->count - n_new] = strdup(src->names[i]);
-                        if (ucc_unlikely(!dst->names[dst->count - n_new])) {
-                            ucc_error("failed to dup config_names_array entry");
-                            return UCC_ERR_NO_MEMORY;
-                        }
-                        n_new--;
+                if (ucc_config_names_search(dst, src->names[i]) < 0) {
+                    dst->names[dst->count++] = strdup(src->names[i]);
+                    if (ucc_unlikely(!dst->names[dst->count - n_new])) {
```
Sergei-Lebedev, line 36: maybe if (ucc_unlikely(!dst->names[dst->count - 1)) ?

src/core/ucc_lib.c
```diff
@@ -213,16 +213,13 @@ static ucc_status_t ucc_cl_lib_init(const ucc_lib_params_t *user_params,
     return status;
 }
 
-static int ucc_tl_is_required(ucc_lib_info_t *lib, ucc_tl_iface_t *tl_iface,
-                              int explicitly)
+static int ucc_tl_is_required(ucc_lib_info_t *lib, ucc_tl_iface_t *tl_iface)
```
Sergei-Lebedev, line 216: is_required and is_requested almost identical, maybe merge them

src/components/cl/hier/cl_hier.h
```diff
@@ -63,7 +63,10 @@ typedef struct ucc_cl_hier_context_config {
 typedef struct ucc_cl_hier_lib {
     ucc_cl_lib_t             super;
     ucc_cl_hier_lib_config_t cfg;
-    ucc_config_names_array_t tls; /*< Intersection of UCC_CL_HIER_TLS vs sbgp_tls */
+    ucc_config_allow_list_t  tls; /*< Intersection of UCC_CL_HIER_TLS
+                                    vs sbgp_tls */
+    ucc_config_names_array_t tls_forced; /*< set of TLs that are
```
Sergei-Lebedev, line 68: seems like tls_force will be part of each CL, then it makes sense to add it to base class

src/components/cl/basic/cl_basic_lib.c
```diff
@@ -52,42 +53,42 @@ static inline ucc_status_t check_tl_lib_attr(const ucc_base_lib_t *lib,
...
-        for (i = 0; i < ucc_global_config.tl_framework.n_components; i++) {
-            tl_iface = ucc_derived_of(
-                ucc_global_config.tl_framework.components[i], ucc_tl_iface_t);
-            ucc_assert(tl_iface);
-            if (UCC_OK != (status = check_tl_lib_attr(lib, tl_iface, attr))) {
-                return status;
-            }
+
+    ucc_assert(tls->array.count >= 1);
+    for (i = 0; i < tls->array.count; i++) {
```
Sergei-Lebedev, line 79: this can be added to super class constructor if we move tls_forced to base class

src/utils/ucc_parser.c
vspetrov: the idea was that i just copy "mode" to out list and the only value i care about is mode = "ALLOW". This means the list is explicitly requested by user. Other 2 modes: negate or allow_all essentially have no special meaning. We have set of tls from user, we have set of ALL tls and we build another list from these 2.

src/utils/ucc_parser.c
vspetrov: looks good (except for mode, see comment above)

src/utils/ucc_parser.c
vspetrov: after discussion with Sergey changed to the new helper data struct with explicit "requested" field

src/core/ucc_lib.c
vspetrov: done

src/components/cl/basic/cl_basic_lib.c
vspetrov: this is different for different CLs. CL/HIER builds tls_forced list based on the corresponding TL lists of SBGP subgroups. hence, can't be moved to common superclass

src/components/cl/hier/cl_hier.h
vspetrov: moved tls_forced field to common super class. The cleanup is also made in base destructor. Initialization is still CL specific

vspetrov:
@Sergei-Lebedev @bureddy @manjugv review comments addressed. Plz have a look

bureddy (approved):

manjugv (approved):

vspetrov:
bot:retest

PR #489, opened 2022-04-23
TL/CUDA: fix CUDA stream sync in A2a(v)

kingchc:
## What
Fix CUDA stream sync in TL/CUDA Alltoall

## Why ?
Application might allocate their CUDA streams for communication, UCC should honor that if provided. Similar to TL/NCCL. 

## How ?
check if there is a stream/ee_context assigned in the task and use it for `cudaEventRecord ` and `cudaStreamWaitEvent `. Otherwise, use the team's internal stream.

Also move `cudaStreamWaitEvent ` before we launch executor and after SHM barrier within the team to ensure we start the A2a(v) when all remote devices are ready.

### DIFF
--- a/src/components/tl/cuda/alltoallv/alltoallv_ce.c
+++ b/src/components/tl/cuda/alltoallv/alltoallv_ce.c
@@ -7,6 +7,7 @@
 
 #include "alltoallv.h"
 #include "components/ec/ucc_ec.h"
+#include "core/ucc_ee.h"
 #include "tl_cuda_cache.h"
 #include "utils/arch/cpu.h"
 #include "utils/arch/cuda_def.h"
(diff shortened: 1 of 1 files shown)


kingchc:
cc @Sergei-Lebedev @vspetrov

src/components/tl/cuda/alltoallv/alltoallv_ce.c
```diff
@@ -34,6 +35,8 @@ ucc_status_t ucc_tl_cuda_alltoallv_setup_start(ucc_tl_cuda_task_t *task)
     ucc_tl_cuda_sync_t *sync = TASK_SYNC(task, UCC_TL_TEAM_RANK(team));
     ucc_status_t        status;
     ucc_coll_args_t    *args = &TASK_ARGS(task);
+    ucc_ee_h            ee   = task->super.ee;
+    cudaStream_t stream = (ee) ? (cudaStream_t)ee->ee_context : team->stream;
```
bureddy, line 39: align "="

Sergei-Lebedev (approved):

PR #494, opened 2022-04-29
TL/CUDA: multiring allgatherv and reduce_scatterv

Sergei-Lebedev:
## What
Support multiple rings in TL CUDA allgather and allgatherv

### DIFF
--- a/src/components/tl/cuda/tl_cuda.h
+++ b/src/components/tl/cuda/tl_cuda.h
@@ -22,6 +22,8 @@
 #endif
 
 #define UCC_TL_CUDA_MAX_PEERS 8
+#define UCC_TL_CUDA_MAX_RINGS 8
+
 #define UCC_TL_CUDA_SUPPORTED_COLLS                                            \
     (UCC_COLL_TYPE_ALLTOALL | UCC_COLL_TYPE_ALLTOALLV |                        \
      UCC_COLL_TYPE_ALLGATHER | UCC_COLL_TYPE_ALLGATHERV |                      \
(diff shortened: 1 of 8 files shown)


Sergei-Lebedev:
added multiring reduce_scatterv

src/components/tl/cuda/tl_cuda_lib.c
```diff
@@ -15,6 +15,11 @@ UCC_CLASS_INIT_FUNC(ucc_tl_cuda_lib_t, const ucc_base_lib_params_t *params,
     UCC_CLASS_CALL_SUPER_INIT(ucc_tl_lib_t, &ucc_tl_cuda.super,
                               &tl_config->super);
     memcpy(&self->cfg, tl_config, sizeof(*tl_config));
+    if (self->cfg.allgather_ring_max_rings > UCC_TL_CUDA_MAX_RINGS) {
```
vspetrov, line 18: same check for reduce_scatter_max_rings ?

vspetrov (approved):
@Sergei-Lebedev I tried to look carefully and it all looks good to me (just minor comment below). 👍

manjugv (approved):

src/components/tl/cuda/tl_cuda.h
```diff
@@ -22,6 +22,8 @@
 #endif
 
 #define UCC_TL_CUDA_MAX_PEERS 8
+#define UCC_TL_CUDA_MAX_RINGS 8
```
bureddy, line 25: will be any scenario in the future where we want to try mode than 8 rings? instead should we allocate memory dynamically

PR #497, opened 2022-05-05
API: Active Set args extension

vspetrov:
## What
Alternative to #479 based on WG discussion


vspetrov:
@wesbland @manjugv plz have a look at the proposal. Gist link shows how p2p can be done using the new extension.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -187,8 +188,17 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
...
+        task->tagged.tag           = task->tagged.tag;
+    } else if (coll_args->args.coll_type == UCC_COLL_TYPE_SEND ||
+               coll_args->args.coll_type == UCC_COLL_TYPE_RECV) {
+        /* Use the maximum tag with the point-to-point bit set as a reserved tag
+         * for point-to-point messages that do not otherwies have a tag set. */
+        task->tagged.tag           = UCC_TL_UCP_MAX_TAG;
+    } else {
+        tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
+        task->tagged.tag           = tl_team->seq_num;
+    }
```
wesbland, line 203: Is this the correct usage? I thought @manjugv [said](https://github.com/openucx/ucc/pull/479#issuecomment-1104566814) that this was incorrect. I didn't quite grok what should be happening here.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -188,13 +188,15 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
...
 
     ucc_coll_task_init(&task->super, coll_args, team);
-    if (coll_args->mask & UCC_COLL_ARGS_FIELD_TAG) {
+    if (UCC_COLL_ARGS_ACTIVE_SET(&coll_args->args)) {
+        task->subset.map    = coll_args->args.active_set.map;
+        task->subset.myrank = (ucc_rank_t)coll_args->args.active_set.my_ep;
+        /* ActiveSet of size 2 can use a dedicated bit in the tag space.
+           Otherwise we need to generate a uniq ID for an Active Set */
+        ucc_assert(coll_args->args.active_set.map.ep_num == 2);
+        task->tagged.tag = UCC_TL_UCP_MAX_TAG;
```
wesbland, line 197: I don't think you want to always set this to `UCC_TL_UCP_MAX_TAG`. Shouldn't this just set the high order bit on its own? If the user is providing their own tag, we want to honor that. So the check for an active set would be orthogonal to the check for `UCC_COLL_ARGS_FIELD_TAG` (i.e., not in the same if/else block).

wesbland (commented):
Looks good to me in general. Just a couple of specific questions.

src/components/tl/nccl/tl_nccl_coll.h
```diff
@@ -44,6 +44,10 @@ ucc_status_t ucc_tl_nccl_alltoallv_init(ucc_tl_nccl_task_t *task);
 
 ucc_status_t ucc_tl_nccl_bcast_init(ucc_tl_nccl_task_t *task);
 
+ucc_status_t ucc_tl_nccl_send_init(ucc_tl_nccl_task_t *task);
+
+ucc_status_t ucc_tl_nccl_recv_init(ucc_tl_nccl_task_t *task);
+
```
wesbland, line 50: I think these can be dropped. They were part of my PR that we don't need anymore.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -208,18 +208,35 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
...
     ucc_coll_task_init(&task->super, coll_args, team);
     if (coll_args->mask & UCC_COLL_ARGS_FIELD_TAG) {
-        task->tagged.tag           = task->tagged.tag;
-    } else if (coll_args->args.coll_type == UCC_COLL_TYPE_SEND ||
-               coll_args->args.coll_type == UCC_COLL_TYPE_RECV) {
-        /* Use the maximum tag with the point-to-point bit set as a reserved tag
-         * for point-to-point messages that do not otherwies have a tag set. */
-        task->tagged.tag           = UCC_TL_UCP_MAX_TAG;
+        task->tagged.tag = coll_args->args.tag;
+    } else if (UCC_COLL_ARGS_ACTIVE_SET(&coll_args->args)) {
```
wesbland, line 218: Shouldn't this be an `if` instead of an `else if` since you can have both `UCC_COLL_ARGS_FIELD_TAG` on and use an active set at the same time?

Or are we saying that you can't use both? I see from the code in `src/components/tl/ucp/tl_ucp_tag.h` below that `UCC_TL_UCP_ACTIVE_SET_2_TAG` is set to a specific tag instead of a specific bit.

src/components/tl/ucp/tl_ucp_tag.h
```diff
@@ -39,16 +39,19 @@
 #define UCC_TL_UCP_SCOPE_BITS_OFFSET                                        \
     (UCC_TL_UCP_ID_BITS + UCC_TL_UCP_SENDER_BITS + UCC_TL_UCP_SCOPE_ID_BITS)
 
-#define UCC_TL_UCP_SCOPE_ID_BITS_OFFSET (UCC_TL_UCP_ID_BITS + UCC_TL_UCP_SENDER_BITS)
+#define UCC_TL_UCP_SCOPE_ID_BITS_OFFSET (UCC_TL_UCP_ID_BITS + \
+                                         UCC_TL_UCP_SENDER_BITS)
```
wesbland, line 43: This is a code formatting only change and should be a separate patch (or dropped).

src/components/tl/ucp/tl_ucp_tag.h
```diff
@@ -39,16 +39,19 @@
...
-#define UCC_TL_UCP_MAX_TAG       UCC_MASK(UCC_TL_UCP_TAG_BITS)
-#define UCC_TL_UCP_RESERVED_TAGS 8
-#define UCC_TL_UCP_MAX_COLL_TAG  (UCC_TL_UCP_MAX_TAG - UCC_TL_UCP_RESERVED_TAGS)
-#define UCC_TL_UCP_SERVICE_TAG   (UCC_TL_UCP_MAX_COLL_TAG + 1)
-#define UCC_TL_UCP_MAX_SENDER    UCC_MASK(UCC_TL_UCP_SENDER_BITS)
-#define UCC_TL_UCP_MAX_ID        UCC_MASK(UCC_TL_UCP_ID_BITS)
+#define UCC_TL_UCP_MAX_TAG           UCC_MASK(UCC_TL_UCP_TAG_BITS)
+#define UCC_TL_UCP_RESERVED_TAGS     8
+#define UCC_TL_UCP_MAX_COLL_TAG     (UCC_TL_UCP_MAX_TAG - \
+                                     UCC_TL_UCP_RESERVED_TAGS)
```
wesbland, line 50: This is a code formatting only change and should be a separate patch (or dropped).

src/components/tl/ucp/tl_ucp_tag.h
```diff
@@ -39,16 +39,19 @@
...
+#define UCC_TL_UCP_SERVICE_TAG      (UCC_TL_UCP_MAX_COLL_TAG + 1)
+#define UCC_TL_UCP_ACTIVE_SET_2_TAG (UCC_TL_UCP_MAX_COLL_TAG + 2)
```
wesbland, line 52: Why is this named `UCC_TL_UCP_ACTIVE_SET_2_TAG` instead of `UCC_TL_UCP_ACTIVE_SET_TAG`?

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -419,14 +419,42 @@ ucc_status_t ucc_tl_nccl_bcast_start(ucc_coll_task_t *coll_task)
...
 
     dt = ucc_to_nccl_dtype[UCC_DT_PREDEFINED_ID(args->src.info.datatype)];
     task->super.status = UCC_INPROGRESS;
     UCC_TL_NCCL_PROFILE_REQUEST_EVENT(coll_task, "nccl_bcast_start", 0);
-    NCCLCHECK_GOTO(ncclBroadcast(src, src, count, dt, root, team->nccl_comm,
-                                 stream),
-                   exit_coll, status, UCC_TL_TEAM_LIB(team));
+
+    if (UCC_COLL_ARGS_ACTIVE_SET(args)) {
+        map  = ucc_active_set_to_ep_map(args);
```
wesbland, line 431: I guess this is fine for now since `size` will always be 2, but this is a a bottleneck in the general case.

wesbland:
I might suggest updating the Gist since the API changed slightly.

src/components/tl/nccl/tl_nccl_coll.c
vspetrov: it is. This is what i meant during WG, when i talked about adding "my_ep (my_rank)" field as part of active set. I said, that w/o it i need O(active_set_size) loop to compute it) @manjugv fyi

src/components/tl/ucp/tl_ucp_coll.h
vspetrov: you are right.

src/components/tl/ucp/tl_ucp_tag.h
vspetrov: added "2" to reflect that this all works for size 2. but after another look i agree lets just call it active_set_tag

vspetrov:
> I might suggest updating the Gist since the API changed slightly.

done

src/components/tl/nccl/tl_nccl_coll.c
wesbland: I'm not sure what the best solution is here. I think it's weird either way. It's not obvious what should go in this field to the user (you set it to `0` for both processes in your example), but it hurts performance without it...

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -207,9 +208,32 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
...
+    if (coll_args->mask & UCC_COLL_ARGS_FIELD_TAG) {
+        task->tagged.tag = coll_args->args.tag;
+    } else {
+        tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
+        task->tagged.tag = tl_team->seq_num;
+    }
+
+    if (UCC_COLL_ARGS_ACTIVE_SET(&coll_args->args)) {
+        task->tagged.tag = (coll_args->mask & UCC_COLL_ARGS_FIELD_TAG)
+            ? coll_args->args.tag : UCC_TL_UCP_ACTIVE_SET_TAG;
```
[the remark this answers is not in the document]
vspetrov: i don't get it a bit.
In you original code change (Honor user tag - that re-used) you add the "user tag bit" in the send/recv functions by checking mask & COLL_ARGS_FIELD_TAG.

So we have these options:
1. No user tags is provided, then we don't set USER_BIT_FLAG but use pre-defined UCC_TL_UCP_ACTIVE_SET_TAG. This way user can only use the api with "init & post ORDERED" semantics - which is common case. Then the matching of sends and recvs is done based on order (they all have same tag). The active set data is separated since UCC_TL_UCP_ACTIVE_SET_TAG is beyond the range of autogenerated (seq_num) tags.
2. User provides tag. Then we just use it. (the bit will be set in send/recv inlines)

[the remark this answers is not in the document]
wesbland: How does UCC tell the difference between a collective that is using active sets and custom tags an a non-active set collective?

Doesn't this seem the same to it?
```c
Bcast(tag=0, active_set={0,1})
Bcast(tag=NULL, active_set=NULL)
// Assumes the implicit tags that we use for ordering start at `0`
```

src/utils/ucc_coll_utils.h
```diff
@@ -246,4 +250,49 @@ static inline size_t ucc_buffer_block_offset(size_t     total_count,
...
+            if (rank == ucc_ep_map_eval(map, i)) {
+                local_rank = i;
+                break;
+            }
+        }
+    default:
+        break;
+    }
+    return local_rank;
+}
```
wesbland, line 285: I think this code breaks if your stride is negative. For example:

```
0 Recv from 1 -> start = 1, stride = -1
1 Send to   0 -> start = 1, stride = -1
```

Similar problems when evaluating the rank in the map in `ucc_ep_map_eval`.

The simplest solution is probably to stash the team size in the map below so it can be referenced when needed.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -207,9 +208,32 @@ ucc_tl_ucp_init_task(ucc_base_coll_args_t *coll_args, ucc_base_team_t *team)
...
+    }
+
     ucc_coll_task_init(&task->super, coll_args, team);
-    tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
-    task->tagged.tag           = tl_team->seq_num;
+    if (coll_args->mask & UCC_COLL_ARGS_FIELD_TAG) {
+        task->tagged.tag = coll_args->args.tag;
+    } else {
+        tl_team->seq_num = (tl_team->seq_num + 1) % UCC_TL_UCP_MAX_COLL_TAG;
+        task->tagged.tag = tl_team->seq_num;
```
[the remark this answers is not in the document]
vspetrov: oh, shoot. of course. it used to be correct in the former version. we should only use seq_nums if NO active_set is present

src/utils/ucc_coll_utils.h
vspetrov: i don't see a problem with negative strides. In fact, it tried to add the support (and test in the gtest) for negative stride. Gtest covers those cases. Also i tried with ucc_sendrecv.cc from gist (check my comment there).
Regarding your example:
start = 1, stride - 1, so active set is {1, 0}
Then the code above would compute:
ucc_ep_map_local_rank for rank 0:
(0 - 1)/(-1) = 1
for rank 1:
(1 - 1)/(-1) = 0
looks correct

to do recv rank 0 posts bcast(active_set={1, 0}, root = 1); to do the corresponding send rank 1 posts bcast(active_set{1, 0}, root = 1);

The difference with my first proposal (where we had "my_ep" as part of active_set) is that now both ranks MUST provide the SAME active_set triplet.

subset of ranks {2,3} from a team of 5 ranks can be given 2 ways: start = 2, stride=1 OR start  = 3, stride - 1. In the former proposal we provided "my_ep"  (which implied the ordering of a set, ie local_set_ranks) and in that case it didn't matter if 2 procs would provided different definitions of active_set (2, 1) or (3, -1) - the correct local ranks were provided via my_ep. Now, we don't have it. The ordering is given by the start,stride It must be the same across active_set participants.

src/components/tl/ucp/tl_ucp_coll.h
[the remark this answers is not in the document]
wesbland: I'm not sure it's that simple either though. Then you'd potentially have a problem if you had overlapping active sets. For example, with a team of 3 processes with two collectives across {0,1} and {1,2}, you could potentially mis-match a message destined for process 1.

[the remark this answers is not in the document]
vspetrov: This fix only for seq_num: don't mess with default non-active set tags.
Regarding your example: recv on rank 1 a matched using tag which consists of:
user_tag_bit | user_tag | scope bits | scope_id bits | source_rank | team_id

In your example, the recvs will be matched based on different "source_rank".

src/utils/ucc_coll_utils.h
wesbland: I did find the problem. It came from me trying to back port these changes to an older internal version of this code. It's working now.

manjugv:
@vspetrov From the overhead perspective does it help to tell whether active sets are enabled/disabled during team creation. Currently,  AFAIT, during team creation, there is no knowledge of active sets.

wesbland:
> @vspetrov From the overhead perspective does it help to tell whether active sets are enabled/disabled during team creation. Currently, AFAIT, during team creation, there is no knowledge of active sets.

Probably not much since you'd still have to have a branch on the critical path to tell whether active sets are turned on or not. I think @vspetrov used the same implementation of the UCP TL bcast with active sets other than doing a quick translation of ranks.

manjugv (approved):

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -53,6 +53,7 @@ static ucc_status_t ucc_tl_ucp_team_preconnect(ucc_tl_ucp_team_t *team)
     if (!team->preconnect_task) {
         team->preconnect_task             = ucc_tl_ucp_get_task(team);
         team->preconnect_task->tagged.tag = 0;
+        team->preconnect_task->super.bargs.args.mask = 0;
```
Sergei-Lebedev, line 56: alignment

src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -71,12 +74,13 @@ ucc_tl_ucp_send_common(void *buffer, size_t msglen, ucc_memory_type_t mtype,
     ucc_status_t        status;
     ucp_ep_h            ep;
     ucp_tag_t           ucp_tag;
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
```
Sergei-Lebedev, line 77: initialized var goes first

src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -131,9 +135,12 @@ ucc_tl_ucp_recv_common(void *buffer, size_t msglen, ucc_memory_type_t mtype,
 {
     ucp_request_param_t req_param;
     ucp_tag_t           ucp_tag, ucp_tag_mask;
+    ucc_coll_args_t    *args = &TASK_ARGS(task);
```
Sergei-Lebedev, line 138: initialized var goes first

Sergei-Lebedev (approved):

PR #492, opened 2022-04-27
API: add float128 and float32(64,128)_complex dt

samnordmann:
## What
Add support for four new datatypes: float128, float32_complex, float64_complex, float128_complex.
- float32_complex and float64_complex are supported on HOST and CUDA. float128 and float128_complex are supported only on HOST.
- The new datatypes are supported by all collectives. float128 supported with the reduction ops: min, max, sum, prod, avg. float32(64,128)_complex supported with the reduction ops: sum, prod, avg.
- Update MPI_test and gtest with the new datatypes. Remark: the gtests are run with float128 and float128 only on mc_reduce tests.
- new datatypes not supported by nccl nor sharp TLs

PR #498, opened 2022-05-05
Gtest update

vspetrov:
## What
A set of gtest improvements:
- Switch to gtest release v1.10.x that has "GTEST_SKIP" macro
- Cleanup/refactor mc_reduce test suite: split cuda/host test types_ops enumerations. (this will be used in PR #492 )
- Add checks for availability of mc cuda component at runtime and proper gtest_skip. This way we can launch gtest on systems with and w/o cuda w/o failures (that we might have in current master)


vspetrov:
bot:retest

vspetrov:
@Sergei-Lebedev @bureddy @manjugv plz review this PR. It is large but simple. All the changes in gtest.h and gtest-all.cc (which are biggest part of the PR) can be ignored in review, since this is just an update of the googletest to the official v1.10 release.

All the rest is simple: mostly refactoring of reduction tests - making groups of type+op pairs for cuda/host cases; and adding proper GTEST_SKIP in the unsupported cases. 

1 failure in openucx.ucc azure check should also be ignored - there is timeout in clang-format check. apparently 15K lines of code too much for git-clang-format to handle in 60mins.

test/gtest/core/test_mc_reduce.h
```diff
@@ -112,256 +66,83 @@ struct ReductionTest<UCC_DT_BFLOAT16, op> {
...
+        TypeOpPair<UCC_DT_FLOAT32, avg>,                \
+        FLOAT_OP_PAIRS(INT32),                          \
+        FLOAT_OP_PAIRS(FLOAT32),                        \
+        FLOAT_OP_PAIRS(FLOAT64),                        \
+        FLOAT_OP_PAIRS(BFLOAT16)
+
+using CollReduceTypeOpsCuda = ::testing::Types<CUDA_OP_PAIRS>;
+
+/* Host supports all combos, so add more to tests */
+using CollReduceTypeOpsHost = ::testing::Types<CUDA_OP_PAIRS,
```
Sergei-Lebedev, line 112: some pairs are missing compared to original version e.g. UCC_DT_INT16 max

test/gtest/core/test_mc_reduce.cc
```diff
@@ -0,0 +1,274 @@
...
+        ucc_status_t status;
+
+        if (UCC_OK !=  ucc_mc_available(mt)) {
+            GTEST_SKIP();
+        }
+        ASSERT_EQ(this->alloc_bufs(mt, num_vec), UCC_OK);
+        status = ucc_mc_reduce_multi(this->buf1, this->buf2, this->res,
+                       num_vec, this->COUNT, this->COUNT*sizeof(*this->buf2),
+                       T::dt, T::redop, mt);
+        if (UCC_ERR_NOT_SUPPORTED == status) {
```
Sergei-Lebedev, line 137: we make pairs so that reduction is always supported then this check is not needed

test/gtest/common/test_ucc.h
```diff
@@ -237,6 +248,16 @@ class UccReq {
...
+        if ((_UccReq).status == UCC_ERR_NOT_SUPPORTED) {    \
+            _action;                                        \
+            GTEST_SKIP();                                   \
+        }                                                   \
+        ASSERT_EQ(UCC_OK, (_UccReq).status);                \
+    } while(0)
+
 void clear_buffer(void *_buf, size_t size, ucc_memory_type_t mt, uint8_t value);
 
 #define PREDEFINED_DTYPES \
```
Sergei-Lebedev, line 263: missing UCC_DT_BFLOAT16

test/gtest/common/test_ucc.cc
```diff
@@ -563,26 +563,32 @@ UccReq::UccReq(UccTeam_h _team, ucc_coll_args_t *args) :
...
     }
-    if (UCC_OK != err_status) {
-        goto err;
-    }
-    return;
-err:
-    reqs.clear();
-    if (err_status == UCC_ERR_NOT_SUPPORTED) {
-        UCC_TEST_SKIP;
+    if (err_st.size() > 0) {
```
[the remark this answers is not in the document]
vspetrov: Destructor of UCC_REQ is called later and it takes care of reqs finalize

test/gtest/common/test_ucc.cc
```diff
@@ -607,26 +614,26 @@ void UccReq::start()
 
 ucc_status_t UccReq::test()
 {
-    ucc_status_t status = UCC_OK;
+    ucc_status_t st = UCC_OK;
     for (auto r : reqs) {
-        status = ucc_collective_test(r);
-        if (UCC_OK != status && UCC_OPERATION_INITIALIZED != status) {
+        st = ucc_collective_test(r);
+        if (UCC_OK != st && UCC_OPERATION_INITIALIZED != st) {
```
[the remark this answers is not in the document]
vspetrov: it was there in "master" - will remove

test/gtest/core/test_mc_reduce.h
vspetrov: I know. We never had ALL the combinations (to limit the time spent in collectives with reductions). I changed it that way: we now test ALL possible combos with test_mc_reduce, AND we test  "every possible dtype" and "every possible op" in the reduce coll tests (reduce, allreduce, reduce_scatter), but now every possible combination of them. 
Also i added explicitly INT32, FLOAT32, FLOAT64 with all arithmetic ops since those are most commonly used. If you think we should add more plz comment - i will extend.

test/gtest/common/test_ucc.h
vspetrov: PREDEFINED_TYPES macro is only used with collectives that do not do any reductions. Do you think we need to add?

test/gtest/core/test_mc_reduce.cc
vspetrov: For test_mc_reduce (not collective but just reduction backend) i don't enumerate separate dtypes for CUDA. I have the enumerations for ALL possible pairs, but skip not-supported. I could create separate set of testing::types for CUDA only. It will be little bit more explicit but then i can remove test skip here. What do you prefer?

test/gtest/common/test_ucc.h
Sergei-Lebedev: yes, why not, we anyway test different datatypes and I don't think we test bfloat16 copy operation somewhere else

test/gtest/core/test_mc_reduce.cc
Sergei-Lebedev: ok, keep it as is pls, i thought we support all ops and dtypes in cuda

test/gtest/common/test_ucc.h
vspetrov: added

bureddy (approved):

manjugv (approved):

vspetrov:
Merged w/o codestyle/openucx.ucc pass - both time out on the code style format - too large change in gtest.h/gtest_all.cc.

PR #517, opened 2022-05-30
(tl_mlx5_3) BUILD: verbs/dv detection

vspetrov:
## What
Adds configure steps to detect ibverbs/mlx5dv availability.

## Why ?
Used to build tl/mlx5

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -148,6 +148,8 @@ AS_IF([test "x$with_docs_only" = xyes],
      AM_CONDITIONAL([HAVE_MPICC], [false])
      AM_CONDITIONAL([HAVE_MPICXX], [false])
      AM_CONDITIONAL([HAVE_PROFILING],[false])
+     AM_CONDITIONAL([HAVE_IBVERBS],[false])
+     AM_CONDITIONAL([HAVE_MLX5DV],[false])
     ],
     [
      AM_CONDITIONAL([DOCS_ONLY], [false])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -161,6 +161,8 @@ AS_IF([test "x$with_docs_only" = xyes],
      m4_include([config/m4/rccl.m4])
      m4_include([config/m4/sharp.m4])
      m4_include([config/m4/mpi.m4])
+     m4_include([config/m4/ibverbs.m4])
+     m4_include([config/m4/mlx5dv.m4])
```
bureddy, line 165: why do we have to separate config for ib and mlx5dv? I would expect ib and mlx5dv headers in the same install path

config/m4/mlx5dv.m4
```diff
@@ -0,0 +1,87 @@
...
+                AC_MSG_RESULT([MLX5DV dir: $check_mlx5dv_dir])
+                AC_SUBST(MLX5DV_CPPFLAGS, "-I$check_mlx5dv_dir/include/")
+            ])
+
+            AS_IF([test "x$check_mlx5dv_libdir" != "x"],
+            [
+                AC_SUBST(MLX5DV_LDFLAGS, "-L$check_mlx5dv_libdir")
+            ])
+
+            AC_SUBST(MLX5DV_LIBADD, "-lmlx5dv")
```
bureddy, line 64: is there such library. isn't it part of libibverbs?

bureddy (approved):

manjugv (approved):

PR #518, opened 2022-05-30
(tl_mlx5_4) UTIL: rcache_get_arg

vspetrov:
## What
Extended version of ucc_rcache_get that supports in/out "args" parameter. 

## Why ?
Used in tl/mlx5 to indicate if rache entry was modified (i.e., if actual memory dereg/reg happened). In this case tl/mlx5 needs to update umr mkeys.

### DIFF
--- a/src/utils/ucc_rcache.h
+++ b/src/utils/ucc_rcache.h
@@ -29,13 +29,16 @@ ucc_rcache_create(const ucc_rcache_params_t *params,
                                         params, name, NULL, rcache_p));
 }
 
+/* [arg] parameter allows passing additional information from mem_reg callabck.
+   For example, it can be used to indicate whether the entry was found in the
+   cache or new registration happened. */
 static inline ucc_status_t
 ucc_rcache_get(ucc_rcache_t *rcache, void *address, size_t length,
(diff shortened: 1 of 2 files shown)


manjugv (approved):

src/utils/ucc_rcache.h
```diff
@@ -38,4 +38,16 @@ ucc_rcache_get(ucc_rcache_t *rcache, void *address, size_t length,
...
+   whether the entry was found in the cache or new registration happened. */
+static inline ucc_status_t
+ucc_rcache_get_arg(ucc_rcache_t *rcache, void *address, size_t length,
+                   void *arg, ucc_rcache_region_t **region_p)
+{
+    return ucs_status_to_ucc_status(ucs_rcache_get(
+                                       rcache, address, length,
+                                       PROT_READ | PROT_WRITE, arg, region_p));
+}
+
```
bureddy, line 52: can't we add arg to ucc_rcache_get() now?

vspetrov: yes, done

bureddy (approved):

PR #525, opened 2022-05-31
UTIL: mpool init fix

vspetrov:
## What
Fixes data corruption with the new mpool interface.
For small elem_sizes (e.g. in mc/cuda elem size is sizeof(ucc_mc_cuda_mpool_header_t) = 16b) setting max_chunk_size = elems_per_chunk * elem_size just seems correct. But elem_size is aligned up internally in UCS. this makes elem_size  > max_chunk_size which is not correctly handled in UCS. Setting grow_factor = 1 and max_chunk_size = SIZE_MAX fixes it and makes behaviour similar to legacy api.


Sergei-Lebedev (approved):

manjugv (approved):

vspetrov:
bot:retest

PR #545, opened 2022-06-22
UTIL: cfg parser respect inherited vars

vspetrov:
## What
Fixes config file parsing with respect to inherited variables

## Why ?
When the variable in the config table is inherited from another parent table (e.g., TLS var in CLs is defined in parent ucc_cl_base_config), and user sets the top variable in the env it gets overwritten by the file configuration.
For example,
if there is cfg file with entry:
UCC_CL_BASIC_TLS=^sharp

and user runs app with: UCC_TLS=ucp , then the value of CL_BASIC_TLS would still be "^sharp", i.e. all except sharp - not what user expects (which is "ucp" only).

## How ?
When applying value from cfg file database need to check if there is a parent variable setting in the environment and then skip it in that case.

### DIFF
--- a/src/utils/ucc_parser.c
+++ b/src/utils/ucc_parser.c
@@ -5,6 +5,7 @@
 
 #include "ucc_parser.h"
 #include "ucc_malloc.h"
+#include "ucc_string.h"
 #include "ucc_log.h"
 #include "khash.h"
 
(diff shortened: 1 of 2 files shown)


bureddy:
if UCC_CL_BASIC_TLS is not in environment, then is it then same UCS_TLS

vspetrov:
> if UCC_CL_BASIC_TLS is not inthe environment, then is it the same as UCS_TLS

@bureddy i don't understand. could you plz elaborate?

bureddy:
> > if UCC_CL_BASIC_TLS is not inthe environment, then is it the same as UCC_TLS
> 
> @bureddy i don't understand. could you plz elaborate?

I mean if the user sets UCC_TL but not UCC_CL_BASIC_TLS. does it alway inherit from UCC_TLS ?

vspetrov:
> > > if UCC_CL_BASIC_TLS is not inthe environment, then is it the same as UCC_TLS
> > 
> > 
> > @bureddy i don't understand. could you plz elaborate?
> 
> I mean if the user sets UCC_TL but not UCC_CL_BASIC_TLS. does it alway inherit from UCC_TLS ?

This was the behavior before file cfg, which broke it. UCC_TLS is the highest var in the hierarchy, all CLs vars inherit from it. W/o config file, we can set UCC_TLS=ucp and this will guarantee all CLs only have ucp. 

Say, i want to disable TL/SHM. It is used in both CLs hier and basic. I could just set UCC_TLS=^shm. But currently due to bug in cfg file, it will not work.

manjugv:
@vspetrov Please add a FAQ entry to capture this example.

manjugv (approved):

janjust:
stale, opened issue since we need to fix this

PR #546, opened 2022-06-22
TL/CUDA: check cuda device in coll init

Sergei-Lebedev:
## What
Check GPU in collective_init matches device used to initialize team. Set GPU device in multithread test

## Why ?
Fixes segfault in multithread tests with TL CUDA

### DIFF
--- a/test/mpi/test_mpi.cc
+++ b/test/mpi/test_mpi.cc
@@ -397,6 +397,10 @@ void UccTestMpi::set_displ_vsizes(std::vector<ucc_test_vsize_flag_t> &_displs_vs
 }
 
 #if defined(HAVE_CUDA) || defined(HAVE_HIP)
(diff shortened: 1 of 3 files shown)


test/mpi/test_mpi.cc
```diff
@@ -623,6 +627,10 @@ typedef struct ucc_test_thread {
 static void *thread_start(void *arg)
 {
     ucc_test_thread_t *t = (ucc_test_thread_t *)arg;
+
+    #if defined(HAVE_CUDA) || defined(HAVE_HIP)
+        set_gpu_device(test_gpu_set_device);
```
vspetrov, line 632: is it aligned?

vspetrov (approved):

test/mpi/test_mpi.cc
Sergei-Lebedev: fixed

manjugv (approved):

PR #547, opened 2022-06-22
TL/RCCL: fix configure logic

edgargabriel:
## What
allow to explicitly disable tl/rccl 

## Why ?
If rccl is installed as part of the rocm software package, the configure script picks it up (correctly) even if the user did not explicitly request it. However, if not desired, the  ```--with-rccl=no``` option did not work correctly due to a minor mistake in the configure script.


edgargabriel:
@vspetrov  To follow up on the discussion that we had today on the call, I tested whether tl/nccl has the same issue as tl/rccl w.r.t to explicitly disabling it, and I can confirm that it does *not* have that problem, it is tl/rccl only. Using ```--with-nccl=no``` works as expected based on my test. This was a bug that creeped into the source code with our initial commit,.

vspetrov (approved):

Sergei-Lebedev (approved):

PR #551, opened 2022-06-24
TL/RCCL and TL/NCCL: move event_destroy call

edgargabriel:
## What
move the invocation of event_destroy() from the coll_finalize() function to the free_task() function .

## Why ?
Running ucc_test_mpi generated warning about event objects not being returned to the mpool. This commit fixes the problem by making sure that the events are correctly managed even if a test is skipped due to an unsupported datatype.


Sergei-Lebedev (approved):

Sergei-Lebedev:
@edgargabriel pls rebase this PR

PR #556, opened 2022-07-04
TL/CUDA: alltoall dst buf registration

Sergei-Lebedev:
## What
Reworked TL/CUDA proxy storage. Each rank keeps list of ranks where this rank is proxy.

## Why ?
In TL/CUDA alltoall/alltoallv:
* No need to go over list of proxy ranks to post copy
* Destination buffer registered only if required

### DIFF
--- a/src/components/tl/cuda/tl_cuda_team_topo.h
+++ b/src/components/tl/cuda/tl_cuda_team_topo.h
@@ -11,22 +11,23 @@
 #include "tl_cuda_topo.h"
 
 typedef struct ucc_tl_cuda_proxy {
-    ucc_rank_t src;
-    ucc_rank_t dst;
-    ucc_rank_t proxy;
+    ucc_rank_t src;   /* source rank */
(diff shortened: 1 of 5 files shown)


vspetrov (approved):

src/components/tl/cuda/tl_cuda_team_topo.h
```diff
@@ -11,31 +11,32 @@
...
 } ucc_tl_cuda_ring_t;
 
 typedef struct ucc_tl_cuda_team_topo {
-    int                     *matrix;
-    int                      num_proxies;
-    ucc_tl_cuda_proxy_t     *proxies;
-    int                      num_rings;
-    ucc_tl_cuda_ring_t      *rings;
+    int                     *matrix;       /* nvlink adjacency matrix */
+    int                      proxy_needed; /* is proxy needed for current rank */
```
manjugv, line 26: Minor - bool?

manjugv (approved):

PR #557, opened 2022-07-06
TL/UCP: fix executor status check

kingchc:
## What
bug fix of TL/UCP when using executor for GPU collectives

## Why ?
`ucc_ee_executor_task_test` may return `UCC_INPROGRESS ` or `UCC_OPERATION_INITIALIZED` when it is not ready. TL/UCP only checks `UCC_INPROGRESS` may cause false results. 

## How ?
Let it return with task status of `UCC_INPROGRESS` as long as it's not `UCC_OK` or failures (i.e., `status > 0`)


kingchc:
cc @Sergei-Lebedev @vspetrov

Sergei-Lebedev (approved):

Sergei-Lebedev:
ok to test

manjugv (approved):

PR #560, opened 2022-07-06
TL/UCP: add special service worker

samnordmann:
## What
Enables the creation and usage of a special ucp worker for service collectives.

- This feature is enabled by setting the env variable `UCC_TL_UCP_SERVICE_WORKER` to 1 (default is 0).
- Any UCX configuration flag for the service worker can be passed with the prefix `UCC_TL_UCP_SERVICE_`. For example, `UCX_TL` and `UCX_NET_DEVICES` of the service worker can be set through the env variables `UCC_TL_UCP_SERVICE_TLS` and `UCC_TL_UCP_SERVICE_NET_DEVICES` respectively.
- The env variable `UCC_TL_UCP_SERVICE_THROTTLING_THRESH` (default is 100) sets the number of call to ucc_context_progress function between two consecutive calls to service worker progress function.

## How ?
- The special worker is created at context initialization. A new struct `ucc_tl_ucp_worker` has been defined and stores all the relevant data for ucp workers. The context `ucc_tl_ucp_context` now contains two workers, the default `worker` and the special `service_worker`.
-  If the feature is disabled, then no worker is created, therefore this new feature causes no overhead to the default use case. 
- the team `ucc_tl_ucp_team` has been supplemented with a pointer `*ucc_tl_ucp_worker` which points to the default `worker` or `service_worker` accordingly.
- `ucc_tl_ucp_get_context_attr` has been modified to include the address exchange for the service worker during wiring up
- If the service worker is used, then it is progressed once every `cfg.service_throttling_thresh=100` times the default worker is progressed
- Context init and cleanup has been refactorized to avoid code duplication.

### DIFF
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -175,11 +175,24 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, pre_reg_mem),
      UCC_CONFIG_TYPE_UINT},
 
-    {"SERVICE_TLS", "",
-     "If empty string (default), uses the same worker for collectives and "
+    {"SERVICE_WORKER", "0",
+     "If set to 0, uses the same worker for collectives and "
      "service. If not, creates a special worker for service collectives "
-     "for which UCX_TL configuration is set to that string",
+     "for which UCX_TL and UCX_NET_DEVICES are configured by the variables "
+     "UCC_TL_UCP_SERVICE_TLS and UCC_TL_UCP_SERVICE_NET_DEVICES respectively",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, service_worker),
+     UCC_CONFIG_TYPE_UINT},
+
+    {"SERVICE_TLS", "",
+     "If not an empty string and UCC_TL_UCP_SERVICE_WORKER!=0, sets the "
+     "UCX_TL configuration of the special service worker to that string",
      ucc_offsetof(ucc_tl_ucp_context_config_t, service_tls),
+     UCC_CONFIG_TYPE_STRING},
+
+    {"SERVICE_NET_DEVICES", "",
+     "If not an empty string and UCC_TL_UCP_SERVICE_WORKER!=0, sets the "
+     "UCX_NET_DEVICES configuration of the service worker to that string",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, service_devs),
      UCC_CONFIG_TYPE_STRING},
 
     {NULL}};
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -175,31 +175,19 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, pre_reg_mem),
      UCC_CONFIG_TYPE_UINT},
 
-    {"SERVICE_WORKER", "0",
+    {"SERVICE_WORKER", "n",
      "If set to 0, uses the same worker for collectives and "
      "service. If not, creates a special worker for service collectives "
      "for which UCX_TL and UCX_NET_DEVICES are configured by the variables "
      "UCC_TL_UCP_SERVICE_TLS and UCC_TL_UCP_SERVICE_NET_DEVICES respectively",
      ucc_offsetof(ucc_tl_ucp_context_config_t, service_worker),
-     UCC_CONFIG_TYPE_UINT},
+     UCC_CONFIG_TYPE_BOOL},
 
     {"SERVICE_THROTTLING_THRESH", "100",
      "Number of call to ucc_context_progress function between two consecutive "
      "calls to service worker progress function",
      ucc_offsetof(ucc_tl_ucp_context_config_t, service_throttling_thresh),
      UCC_CONFIG_TYPE_UINT},
-
-    {"SERVICE_TLS", "",
-     "If not an empty string and UCC_TL_UCP_SERVICE_WORKER!=0, sets the "
-     "UCX_TL configuration of the special service worker to that string",
-     ucc_offsetof(ucc_tl_ucp_context_config_t, service_tls),
-     UCC_CONFIG_TYPE_STRING},
-
-    {"SERVICE_NET_DEVICES", "",
-     "If not an empty string and UCC_TL_UCP_SERVICE_WORKER!=0, sets the "
-     "UCX_NET_DEVICES configuration of the service worker to that string",
-     ucc_offsetof(ucc_tl_ucp_context_config_t, service_devs),
-     UCC_CONFIG_TYPE_STRING},
 
     {NULL}};
 
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -183,6 +183,12 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, service_worker),
      UCC_CONFIG_TYPE_UINT},
 
+    {"SERVICE_THROTTLING_THRESH", "100",
+     "Number of call to ucc_context_progress function between two consecutive "
+     "calls to service worker progress function",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, service_throttling_thresh),
+     UCC_CONFIG_TYPE_UINT},
+
     {"SERVICE_TLS", "",
      "If not an empty string and UCC_TL_UCP_SERVICE_WORKER!=0, sets the "
      "UCX_TL configuration of the special service worker to that string",
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -175,6 +175,13 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, pre_reg_mem),
      UCC_CONFIG_TYPE_UINT},
 
+    {"SERVICE_TLS", "",
+     "If empty string (default), uses the same worker for collectives and "
+     "service. If not, creates a special worker for service collectives "
+     "for which UCX_TL configuration is set to that string",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, service_tls),
+     UCC_CONFIG_TYPE_STRING},
+
     {NULL}};
 
 UCC_CLASS_DEFINE_NEW_FUNC(ucc_tl_ucp_lib_t, ucc_base_lib_t,
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -89,9 +89,11 @@ ucc_tl_ucp_context_service_init(const char *prefix, ucp_params_t ucp_params,
     }
     UCP_CHECK(ucp_config_read(service_prefix, NULL, &ucp_config),
               "failed to read ucp configuration", err_cfg_read, ctx);
+    ucc_free(service_prefix);
 
     UCP_CHECK(ucp_init(&ucp_params, ucp_config, &ucp_context_service),
               "failed to init ucp context for service worker", err_cfg, ctx);
+    ucp_config_release(ucp_config);
 
     UCP_CHECK(ucp_worker_create(ucp_context_service, &worker_params,
                                 &ucp_worker_service),
@@ -151,6 +153,7 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
     prefix[strlen(prefix) - 1] = '\0';
     UCP_CHECK(ucp_config_read(prefix, NULL, &ucp_config),
               "failed to read ucp configuration", err_cfg_read, self);
+    ucc_free(prefix);
 
     ucp_params.field_mask =
         UCP_PARAM_FIELD_FEATURES | UCP_PARAM_FIELD_TAG_SENDER_MASK;
@@ -172,6 +175,7 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_context_t,
 
     UCP_CHECK(ucp_init(&ucp_params, ucp_config, &ucp_context),
               "failed to init ucp context", err_cfg, self);
+    ucp_config_release(ucp_config);
 
     context_attr.field_mask = UCP_ATTR_FIELD_MEMORY_TYPES;
     UCP_CHECK(ucp_context_query(ucp_context, &context_attr),
--- a/src/components/tl/ucp/tl_ucp_context.c
+++ b/src/components/tl/ucp/tl_ucp_context.c
@@ -60,49 +60,40 @@ static inline ucc_status_t ucc_tl_ucp_context_service_init(
     ucp_worker_h  ucp_worker_service;
     ucs_status_t  status;
 
-    if (ctx->cfg.service_worker != 0) {
-        ctx->service.is_used = 1;
-        if (*ctx->cfg.service_tls) {
-            UCP_CHECK(
-                ucp_config_modify(ucp_config, "TLS", ctx->cfg.service_tls),
-                "failed to set UCX_TLS for service worker", err_cfg, ctx);
-        }
-        if (*ctx->cfg.service_devs) {
-            UCP_CHECK(ucp_config_modify(ucp_config, "NET_DEVICES",
-                                        ctx->cfg.service_devs),
-                      "failed to set UCX_NET_DEVICES for service worker",
-                      err_cfg, ctx);
-        }
-        UCP_CHECK(ucp_init(&ucp_params, ucp_config, &ucp_context_service),
-                  "failed to init ucp context for service worker", err_cfg,
-                  ctx);
+    if (*ctx->cfg.service_tls) {
+        UCP_CHECK(ucp_config_modify(ucp_config, "TLS", ctx->cfg.service_tls),
+                  "failed to set UCX_TLS for service worker", err_cfg, ctx);
+    }
+    if (*ctx->cfg.service_devs) {
+        UCP_CHECK(
+            ucp_config_modify(ucp_config, "NET_DEVICES", ctx->cfg.service_devs),
+            "failed to set UCX_NET_DEVICES for service worker", err_cfg, ctx);
+    }
+    UCP_CHECK(ucp_init(&ucp_params, ucp_config, &ucp_context_service),
+              "failed to init ucp context for service worker", err_cfg, ctx);
 
-        UCP_CHECK(ucp_worker_create(ucp_context_service, &worker_params,
-                                    &ucp_worker_service),
-                  "failed to create ucp service worker", err_worker_create,
-                  ctx);
+    UCP_CHECK(ucp_worker_create(ucp_context_service, &worker_params,
+                                &ucp_worker_service),
+              "failed to create ucp service worker", err_worker_create, ctx);
 
-        ctx->service.ucp_context    = ucp_context_service;
-        ctx->service.ucp_worker     = ucp_worker_service;
-        ctx->service.worker_address = NULL;
+    ctx->service.ucp_context    = ucp_context_service;
+    ctx->service.ucp_worker     = ucp_worker_service;
+    ctx->service.worker_address = NULL;
 
-        CHECK(UCC_OK != ucc_context_progress_register(
-                            params->context,
-                            (ucc_context_progress_fn_t)ucp_worker_progress,
-                            ctx->service.ucp_worker),
-              "failed to register progress function for service worker",
-              err_thread_mode, UCC_ERR_NO_MESSAGE, ctx);
-
-        CHECK(
-            UCC_OK != ucc_tl_ucp_eps_ephash_init(params, ctx,
-                                                 &ctx->service.ep_hash,
-                                                 &ctx->service.eps),
-            "failed to allocate memory for endpoint storage for service worker",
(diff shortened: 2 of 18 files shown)


manjugv:
CC @kingchc

manjugv:
Ping @kingchc

kingchc (commented):
@samnordmann - thanks for preparing this PR, this looks good to me overall. Please see my comments regrading to allow passing one more env. variable (UCX_NET_DEVICES) to service team.

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -13,6 +13,77 @@
...
+    ucc_tl_ucp_context_t *ctx)
+{
+    ucc_status_t  ucc_status = UCC_OK;
+    ucp_context_h ucp_context_service;
+    ucp_worker_h  ucp_worker_service;
+    ucs_status_t  status;
+
+    if (*ctx->cfg.service_tls) {
+        ctx->service.is_used = 1;
+        ucp_config_modify(ucp_config, "TLS", ctx->cfg.service_tls);
```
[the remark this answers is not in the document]
samnordmann: done

src/components/tl/ucp/tl_ucp.c
```diff
@@ -175,6 +175,13 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, pre_reg_mem),
      UCC_CONFIG_TYPE_UINT},
 
+    {"SERVICE_TLS", "",
+     "If empty string (default), uses the same worker for collectives and "
+     "service. If not, creates a special worker for service collectives "
+     "for which UCX_TL configuration is set to that string",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, service_tls),
+     UCC_CONFIG_TYPE_STRING},
+
```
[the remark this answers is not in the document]
samnordmann: I added as you suggested the variable UCC_TL_UCP_SERVICE_NET_DEVICES to modify the field UCX_NET_DEVICES of the service worker. Lmk if passing all "UCC_TL_UCP_SERVICE_" to UCX would be useful

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -407,23 +525,40 @@ ucc_status_t ucc_tl_ucp_get_context_attr(const ucc_base_context_t *context,
                                          ucc_base_ctx_attr_t      *attr)
 {
     ucc_tl_ucp_context_t *ctx = ucc_derived_of(context, ucc_tl_ucp_context_t);
+    uint64_t *            offset = (uint64_t *)attr->attr.ctx_addr;
```
kingchc, line 528: minor: align `=`

kingchc (approved):
Looks good to me!

src/components/tl/ucp/tl_ucp.c
```diff
@@ -175,6 +175,20 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
...
+    {"SERVICE_NET_DEVICES", "",
```
vspetrov, line 185: i suggest we add another variable that explicitly enables alternative worker for service. Currently, this flow is triggered if SERVICE_TLS is set. Suppose, user wants a dedicated NET iface for service but default set of TLS (user might not know anything about UCX TLS). Then only setting SERVICE_NET_DEVICES would not work. Lets add "UCC_TL_UCP_SERVICE_WORKER=y/n". If that var is set we use the flow. SERVICE_TLS and SERVICE_NET_DEVICES are parsed as is. @kingchc wdyt?

src/schedule/ucc_schedule.h
```diff
@@ -104,6 +104,7 @@ typedef struct ucc_coll_task {
     double   start_time; /* timestamp of the start time:
                             either post or triggered_post */
     uint32_t seq_num;
+    int is_service; /* bool flag specifying if using a special service worker */
```
vspetrov, line 107: this is smth specific to TL/UCP - should not be in common core data structure imho.

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -13,6 +13,99 @@
 #include "schedule/ucc_schedule_pipelined.h"
 #include <limits.h>
 
+static inline ucc_status_t ucc_tl_ucp_context_service_init(
```
vspetrov, line 16: i would try to unify this function for both workers to reduce code duplication

src/components/tl/ucp/tl_ucp_sendrecv.h
```diff
@@ -152,8 +154,8 @@ ucc_tl_ucp_recv_common(void *buffer, size_t msglen, ucc_memory_type_t mtype,
     req_param.memory_type = ucc_memtype_to_ucs[mtype];
     req_param.user_data   = (void *)task;
     task->tagged.recv_posted++;
-    return ucp_tag_recv_nbx(UCC_TL_UCP_WORKER(team), buffer, 1, ucp_tag,
-                            ucp_tag_mask, &req_param);
+    return ucp_tag_recv_nbx(UCC_TL_UCP_TEAM_WORKER(team, task), buffer, 1,
```
Sergei-Lebedev, line 157: Since service team is a separate ucc team I would rather cached ucp worker on team i.e. service team worker points to service worker or regular worker depending on whether config flag is set or not. Then we can remove is_service from task and remove extra branch from UCC_TL_UCP_TEAM_WORKER

src/components/tl/ucp/tl_ucp.c
kingchc: Makes sense to be explicit. Can this be documented somewhere?

src/components/tl/ucp/tl_ucp_sendrecv.h
samnordmann: I am not sure to understand, what do you mean by "separate ucc team"? The service team is of type `ucc_tl_ucp_team_t` so the only way to point to the worker is first to point to the context which itself stores both ucp_worker and service.ucp_worker.
Unless we duplicate entirely the context for the service team (and replace the regular fields with the one from service), I don't know how to make the "team point to service worker" -- is it what you suggest, to create a special context for the service team?

src/components/tl/ucp/tl_ucp.c
samnordmann: Ok it makes sense. I implement it in a new separated commit.
About the documentation @vspetrov what do you suggest?

src/schedule/ucc_schedule.h
samnordmann: It has been removed, now `is_service` is deduced from the team

src/components/tl/ucp/tl_ucp_context.c
samnordmann: Sure. I did what I could in the separate commit `TL/UCP: refactor context init`. Lmk if you think this could be done better

src/components/tl/ucp/tl_ucp.c
vspetrov: > Ok it makes sense. I implement it in a new separated commit. About the documentation @vspetrov what do you suggest?

Parameter description will be available via ucc_info. Also let's add Wiki FAQ page: "custom service ucp worker"

src/components/tl/ucp/tl_ucp_sendrecv.h
vspetrov: @samnordmann i think what Sergey meant is smth like this:
say we add
```
typedef struct ucc_tl_ucp_worker {
    ucp_context_t ucp_context;
    ucp_worker_t  ucp_workerk;
    ....
    addr
    addrlen
    eps
    ep_hash
} ucc_tl_ucp_worker_t;
```

Then on context:
```
typedef struct ucc_tl_ucp_context {
   ...
    ucc_tl_ucp_worker_t worker;
    ucc_tl_ucp_worker_t service_worker;
  ...
} ucc_tl_ucp_context_t;
```

Team will have a pointer:
```
typedef struct ucc_tl_ucp_team {
   ...
   ucc_tl_ucp_worker_t *worker;
   ...
} ucc_tl_ucp_team_t;
```
Then during team creation:
```
ucc_tl_ucp_team_create_test () {
  ...
  if (TEAM_IS_SERVICE(team) && ctx->cfg.service_worker) {
     team->worker = &ctx->service_worker;
  } else {
     team->worker = &ctx->worker;
  }
}
```

In that case you never need to evaluate "is_service" condition ever again. you always just use team->worker.ucp_worker or team->worker.ep_hash, etc.

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -81,13 +95,13 @@ void ucc_tl_ucp_close_eps(ucc_tl_ucp_context_t *ctx)
...
      param.flags        = 0; // 0 means FLUSH
-     ep                 = get_next_ep_to_close(ctx, &i);
+     ep                 = get_next_ep_to_close(ep_hash, eps, ctx, &i);
      while (ep) {
          close_req = ucp_ep_close_nbx(ep, &param);
 
          if (UCS_PTR_IS_PTR(close_req)) {
              do {
-                 ucp_worker_progress(ctx->ucp_worker);
+                 ucp_worker_progress(worker);
```
vspetrov, line 104: i think here we should be progressing both workers if 2 are available, otherwise we can get a deadlock
This in general applies to all do {} while blocking loops where we progress ucp worker

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -13,6 +13,106 @@
...
+                  "failed to create ucp service worker", err_worker_create,
+                  ctx);
+
+        ctx->service.ucp_context    = ucp_context_service;
+        ctx->service.ucp_worker     = ucp_worker_service;
+        ctx->service.worker_address = NULL;
+
+        CHECK(UCC_OK != ucc_context_progress_register(
+                            params->context,
+                            (ucc_context_progress_fn_t)ucp_worker_progress,
```
vspetrov, line 91: I also suggest a small change/opt regarding progress. I think we should add 
ucc_tl_ucp_worker_progress function that looks smth like this:
```
{
        ucp_worker_prgoress(ctx->ucp_worker);
        if (ctx->service_worker) {
              int throttling_count = ucc_atomic_fadd32(ctx->service_worker_throttling_count, 1);
              if (throttling_count == ctx->cfg.service_worker_throuttling_count) {
                  ctx->service_worker_throttling_count = 0;
                  ucp_worker_prgoress(ctx->service_worker);
              }
        }
}
```

The idea is that ucp_worker_prgoress is very costly operation (multiple fn calls, cq polling, events polling etc) and we want to minimize its effect. Service team is used only in control path, so most of hte time we can progress it only every so often. We can add smth like "throttling parameter", say = 100. Then service_worker progress is called only once in 100 tl_ucp progresses

vspetrov (changes requested):

src/components/tl/ucp/tl_ucp_context.c
samnordmann: It is implemented in the commit [TL/UCP: add throttling param for progress fn](https://github.com/openucx/ucc/pull/560/commits/18c4a87658b8be9ea1c38e37aa0be7a16d1f159f)
The parameter is set by the env variable `UCC_TL_UCP_SERVICE_THROTTLING_THRESH`, default=100

src/components/tl/ucp/tl_ucp_ep.c
samnordmann: done, see [TL/UCP: fix deadlock w/ service & default workers](https://github.com/openucx/ucc/pull/560/commits/67f6a2b1c6cd318c0248da669d379565aa5f0618)

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -233,30 +312,46 @@ ucc_status_t ucc_tl_ucp_rinfo_destroy(ucc_tl_ucp_context_t *ctx)
...
-    ucc_context_progress_deregister(
-        self->super.super.ucc_context,
-        (ucc_context_progress_fn_t)ucp_worker_progress, self->ucp_worker);
-    if (self->worker_address) {
-        ucp_worker_release_address(self->ucp_worker, self->worker_address);
-    }
-    ucp_worker_destroy(self->ucp_worker);
     ucc_mpool_cleanup(&self->req_mp, 1);
-    ucp_cleanup(self->ucp_context);
+    ucc_tl_ucp_cleanup_worker(self->ucp_worker, self->ep_hash, self->eps,
```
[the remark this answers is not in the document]
samnordmann: done, see [TL/UCP: fix context cleanup](https://github.com/openucx/ucc/pull/560/commits/f8ce0851d6a971e01c9c972db49f3a1102e81276)

The error pointed out by @Sergei-Lebedev 

> warnings in linter seem to be relevant
> 
> ```
> [1661245989.981908] [fv-az72-584:44720:561]           flush.c:28   UCX  ERROR req 0x62e000e26080: error during flush: Endpoint timeout, flush comp 0x62e000e26110 count reduced to 2
> ```

disappeared, except for the gtests `test_team.team_create_multiple_preconnect_ctx_local` and `test_team.team_create_destroy_ctx_local/3`, but those are also present in upstream/master

src/components/tl/ucp/tl_ucp_sendrecv.h
samnordmann: done, [TL/UCP: add worker struct and ptr in team](https://github.com/openucx/ucc/pull/560/commits/8f45ade0fa9ce5c402214bde92057980b79527cd)

Note that the branching to determine whether the team is the service team still needs to be done and can 't be avoided in the function `ucc_tl_ucp_connect_team_ep` from `src/components/tl/ucp/tl_ucp_ep.c`

src/components/tl/ucp/tl_ucp.c
```diff
@@ -175,6 +175,32 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_context_config_t, pre_reg_mem),
      UCC_CONFIG_TYPE_UINT},
 
+    {"SERVICE_WORKER", "0",
```
vspetrov, line 178: @samnordmann @Sergei-Lebedev i've opened a PR against Sam's branch : https://github.com/samnordmann/ucc/pull/1. I think this is more convenient, and then we don't need to add these service variables to TL/UCP. wdyt?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -13,6 +13,109 @@
...
+    }
+
+unsigned ucc_tl_ucp_worker_progress(void *progress_arg)
+{
+    ucc_tl_ucp_context_t *ctx = (ucc_tl_ucp_context_t *)progress_arg;
+    unsigned              ret;
+
+    ret = ucp_worker_progress(ctx->worker.ucp_worker);
+
+    if (ctx->cfg.service_worker != 0) {
```
vspetrov, line 38: @samnordmann now when i look at this, i think we should 2 different functions: 1 to be used when service_worker is there (like this function here) and another one with default only. We can make the decision at ctx creation and register only the required fn, thus eliminating the branch

src/components/tl/ucp/tl_ucp.c
```diff
@@ -175,6 +175,32 @@ static ucs_config_field_t ucc_tl_ucp_context_config_table[] = {
...
+     "If set to 0, uses the same worker for collectives and "
+     "service. If not, creates a special worker for service collectives "
+     "for which UCX_TL and UCX_NET_DEVICES are configured by the variables "
+     "UCC_TL_UCP_SERVICE_TLS and UCC_TL_UCP_SERVICE_NET_DEVICES respectively",
+     ucc_offsetof(ucc_tl_ucp_context_config_t, service_worker),
+     UCC_CONFIG_TYPE_UINT},
```
vspetrov, line 184: UCC_CONFIG_TYPE_UINT -> UCC_CONFIG_TYPE_BOOL (the variable should get "y" or "n")

src/components/tl/ucp/tl_ucp_ep.c
```diff
@@ -43,35 +51,40 @@ ucc_status_t ucc_tl_ucp_connect_team_ep(ucc_tl_ucp_team_t *team,
...
+                              : TL_UCP_EP_ADDR_WORKER(addr);
+
+    return ucc_tl_ucp_connect_ep(ctx, use_service_worker, ep, addr);
 }
 
 /* Finds next non-NULL ep in the storage and returns that handle
    for closure. In case of "hash" storage it pops the item,
    in case of "array" sets it to NULL */
-static inline ucp_ep_h get_next_ep_to_close(ucc_tl_ucp_context_t *ctx, int *i)
+static inline ucp_ep_h get_next_ep_to_close(ucc_tl_ucp_worker_t   worker,
```
vspetrov, line 68: why to pass worker by value here? i would pass by ptr to avoid struct copying

src/components/tl/ucp/tl_ucp.c
samnordmann: I think it is great. I merge your PR to my branch

src/components/tl/ucp/tl_ucp.c
samnordmann: ok

src/components/tl/ucp/tl_ucp_ep.c
samnordmann: ok, done

vspetrov (approved):

src/components/tl/ucp/tl_ucp_ep.h
```diff
@@ -70,10 +80,10 @@ static inline ucc_status_t ucc_tl_ucp_get_ep(ucc_tl_ucp_team_t *team, ucc_rank_t
             *ep = NULL;
             return status;
         }
-        if (ctx->eps) {
-            ctx->eps[ctx_rank] = *ep;
+        if (team->worker->eps) {
+            team->worker->eps[ctx_rank] = *ep;
         } else {
-            tl_ucp_hash_put(ctx->ep_hash, h->ctx_id, *ep);
+            tl_ucp_hash_put(team->worker->ep_hash, h->ctx_id, *ep);
```
vspetrov, line 86: @samnordmann plz see linter warnings here: https://github.com/openucx/ucc/actions/runs/3082772787/jobs/4982882880

here linter has a false positive: it assumes "h" can be NULL at line 86. Either use annotation to suppress or try to rewrite the same code a bit (maybe try "if (!h) " instead of "if (team->worker->eps)" at line 83).

Another warning is tl_ucp_context.c line 275: ucc_free(prefix) - double free. This is my fault (the PR we merged into your branch) - should be a simple fix (e.g. set prefix ot NULL after free and check in the end)

samnordmann: done. Btw I fixed another issue: `prefix `was freed and then passed as an argument to `ucc_tl_ucp_context_service_init `

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -96,6 +96,12 @@ ucc_status_t ucc_tl_ucp_team_create_test(ucc_base_team_t *tl_team)
     ucc_tl_ucp_context_t *ctx  = UCC_TL_UCP_TEAM_CTX(team);
     ucc_status_t          status;
 
+    if (USE_SERVICE_WORKER(team)) {
+        team->worker = &ctx->service_worker;
```
Sergei-Lebedev, line 100: does this line mean that we will use service_worker for all the subsequent communications or I'm missing something?

vspetrov: @samnordmann yes, caching worker this way does not work. we have to use different workers for service colls and all the rest. so looks like we need the implementation with the branch - the one you originally madet.

samnordmann: > does this line mean that we will use service_worker for all the subsequent communications or I'm missing something?
Yes.
I thought this is what we wanted, but ok I think I understand the problem now.



> @samnordmann yes, caching worker this way does not work. we have to use different workers for service colls and all the rest. so looks like we need the implementation with the branch - the one you originally made.

I see, so to be clear: adding a flag `is_service` at the task level and passing the flag to the relevant functions, right ?

vspetrov: or we can cache "worker" on "task". Like you do now, but on "per-task" basis. Then you take branch once during task init.

vspetrov: you are right. lets have a flag then. @Sergei-Lebedev ?

samnordmann: > does this line mean that we will use service_worker for all the subsequent communications or I'm missing something?

I am actually not sure to understand the issue you raised in the first place @Sergei-Lebedev.
This means that we will use service_worker for all subsequent communications **in the service team**. The other teams will use the regular worker.
This is what we want no ? The service team is a special team only used for service collectives, right ?
If yes then there is nothing to fix.

Sergei-Lebedev: yes, missed it, thanks @samnordmann

Sergei-Lebedev (approved):

src/components/tl/ucp/tl_ucp_team.c
vspetrov: @samnordmann yes, you are right. Plz fix 1 liner error:
```
/home/runner/work/ucc/ucc/src/components/tl/ucp/tl_ucp_context.c:362:68: error: parameter 'ctx' is unused [misc-unused-parameters,-warnings-as-errors]
                                             ucc_tl_ucp_context_t *ctx)
```
by adding "//NOLINT and lets merge this

samnordmann: The warning is actually relevant. I removed ctx from the arguments.

vspetrov:
bot:retest

vspetrov:
bot:retest

PR #564, opened 2022-07-11
TL/SELF/CUDA/SHARP: fix lib attr flags

Sergei-Lebedev:
## What
attr flags are not initialized. Context fot these TLs is not created since core
context assumes service team is required.

## Why ?
Fixes gtest
```
[  FAILED  ] test_team.team_create_destroy_ctx_local/0, where GetParam() = 1 (1162 ms)
```


Sergei-Lebedev:
cc @samnordmann

vspetrov (approved):

PR #568, opened 2022-07-14
TL/SHARP: use internal oob

Sergei-Lebedev:
## What
Add support for internal OOB  in TL SHARP

## Why ?
Internal OOB has native support for broadcast, no need to allocate temp buffer and use allgather to emulate broadcast as it's done for SHARP OOB now.
Need to set UCC_TL_SHARP_USE_INTERNAL_OOB=1 to enable internal OOB in TL SHARP.

### DIFF
--- a/src/components/tl/sharp/tl_sharp.c
+++ b/src/components/tl/sharp/tl_sharp.c
@@ -17,6 +17,11 @@ static ucc_config_field_t ucc_tl_sharp_lib_config_table[] = {
     {"", "", NULL, ucc_offsetof(ucc_tl_sharp_lib_config_t, super),
      UCC_CONFIG_TYPE_TABLE(ucc_tl_lib_config_table)},
 
+    {"USE_INTERNAL_OOB", "n",
+     "Use service team to create sharp context",
+     ucc_offsetof(ucc_tl_sharp_lib_config_t, use_internal_oob),
+     UCC_CONFIG_TYPE_BOOL},
+
     {NULL}};
 
 static ucc_config_field_t ucc_tl_sharp_context_config_table[] = {
@@ -47,6 +52,7 @@ static ucc_config_field_t ucc_tl_sharp_context_config_table[] = {
      "Seed for random sharp job ID. (0 - use default).",
      ucc_offsetof(ucc_tl_sharp_context_config_t, rand_seed),
      UCC_CONFIG_TYPE_UINT},
+
     {NULL}};
 
 UCC_CLASS_DEFINE_NEW_FUNC(ucc_tl_sharp_lib_t, ucc_base_lib_t,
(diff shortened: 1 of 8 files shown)


src/components/tl/sharp/tl_sharp.c
```diff
@@ -17,6 +17,11 @@ static ucc_config_field_t ucc_tl_sharp_lib_config_table[] = {
     {"", "", NULL, ucc_offsetof(ucc_tl_sharp_lib_config_t, super),
      UCC_CONFIG_TYPE_TABLE(ucc_tl_lib_config_table)},
 
+    {"USE_INTERNAL_OOB", "n",
```
vspetrov, line 20: At some point we probably want to make it "y" by default, but need to resolve potential deadlock issue (see below)

src/components/tl/sharp/tl_sharp_context.c
```diff
@@ -126,6 +126,123 @@ static int ucc_tl_sharp_oob_bcast(void *arg, void *buf, int size, int root)
...
+
+    rbuf = ucc_malloc(sizeof(char) * subset.map.ep_num, "tmp_barrier");
+    if (!rbuf) {
+        tl_error(ctx->super.super.lib,
+                 "failed to allocate %zd bytes for tmp barrier array",
+                 sizeof(char) * subset.map.ep_num);
+        return UCC_ERR_NO_MEMORY;
+    }
+
+    status = UCC_TL_TEAM_IFACE(steam)->scoll.allgather(&steam->super, &sbuf,
```
vspetrov, line 152: why not service allreduce? it is faster and no need to malloc

src/components/tl/sharp/tl_sharp_context.c
```diff
@@ -126,6 +126,123 @@ static int ucc_tl_sharp_oob_bcast(void *arg, void *buf, int size, int root)
...
+                                                       subset, &req);
+    if (status != UCC_OK) {
+        tl_error(ctx->super.super.lib, "tl sharp gather failed\n");
+        return status;
+    }
+
+    do {
+        ucc_context_progress(ctx->super.super.ucc_context);
+        status = ucc_collective_test(&req->super);
+    } while (status == UCC_INPROGRESS);
```
vspetrov, line 207: i have one concert about this blocking wait here. When we create team (not context) - means e.g. when we create OMPI comm - a process will block there and will not progress runtime (e.g. MPI p2p). I think we can trigger deadlock... @bureddy what do you think? The only right solution i believe is to make sharp_team creation non-blocking similar to UCC. But this will require work.

bureddy (approved):

Sergei-Lebedev:
bot:retest

src/components/tl/sharp/tl_sharp_context.c
vspetrov: after some debug we couldn't reproduce the hang (looks like additional sync happens in OMPI anyways that prevents that).

vspetrov (approved):

vspetrov:
closed and replaced with #579 (additional fix required)

PR #571, opened 2022-07-21
CORE: option to disable cfg file

vspetrov:
## What
Adds option to disable config files usage

## Why ?
If config file is installed into default path (HOME/ucc.conf) or ucc_install/share/ucc.conf then it will be used and user has no option to not use it (only put another empty file and point UCC_CONFIG_FILE to it, which is not convenient).

## How ?
Now, default value for UCC_CONFIG_FILE is "auto" which means search in default paths. "n", "no", "0", "" means - don't use cfg file.

### DIFF
--- a/src/core/ucc_constructor.c
+++ b/src/core/ucc_constructor.c
@@ -72,7 +72,12 @@ static ucc_status_t ucc_check_config_file(void)
     char *               filename;
 
     /* First check the UCC_CONFIG_FILE - most precedence */
-    if (strlen(cfg->cfg_filename) > 0) {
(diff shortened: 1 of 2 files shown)


bureddy:
what if the user just specifies no value(empty string) ` UCC_CONFIG_FILE=` ?

vspetrov:
> what if the user just specifies no value(empty string) ` UCC_CONFIG_FILE=` ?

It is also treated as "disable cfg file" - similar to UCX behavior

bureddy (approved):

vspetrov:
bot:retest

manjugv (approved):

PR #573, opened 2022-07-22
CORE: fix memory leak

kingchc:
## What
fix memory leak

## Why ?
`ucc_global_config.component_path` initially gets allocated implicitly by ucs (as in a config opt table), but it gets changed when loading libraries at runtime later, and we lost the pointer. Hence, a memory leak.

## How ?
~~reallocate `ucc_global_config.component_path` if we happen to reassign the pointer.~~ (see below disucssion)

### DIFF
--- a/src/core/ucc_constructor.c
+++ b/src/core/ucc_constructor.c
@@ -42,6 +42,7 @@ static int callback(struct dl_phdr_info *info, size_t size, void *data)
         component_path = (char *)ucc_malloc(len, "component_path");
         if (!component_path) {
             ucc_free(install_path);
+            ucc_global_config.install_path = NULL;
             ucc_error("failed to allocate %zd bytes for component_path", len);
             return -1;
         }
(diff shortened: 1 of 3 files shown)


kingchc:
cc @Sergei-Lebedev @vspetrov

vspetrov:
ok to test

vspetrov (commented):
@kingchc you are correct about the leak. I'm not sure that this solution is best though. This way you indirectly assume ucs_config_parser_release_opts implementation, ie. you assume it will just call "free" on component_path. It is currently true, but can be different in general.

We already have the second field in global_opts: component_path_default - which is unused. I think the proper fix would be to not alter component_path parameter (so it is released by ucs as part of cfg) but store custom path that we find in the second component_path_default (or rename it ), and in the destructor also release it with free explicitly

kingchc:
> @kingchc you are correct about the leak. I'm not sure that this solution is best though. This way you indirectly assume ucs_config_parser_release_opts implementation, ie. you assume it will just call "free" on component_path. It is currently true, but can be different in general.
> 
> We already have the second field in global_opts: component_path_default - which is unused. I think the proper fix would be to not alter component_path parameter (so it is released by ucs as part of cfg) but store custom path that we find in the second component_path_default (or rename it ), and in the destructor also release it with free explicitly

Makes sense. To minimize changes, I think we can just use `component_path` and `component_path_default` (perhaps rename to sth like `component_path_customized`), and keep the latter in config table for ucs. We do malloc for `cfg->component_path` as the final path to be used for loading components, and free it in destructor. So we don't touch the buffer ucs is handling. 

In this way, there are two possible sources:
1. `UCC_COMPONENT_PATH` is provided: read it from `cfg->component_path_customized` and copy it to `cfg->component_path`
2. `UCC_COMPONENT_PATH` is not provided (i.e., `cfg->component_path_customized` is empty): search the path as usual, and set `cfg->component_path` accordingly

What do you think? I can update the PR shortly.

kingchc:
Closing as fixed in #586

PR #577, opened 2022-07-26
CORE: reorganize pre-defined ucc name

kingchc:
## What
Move definition of UCC component path and library name to `config.h`

## Why ?
In the case of using customized UCC component path and library name, those changes need to be made in `src/core/ucc_constructor.c` and can not be upstreamed. This PR moves pre-defined UCC component path and library name to `config.h`, which is not tracked and can be easily customized without disturbing the upstream process.

## How ?
- `UCC_COMPONENT_LIBDIR` is always the same as `UCC_MODULE_SUBDIR`
- move `UCC_LIB_SO_NAME` to configure.ac (will generate `config.h` which is not tracked and can be easily changed) 
- minor: code formatting of `src/core/ucc_constructor.c`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -100,6 +100,7 @@ AC_SUBST([localmoduledir], ['$(abs_top_builddir)/modules']) # local directory fo
 AC_SUBST([objdir],         [${objdir}])                     # libtool objects dir, usually .libs
 AC_SUBST([shrext],         [${shrext_cmds}])                # libtool shared library extension
 AC_DEFINE_UNQUOTED([UCC_MODULE_SUBDIR], ["${modulesubdir}"], [UCC module sub-directory])
+AC_DEFINE_UNQUOTED([UCC_LIB_SO_NAME], ["libucc.so"], [Default name of UCC shared object])
 
 #
 # Additional m4 files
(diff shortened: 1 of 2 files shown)


kingchc:
This is not ready for review and will likely be changed, but I would like to get some early feedback and discuss it on UCC WG. (cc @manjugv @vspetrov @Sergei-Lebedev)

vspetrov:
overall looks good to me. lets not make code-style alignment changes not related to the actual PR.

kingchc:
> overall looks good to me. lets not make code-style alignment changes not related to the actual PR.

thanks for the feedback. I reverted the changes of code-style and make this PR ready for review.

vspetrov:
@kingchc this is a bit strange since setting UCC_SONAME does not really change the name of the result library (it is still libucc.so).
I propose new PR (https://github.com/openucx/ucc/pull/586) that fixes both: your case with custom library name and memleak (#573 ). Wdyt?

kingchc:
> @kingchc this is a bit strange since setting UCC_SONAME does not really change the name of the result library (it is still libucc.so). I propose new PR (#586) that fixes both: your case with custom library name and memleak (#573 ). Wdyt?

@vspetrov - Right, I realized it would be a bigger change with multiple makefiles to support such corner case, so I have not don it. I agreed current PR is confusing. Your proposal in #586 to remove them all makes sens, I will review/test that one and close this PR and #573. Thanks!

kingchc:
Closing as fixed in #586

PR #587, opened 2022-08-04
TOOLS: add ec ops to perftest

Sergei-Lebedev:
## What
Benchmark operation in EC components with perftest
Following operations are supported
* reducedt - strided reduce multi with different number of vectors
* memcpy - memory copy
* triggered and non-triggered modes

### DIFF
--- a/tools/perf/ucc_pt_benchmark.cc
+++ b/tools/perf/ucc_pt_benchmark.cc
@@ -3,54 +3,65 @@
 #include "components/mc/ucc_mc.h"
 #include "ucc_perftest.h"
 #include "utils/ucc_coll_utils.h"
+#include "core/ucc_ee.h"
 
 ucc_pt_benchmark::ucc_pt_benchmark(ucc_pt_benchmark_config cfg,
                                    ucc_pt_comm *communicator):
     config(cfg),
     comm(communicator)
 {
-    switch (cfg.coll_type) {
-    case UCC_COLL_TYPE_ALLGATHER:
+    switch (cfg.op_type) {
+    case UCC_PT_OP_TYPE_ALLGATHER:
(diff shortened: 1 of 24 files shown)


tools/perf/ucc_pt_benchmark.cc
```diff
@@ -162,13 +176,69 @@ ucc_status_t ucc_pt_benchmark::run_single_test(ucc_coll_args_t args,
...
+    }
+
+    UCCCHECK_GOTO(ucc_ee_executor_stop(executor), exit_err, st);
+    if (niter != 0) {
+        time /= niter;
+    }
+    return UCC_OK;
+
+free_req:
+    ucc_ee_executor_task_finalize(task);
```
vspetrov, line 230: do you need executor_stop here?

tools/perf/ucc_pt_comm.cc
```diff
@@ -25,12 +26,12 @@ void ucc_pt_comm::set_gpu_device()
     int dev_count = 0;
 
     if (ucc_pt_cudaGetDeviceCount(&dev_count) == 0 && dev_count != 0) {
-	ucc_pt_cudaSetDevice(bootstrap->get_local_rank() % dev_count);
-	return;
+    	ucc_pt_cudaSetDevice(bootstrap->get_local_rank() % dev_count);
+	    return;
```
vspetrov, line 30: misaligned?

vspetrov (approved):
Looks good. need to rebase on EE task args changes

tools/perf/ucc_pt_benchmark.cc
Sergei-Lebedev: yes, fixed

tools/perf/ucc_pt_comm.cc
Sergei-Lebedev: fixed

bureddy (approved):

PR #590, opened 2022-08-05
UTIL: fix coll_args_msgsize routine

vspetrov:
## What
ucc_coll_args_msgsize is used to  compute msgsize of collective operation which is defined by the ucc_coll_args_t. However args is not enough. For some collectives the msgsize depends on team_size and team_rank (to detect root). At the same time this fn can be called for a collective defined on a TL team which is a subset of CORE ucc_team. In this case we need to use proper  rank/size info to compute total coll size, i.e. using core team->rank/size is incorrect.

## Why ?
Current implementation leads to crash with CL/HIER (when ucc_coll_init is called for a subgroup).

## How ?
Pass rank/size explicitly to ucc_coll_args_msgsize. For ucc_coll_init we will store rank/size on the score_map.

### DIFF
--- a/src/coll_score/ucc_coll_score_map.c
+++ b/src/coll_score/ucc_coll_score_map.c
@@ -10,6 +10,11 @@
 
 typedef struct ucc_score_map {
     ucc_coll_score_t *score;
+    /* Size, rank of the process in the base_team associated with that
+       score_map. It can be CL or TL team, which can be a subset of a
+       core UCC team */
+    ucc_rank_t        team_size;
+    ucc_rank_t        team_rank;
 } ucc_score_map_t;
 
 ucc_status_t ucc_coll_score_build_map(ucc_coll_score_t *score,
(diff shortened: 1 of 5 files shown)


src/coll_score/ucc_coll_score_map.c
```diff
@@ -66,9 +78,12 @@ ucc_status_t ucc_coll_score_map_lookup(ucc_score_map_t      *map,
                                        ucc_base_coll_args_t *bargs,
                                        ucc_msg_range_t     **range)
 {
-    ucc_memory_type_t mt      = ucc_coll_args_mem_type(bargs);
+    ucc_memory_type_t mt      = ucc_coll_args_mem_type(&bargs->args,
```
Sergei-Lebedev, line 81: minor, maybe change input parameter from ucc_bace_coll_args to ucc_coll_args since base_args is not needed here

Sergei-Lebedev (approved):

src/coll_score/ucc_coll_score_map.c
vspetrov: it's same pointer in the end of the day. i wanted to have change smaller. maybe we will need to pass smth else from base_args. i would keep it.

PR #593, opened 2022-08-09
BUILD/ROCM: check ROCm version in configure logic

edgargabriel:
## What
check ROCm version in configure logic

## Why
This allows us to be more accurate when specifying include paths. Reduces the number of warnings generated for the rocm ec component.

### DIFF
--- a/config/m4/rocm.m4
+++ b/config/m4/rocm.m4
@@ -39,8 +39,8 @@ AC_DEFUN([ROCM_BUILD_FLAGS],
 # Parse value of ARG into appropriate LIBS, LDFLAGS, and
 # CPPFLAGS variables.
 AC_DEFUN([HIP_BUILD_FLAGS],
-    $4="-D__HIP_PLATFORM_AMD__ -I$1/include/hip -I$1/include  -I$1/hip/include"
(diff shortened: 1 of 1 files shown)


vspetrov (approved):

edgargabriel:
if possible, this patch would be good to have on the 1.1 branch as well, it reduces the number of warnings in the ec/rocm with rocm 5.2.x tremendously

Sergei-Lebedev (approved):

vspetrov:
> if possible, this patch would be good to have on the 1.1 branch as well, it reduces the number of warnings in the ec/rocm and mc/rocm components with rocm 5.2.x tremendously

@edgargabriel could you plz open a PR against v1.1.x ?

edgargabriel:
@vspetrov short question: what is the preferred way to do the pr for v1.1, cherry-picking the commit from master or filing a completely new pr? ( I have trouble cherry-picking the commit id, it claims 'fatal: bad object ` for that commit id, which is kind of weird).

vspetrov:
@edgargabriel i've created a cherry-pick https://github.com/openucx/ucc/pull/594

PR #596, opened 2022-08-14
Ec cuda support dt ops

samnordmann:
## What
_Describe what this PR is doing._ 

## Why ?
_Justification for the PR. If there is existing issue/bug please reference. For
bug fixes why and what can be merged in a single item._

## How ?
_It is optional but for complex PRs please provide information about the design,
architecture, approach, etc._


src/components/ec/cuda/kernel/ec_cuda_executor.cu
```diff
@@ -291,22 +451,11 @@ __global__ void executor_kernel(volatile ucc_ec_cuda_executor_t *eee,
         if (worker_done) {
```
samnordmann, line 451: the line 
`args = tasks[cidx_local].args;`
could be replaced by
```
                memcpy(&args, &tasks[cidx_local].args,
                       sizeof(ucc_ee_executor_task_args_t) -
                           2 * sizeof(uint16_t));
```
since we don't need to copy the attributes task_type and flags. However this results in poorer performances, which I find very counter intuitive.
I also tried the solution where the size of the task args is passed to the kernel as an argument and only the exact size is copied to the device (currently, whatever the task type, the nbr of bytes we copy are the one of the multi_copy task args which is the largest). But this induces an extra read into the global memory (to retrieve the size) which hurts performances in a dramatic way.

PR #598, opened 2022-08-15
TEST: adding support for persistent coll to mpi test

shimmybalsam:
## What
Adding support for persistent colls to test mpi

### DIFF
--- a/src/core/ucc_coll.c
+++ b/src/core/ucc_coll.c
@@ -58,6 +58,16 @@ static ucc_status_t ucc_check_coll_args(const ucc_coll_args_t *coll_args,
                                            coll_args->dst.info);
         }
         break;
+    case UCC_COLL_TYPE_BCAST:
+        if (UCC_IS_INPLACE(*coll_args)) {
+            ucc_warn("Inplace not supported in bcast");
+        }
+        break;
+    case UCC_COLL_TYPE_BARRIER:
+        if (UCC_IS_INPLACE(*coll_args)) {
+            ucc_warn("Inplace not supported in barrier");
+        }
+        break;
     default:
         return UCC_OK;
     }
--- a/src/core/ucc_coll.c
+++ b/src/core/ucc_coll.c
@@ -59,13 +59,12 @@ static ucc_status_t ucc_check_coll_args(const ucc_coll_args_t *coll_args,
         }
         break;
     case UCC_COLL_TYPE_BCAST:
-        if (UCC_IS_INPLACE(*coll_args)) {
-            ucc_warn("Inplace not supported in bcast");
-        }
-        break;
     case UCC_COLL_TYPE_BARRIER:
+    case UCC_COLL_TYPE_FANIN:
+    case UCC_COLL_TYPE_FANOUT:
         if (UCC_IS_INPLACE(*coll_args)) {
-            ucc_warn("Inplace not supported in barrier");
+            ucc_warn("Inplace flag for %s is not defined by UCC API",
+                     ucc_coll_type_str(coll_args->coll_type));
         }
         break;
     default:
--- a/test/mpi/test_allgatherv.cc
+++ b/test/mpi/test_allgatherv.cc
@@ -103,7 +103,7 @@ ucc_status_t TestAllgatherv::set_input()
     return UCC_OK;
 }
 
-ucc_status_t TestAllgatherv::reset_sbuf()
+ucc_status_t TestAllgatherv::reset_sbuf(int persistent = 0)
 {
     return UCC_OK;
 }
--- a/test/mpi/test_allgatherv.cc
+++ b/test/mpi/test_allgatherv.cc
@@ -60,7 +60,7 @@ TestAllgatherv::TestAllgatherv(ucc_test_team_t &_team, TestCaseParams &params) :
     UCC_MALLOC_CHECK(check_buf);
     fill_counts_and_displacements(size, count, counts, displacements);
 
-    if (TEST_NO_INPLACE == inplace) {
+    if (!inplace) {
         UCC_CHECK(ucc_mc_alloc(&sbuf_mc_header, counts[rank] * dt_size, mem_type));
         sbuf                   = sbuf_mc_header->addr;
         args.src.info.buffer   = sbuf;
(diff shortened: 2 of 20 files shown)


test/mpi/test_allgatherv.cc
```diff
@@ -65,8 +65,13 @@ TestAllgatherv::TestAllgatherv(ucc_test_team_t &_team, TestCaseParams &params) :
         UCC_CHECK(ucc_mc_alloc(&sbuf_mc_header, counts[rank] * dt_size, mem_type));
         sbuf = sbuf_mc_header->addr;
     } else {
-        args.mask = UCC_COLL_ARGS_FIELD_FLAGS;
-        args.flags = UCC_COLL_ARGS_FLAG_IN_PLACE;
+        args.mask  |= UCC_COLL_ARGS_FIELD_FLAGS;
+        args.flags |= UCC_COLL_ARGS_FLAG_IN_PLACE;
+    }
+
+    if (persistent) {
```
Sergei-Lebedev, line 72: i think it's better to set UCC_COLL_ARGS_FLAG_PERSISTENT in TestCase ctor, then you don't need to copy paste same code for each collective subclass. Also do the same for inplace flag

shimmybalsam: Will move for persistent.
Regarding inplace though, as it is not supported in bcast and barrier, setting flags in ctor will be problematic. I can skip test in inplace case in bcast and barrier, or that the inplace flag needs to be set in each coll seperately as is now. What do you think @Sergei-Lebedev?

Sergei-Lebedev: @manjugv @vspetrov if we set UCC_COLL_ARGS_FLAG_IN_PLACE for collectives that don't support inplace mode such as barrier what is the expected behavior? is it an error i.e. invalid argument or for such collective inplace flag simply ignored?

vspetrov: It is user responsibility to clear/init coll_args->mask. So, my take is, if user sets UCC_COLL_ARGS_FIELD_FLAGS as mask bit, then user must initialize "flags" properly. And if user put IN_PLACE for bcast in that case, it is invalid param. We could just ignore that of course, but i think good implementation should print error. Because, this means user set the flag on purpose (since both mask value is set and flag is set) - so user must understand that he is not getting what he wants.

manjugv: Agree @vspetrov. We can print warning instead of error though.

src/core/ucc_coll.c
```diff
@@ -58,6 +58,19 @@ static ucc_status_t ucc_check_coll_args(const ucc_coll_args_t *coll_args,
                                            coll_args->dst.info);
         }
         break;
+    case UCC_COLL_TYPE_BCAST:
+        if (UCC_IS_INPLACE(*coll_args)) {
+            ucc_warn("Inplace not supported in bcast");
+        }
+        break;
+    case UCC_COLL_TYPE_BARRIER:
```
Sergei-Lebedev, line 66: why barrier doesn't support persistent? I guess same is true for fanin and fanout

shimmybalsam: barrier (and fanin, fanout) supports persistent inside UCC collectives, but not in mpi test since it is not implemented in the test as a single collective test. 
And in general we dont have fanin/fanout mpi tests.

Sergei-Lebedev: makes sense, but then in core we should check only for inplace flag since persistent is ok to use

shimmybalsam: removed persistent check in core

test/mpi/test_mpi.cc
```diff
@@ -447,57 +458,61 @@ void set_gpu_device(test_set_gpu_device_t set_device)
 #endif
 
 std::vector<ucc_status_t> UccTestMpi::exec_tests(
-        std::vector<std::shared_ptr<TestCase>> tcs, bool triggered)
+        std::vector<std::shared_ptr<TestCase>> tcs, bool triggered,
+                                                    int  persistent)
```
vspetrov, line 462: why "int persistent"? should be ucc_test_mpi_persistent_t

vspetrov: or event just make it "bool" as triggered. and same for inplace

test/mpi/test_mpi.cc
```diff
@@ -304,6 +305,16 @@ int ucc_coll_inplace_supported(ucc_coll_type_t c)
     }
 }
 
+int ucc_coll_persistent_supported(ucc_coll_type_t c)
+{
+    switch(c) {
+    case UCC_COLL_TYPE_BARRIER:
```
vspetrov, line 311: @shimmybalsam i had another look. Lets add support for barrier as well. It looks easy. In the test_barrier.cc, if "persistent" we just don't need to re-init barrier request:
```
 82       if (i < size - 1) {
  83           UCC_CHECK(ucc_collective_finalize(req));
  84            UCC_CHECK(ucc_collective_init(&args, &req, team.team));
  85      }
```
The lines above should be skipped in persistent case. this looks all .

vspetrov (changes requested):

test/mpi/test_mpi.cc
shimmybalsam: added

test/mpi/test_mpi.cc
shimmybalsam: changed both inplace and persistent to bool

vspetrov:
bot:retest

vspetrov:
bot:retest

src/core/ucc_coll.c
```diff
@@ -58,6 +58,16 @@ static ucc_status_t ucc_check_coll_args(const ucc_coll_args_t *coll_args,
...
+    case UCC_COLL_TYPE_BCAST:
+        if (UCC_IS_INPLACE(*coll_args)) {
+            ucc_warn("Inplace flag for bcast is not defined by UCC API");
+        }
+        break;
+    case UCC_COLL_TYPE_BARRIER:
+        if (UCC_IS_INPLACE(*coll_args)) {
+            ucc_warn("Inplace flag for barrier is not defined by UCC API");
+        }
+        break;
```
Sergei-Lebedev, line 70:
```suggestion
    case UCC_COLL_TYPE_BCAST:
    case UCC_COLL_TYPE_BARRIER:
    case UCC_COLL_TYPE_FANIN:
    case UCC_COLL_TYPE_FANOUT
        if (UCC_IS_INPLACE(*coll_args)) {
            ucc_warn("Inplace flag for %s is not defined by UCC API", ucc_coll_str(coll_args->coll_type));
        }
        break;
```

Sergei-Lebedev (approved):

PR #548, opened 2022-06-22
TL/CUDA: fix exec task finalize

Sergei-Lebedev:
## What
Alltoall fixes
* Fix executor task finalize is not being called
* Don't post executor task with zero size

PR #599, opened 2022-08-16
TL/CUDA: handle empty buffers for IPC registration cache

kingchc:
## What
Handling empty buffers for in TL/CUDA's IPC registration cache

## Why ?
https://github.com/openucx/ucc/pull/548 handles the case that zero counts/sizes when posting ops to the executor kernel. However, in some use cases like CL/HIER or skewed alltoallv, some ranks may enter Alltoallv with empty src or dst buffers (or zero total count). Need to handle it gracefully instead of error out when perfoming IPC memops.

## How ?

Silently return `UCC_OK` in `ucc_tl_cuda_map_memhandle` and `ucc_tl_cuda_mem_info_get` if null ptr and/or zero size are passed


kingchc:
cc @Sergei-Lebedev @vspetrov @bureddy

vspetrov:
ok to test

src/components/tl/cuda/tl_cuda_cache.c
```diff
@@ -145,6 +145,9 @@ ucc_tl_cuda_map_memhandle(const void *d_ptr, size_t size,
                           cudaIpcMemHandle_t mem_handle, void **mapped_addr,
                           ucc_tl_cuda_cache_t *cache)
 {
+    if (d_ptr == NULL || size == 0) {
```
vspetrov, line 148: should we set mapped_addr to some value, e.g. NULL? otherwise we return UCC_OK and leave *mapped_addr uninitialized which looks fishy

kingchc: set `*mapped_addr` to NULL.

vspetrov (approved):

Sergei-Lebedev (approved):

PR #600, opened 2022-08-18
CL/HIER: allreduce rab fix

vspetrov:
## What
Fixes:
[1660824500.613377] [swx-clx02:46991:0] bcast_sag_knomial.c:90   TL_UCP ERROR failed to init scatter_knomial task
observed in ucc_test_mpi with CL/HIER

## How ?
Don't check for INPLACE flag in tl_ucp_scatter_knomial. Currently it is not generic scatter - it is special implementation for bcast_sag.

### DIFF
--- a/src/components/cl/hier/allreduce/allreduce_rab.c
+++ b/src/components/cl/hier/allreduce/allreduce_rab.c
@@ -39,6 +39,9 @@ UCC_CL_HIER_PROFILE_FUNC(ucc_status_t, ucc_cl_hier_allreduce_rab_init,
     ucc_base_coll_args_t args;
     int                  n_tasks, i;
 
+    if (coll_args->args.op == UCC_OP_AVG) {
+        return UCC_ERR_NOT_SUPPORTED;
+    }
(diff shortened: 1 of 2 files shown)


bureddy (approved):

Sergei-Lebedev:
I think the proper fix should be in scatter knomial or bcast sag. Bcast indeed doesn't have in-place flag and I would expect that bcast implementation just ignores that flag if it's set instead of throwing an error.

vspetrov:
> I think the proper fix should be in scatter knomial or bcast sag. Bcast indeed doesn't have in-place flag and I would expect that bcast implementation just ignores that flag if it's set instead of throwing an error.

Scatter knomial was planned to be used as "scatter" collective (even though currently only used as part of bcast). And it can not ignore INPLACE flag because inplace is an option for scatter. Bcast SAG does not know anything about scatter implementation.

Sergei-Lebedev (approved):

manjugv:
@vspetrov do you need to port this to 1.1?

vspetrov:
> @vspetrov do you need to port this to 1.1?

Ported #617

PR #610, opened 2022-08-29
EC/CUDA: implement all the exec reductions

vspetrov:
## What
Potential Alternative for #596 . This PR implements ALL the reductinos (dt/ops) in the ec/cuda executor for persistent mode. It is done by making "device template" functions (common both for persistent and interruptible) + some macro magic + switch/case in the executor kernel. 

## Why ?
Need to support all the reductions. #596 currently demonstrates some perf degradation.

I only did a quick check on ucc_perftest -c reducedt and didn't see a regression there. However, more detail perf check is required. @samnordmann plz take this PR and do a perf regression check similar to what you did for 596. Lets then decide which approach we take.

### DIFF
--- a/src/components/ec/cuda/ec_cuda_executor_persistent.c
+++ b/src/components/ec/cuda/ec_cuda_executor_persistent.c
@@ -16,34 +16,6 @@ ucc_cuda_executor_persistent_task_post(ucc_ee_executor_t *executor,
                                                        ucc_ec_cuda_executor_t);
     int                     max_tasks = EC_CUDA_CONFIG->exec_max_tasks;
     ucc_ee_executor_task_t *ee_task;
-    ucc_datatype_t          dt;
-    ucc_reduction_op_t      op;
-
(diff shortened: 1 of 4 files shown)


samnordmann:
Here is the data collected on dgx that compares PR#596, the present PR, and the master branch.
What we observe is that the present PR does not lead to performance degradation whereas PR#596 does. So this approach seems to be preferable.

Also I observe that the present PR performs way better than master for large msg sizes.

I thus think that this PR is the good implementation
[perftest comparison - Spreadsheet2.xlsx](https://github.com/openucx/ucc/files/9526809/perftest.comparison.-.Spreadsheet2.xlsx)

samnordmann (approved):

samnordmann:
A fix is needed for triggered collective with several workers. See the [commit](https://github.com/vspetrov/ucc/pull/4)

samnordmann:
Replaced by [PR#670](https://github.com/openucx/ucc/pull/670)

PR #612, opened 2022-08-30
CORE: correctly handle failure in ctx_create epilog

vspetrov:
## What
Properly handle potential failures that happen during TL context_create_epilog call.  

## Why ?
Current behavior: if context_create_epilog fails -> ucc context creation fails -> job fails.
Expected behavior: if TL that failed to perform context_create_epilog is not explicitly required job must proceed w/o it. Example: TL/SHARP not enabled explicitly (in a "try" mode), it might fail to initialize but we should proceed.

## How ?
* Check if the TL is explicitly requested
* The order of steps during ucc_context create is: 1) create all TL contexts, 2) create all CL contexts, 3) perform global address exchange (if required via params) - all TLs/CLs ctx objects are required for that, 4) allocate topo object if required; 5) allocate ctx service team (if required); 5) call epilog
* When CL contexts are created at step 2 they check which TLs are available for them. However, some of those TLs might get invalidated later at step 5 if tl_ctx_craete_epilog will fail
* In that case we need to remove failed TL contexts from the CLs - we can do that since those TLs were NOT explicitly requested (otherwise the job would stop)
* This is done by moving array of tl_context_t ptrs to the base ucc_cl_context_t structure, so that it is available at CORE level.

### DIFF
--- a/src/core/ucc_context.c
+++ b/src/core/ucc_context.c
@@ -551,6 +551,27 @@ poll:
     return UCC_OK;
 }
 
+static void remove_tl_ctx_from_array(ucc_tl_context_t **array, unsigned *size,
+                                     ucc_tl_context_t *tl_ctx)
+{
+    int i;
+
+    for (i = 0; i < (*size); i++) {
+        if (array[i] == tl_ctx) {
+            break;
+        }
+    }
+    if (i == (*size)) {
+        /* given tl_ctx is not part of array */
+        return;
+    }
+    /* decrement array size and do cyclic shift */
+    (*size)--;
+    for (; i < (*size); i++) {
+        array[i] = array[i + 1];
+    }
+}
+
 ucc_status_t ucc_context_create_proc_info(ucc_lib_h                   lib,
                                           const ucc_context_params_t *params,
                                           const ucc_context_config_h  config,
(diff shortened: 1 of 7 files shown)


manjugv (approved):

manjugv:
@vspetrov I think we should port this to v1.1.

manjugv:
@Sergei-Lebedev @bureddy if you can review this PR, it would be great. We would like to merge this for v1.1 release.

src/core/ucc_context.c
```diff
@@ -759,9 +779,22 @@ ucc_status_t ucc_context_create(ucc_lib_h lib,
...
+                } else {
+                    ucc_debug("ctx create epilog for %s failed: %s",
+                              tl_lib->iface->super.name, ucc_status_string(status));
+                    tl_lib->iface->context.destroy(&tl_ctx->super);
+                    for (j = 0; j < ctx->n_cl_ctx; j++) {
+                        remove_tl_ctx_from_array(ctx->cl_ctx[j]->tl_ctxs,
+                                                 &ctx->cl_ctx[j]->n_tl_ctxs,
+                                                 tl_ctx);
+                    }
+                    remove_tl_ctx_from_array(ctx->tl_ctx, &ctx->n_tl_ctx,
```
bureddy, line 795: is tl_ctx is the union of all the cl_ctx[] in this context?

bureddy (approved):

src/core/ucc_context.c
vspetrov: it is ALL TL contexts opened at CORE level. In case of failure in epiloge for some tl_ctx i firstly loop through ALL CLs and remote this TL ctx from their tl arrays, and then remove it also from CORE level tl storage

Sergei-Lebedev (approved):

PR #614, opened 2022-08-31
CORE: Pipelining params

vspetrov:
## What
Unifies pipelining parameters. Adds ucc_pipeline_params_t and the interface for user to set them + cfg var parser. 

## Why ?
Each time we add another pipelined alg we have to introduce in general 5 user parameters (fragmentation threshold, frag_size, n_frags, pipeline_depth, pipeline order). It results in a lot of code duplication inside UCC and parameters list inflation for user.

depends on #597 and #608

example: -x UCC_TL_UCP_ALLREDUCE_SRA_KN_PIPELINE=thresh=30k:frag_size=16k:parallel
instead of: UCC_TL_UCP_ALLREDUCE_SRA_KN_FRAG_THRESH=30k -x UCC_TL_UCP_ALLREDUCE_SRA_KN_FRAG_SIZE=16k -x UCC_TL_UCP_ALLREDUCE_SRA_KN_PIPELINE=parallel


manjugv (approved):

bureddy (approved):
Can we change the param name to ` xxxx_PIPELINE_TUNE`

Sergei-Lebedev:
replaced by #675

PR #619, opened 2022-09-04
EC/CUDA: use one kernel with cooperative launch

samnordmann:
## What
Use one cooperative kernel for the persistent kernel in EC/CUDA. Tests show that this has little/negligible effects on the performances (cf spreadsheet [PR coop group.xlsx](https://github.com/openucx/ucc/files/9487807/PR.coop.group.xlsx)).

## Why ?
In the present implementation, we have to launch three kernels, `executor_start`, `executor_kernel`, `executor_shutdown_ack`. Only `executor_kernel` actually performs computation, whereas the two other kernels are only used to synchronize the grid, which is a primitive function in cooperative launches.

The cooperative group primitives offer more flexibility that could also result in further improvements in EC/CUDA.

## How ?
The feature is enabled by default but can be controlled by the env variable `UCC_EC_CUDA_USE_COOPERATIVE_LAUNCH`.
At initialization, a we test if the device supports cooperative launch ; if not, the feature is disabled and the program goes on without cooperative launch.

### DIFF
--- a/config/m4/cuda.m4
+++ b/config/m4/cuda.m4
@@ -63,6 +63,13 @@ AS_IF([test "x$cuda_checked" != "xyes"],
                [AC_CHECK_LIB([cudart], [cudaGetDeviceCount],
                              [CUDA_LIBS="$CUDA_LIBS -lcudart"], [cuda_happy="no"])])
 
+         AC_CHECK_HEADERS([cooperative_groups.h],
+                        [flag=1
+                        AC_MSG_RESULT([yes])],
+                        [flag=0
+                        AC_MSG_RESULT([no])])
+           AC_DEFINE_UNQUOTED(CUDA_SUPPORTS_COOP_LAUNCH, $flag, [if CUDA supports cooperative launch.])
+
         # Check nvml header files
         AC_CHECK_HEADERS([nvml.h],
                          [nvml_happy="yes"],
--- a/config/m4/cuda.m4
+++ b/config/m4/cuda.m4
@@ -63,13 +63,6 @@ AS_IF([test "x$cuda_checked" != "xyes"],
                [AC_CHECK_LIB([cudart], [cudaGetDeviceCount],
                              [CUDA_LIBS="$CUDA_LIBS -lcudart"], [cuda_happy="no"])])
 
-         AC_CHECK_HEADERS([cooperative_groups.h],
-                        [flag=1
-                        AC_MSG_RESULT([yes])],
-                        [flag=0
-                        AC_MSG_RESULT([no])])
-           AC_DEFINE_UNQUOTED(CUDA_SUPPORTS_COOP_LAUNCH, $flag, [if CUDA supports cooperative launch.])
-
         # Check nvml header files
         AC_CHECK_HEADERS([nvml.h],
                          [nvml_happy="yes"],
--- a/src/components/ec/cuda/ec_cuda.c
+++ b/src/components/ec/cuda/ec_cuda.c
@@ -73,10 +73,10 @@ static ucc_config_field_t ucc_ec_cuda_config_table[] = {
      ucc_offsetof(ucc_ec_cuda_config_t, reduce_num_blocks),
      UCC_CONFIG_TYPE_ULUNITS},
 
-    {"USE_COOPERATIVE_LAUNCH", "0",
+    {"USE_COOPERATIVE_LAUNCH", "1",
      "whether to use cooperative launch in persistent kernel executor",
      ucc_offsetof(ucc_ec_cuda_config_t, use_cooperative_launch),
-     UCC_CONFIG_TYPE_UINT},
+     UCC_CONFIG_TYPE_BOOL},
 
     {NULL}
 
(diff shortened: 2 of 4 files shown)


config/m4/cuda.m4
```diff
@@ -63,6 +63,13 @@ AS_IF([test "x$cuda_checked" != "xyes"],
                [AC_CHECK_LIB([cudart], [cudaGetDeviceCount],
                              [CUDA_LIBS="$CUDA_LIBS -lcudart"], [cuda_happy="no"])])
 
+         AC_CHECK_HEADERS([cooperative_groups.h],
```
Sergei-Lebedev, line 66: do we need to check for this? support for cooperative group was added long before CUDA 11 which is min supported in UCC

src/components/ec/cuda/kernel/ec_cuda_executor.cu
```diff
@@ -253,6 +258,77 @@ __device__ void executor_copy_multi(ucc_eee_task_copy_multi_t *task)
     }
 }
 
+__global__ void
+executor_kernel_cooperative(volatile ucc_ec_cuda_executor_t *eee, int q_size)
```
Sergei-Lebedev, line 262: pls unify executor_kernel_cooperative and executor_kernel, they are almost identical

src/components/ec/cuda/ec_cuda.c
```diff
@@ -73,6 +73,11 @@ static ucc_config_field_t ucc_ec_cuda_config_table[] = {
      ucc_offsetof(ucc_ec_cuda_config_t, reduce_num_blocks),
      UCC_CONFIG_TYPE_ULUNITS},
 
+    {"USE_COOPERATIVE_LAUNCH", "0",
+     "whether to use cooperative launch in persistent kernel executor",
+     ucc_offsetof(ucc_ec_cuda_config_t, use_cooperative_launch),
+     UCC_CONFIG_TYPE_UINT},
```
Sergei-Lebedev, line 79: UCC_CONFIG_TYPE_BOOL since it's either yes or no

src/components/ec/cuda/ec_cuda.c
```diff
@@ -319,6 +325,18 @@ static ucc_status_t ucc_ec_cuda_init(const ucc_ec_params_t *ec_params)
...
     }
+
+    if (cfg->use_cooperative_launch == 1) {
+        cudaDeviceGetAttribute(&supportsCoopLaunch,
+                               cudaDevAttrCooperativeLaunch, device);
+        if (CUDA_SUPPORTS_COOP_LAUNCH == 0 || supportsCoopLaunch == 0) {
+            ec_error(&ucc_ec_cuda.super,
+                     "CUDA cooperative groups are not supported. "
+                     "Please set UCC_EC_CUDA_USE_COOPERATIVE_LAUNCH to 0");
+            return UCC_ERR_NOT_SUPPORTED;
```
Sergei-Lebedev, line 336: maybe print warning but continue and use noncooperative launch?

samnordmann: I fixed that in the last commit. Now, the parameter `use_cooperative_launch` is initialized to 1 by default, and if the device does not support coop launch it is switched to 0 and a debug message is printed with ec_debug.

src/components/ec/cuda/ec_cuda.c
samnordmann: done. Now the default is 1

config/m4/cuda.m4
samnordmann: Ok I see, I erased it

src/components/ec/cuda/kernel/ec_cuda_executor.cu
```diff
@@ -364,10 +380,20 @@ ucc_status_t ucc_ec_cuda_persistent_kernel_start(ucc_ec_cuda_executor_t *eee)
     int          nb     = EC_CUDA_CONFIG->exec_num_workers;
     int          nt     = EC_CUDA_CONFIG->exec_num_threads;
     int          q_size = EC_CUDA_CONFIG->exec_max_tasks;
-
-    executor_start<<<1, 1, 0, stream>>>(eee->dev_state, eee->dev_cidx);
-    executor_kernel<<<nb, nt, 0, stream>>>(eee, q_size);
-    executor_shutdown_ack<<<1, 1, 0, stream>>>(eee->dev_state);
+    int          useCoopLaunch = EC_CUDA_CONFIG->use_cooperative_launch;
+
+    printf("UseCoopLaunch = %i\n", useCoopLaunch);
```
Sergei-Lebedev, line 385: remove printf

src/components/ec/cuda/kernel/ec_cuda_executor.cu
```diff
@@ -288,28 +290,36 @@ __device__ void executor_copy_multi(ucc_eee_task_copy_multi_t *task)
    }
 }
 
-
 __global__ void executor_kernel(volatile ucc_ec_cuda_executor_t *eee,
-                                int q_size)
+                                int q_size, bool useCoopLaunch)
```
Sergei-Lebedev, line 294: i think it's better to use template parameter for useCoolLaunch

src/components/ec/cuda/ec_cuda.c
```diff
@@ -319,6 +325,18 @@ static ucc_status_t ucc_ec_cuda_init(const ucc_ec_params_t *ec_params)
             return UCC_ERR_NOT_SUPPORTED;
         }
     }
+
+    if (cfg->use_cooperative_launch == 1) {
+        cudaDeviceGetAttribute(&supportsCoopLaunch,
+                               cudaDevAttrCooperativeLaunch, device);
+        if (!supportsCoopLaunch) {
+            cfg->use_cooperative_launch = 0;
+            ec_debug(&ucc_ec_cuda.super,
```
Sergei-Lebedev, line 334: I would change ec_debug to ec_warn, because we want to let user know that default cooperative group mode is not working

Sergei-Lebedev (approved):

vspetrov (approved):

PR #628, opened 2022-09-14
Fix issue with uninitialized config when loading components

### LINKED ISSUE #627: Uninitialized global_config when loading tlcp components
ucc_components_load() fails with tlcp components  because global_config.component_path is not set before the tl_ucp_iface_init constructor is called.

BrodyWilliams:
Resolves #627 by checking if configuration is initialized and, if not, calls ucc_constructor.


BrodyWilliams:
No longer necessary. Problem resolved elsewhere.

PR #632, opened 2022-09-24
BUILD: fix nvcc arch flag fo CUDA 11.0

kingchc:
## What
fix build error for CUDA 11.x and earlier versions

## Why ?
the nvcc arch flags (e.g., compute_86, sm_86) recently added (#601) are not supported before CUDA 11.1 and may result in compilation error.

## How ?
only add `ARCH12` nvcc flags (sm_86, etc) for CUDA 11.1+

### DIFF
--- a/config/m4/cuda.m4
+++ b/config/m4/cuda.m4
@@ -107,7 +107,9 @@ AS_IF([test "x$cuda_checked" != "xyes"],
          AS_IF([test "x$cuda_happy" = "xyes"],
                [AS_IF([test "x$with_nvcc_gencode" = "xdefault"],
                       [AS_IF([test $CUDA_MAJOR_VERSION -eq 11],
-                             [NVCC_ARCH="${ARCH7} ${ARCH8} ${ARCH9} ${ARCH10} ${ARCH11} ${ARCH12}"])],
+			     [AS_IF([test $CUDA_MINOR_VERSION -lt 1],
+                                    [NVCC_ARCH="${ARCH7} ${ARCH8} ${ARCH9} ${ARCH10} ${ARCH11}"],
+                                    [NVCC_ARCH="${ARCH7} ${ARCH8} ${ARCH9} ${ARCH10} ${ARCH11} ${ARCH12}"])])],
                       [NVCC_ARCH="$with_nvcc_gencode"])
                 AC_SUBST([NVCC_ARCH], ["$NVCC_ARCH"])])
          LDFLAGS="$save_LDFLAGS"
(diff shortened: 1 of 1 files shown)


kingchc:
cc @zasdfgbnm @vspetrov @Sergei-Lebedev

vspetrov (approved):

vspetrov:
@kingchc plz rename commit title: CONFIG->BUILD. otherwise Codestyle fails

zasdfgbnm:
I think ARCH12 (SM86) is introduced in CUDA 11.1, instead of CUDA 12?

config/m4/cuda.m4
```diff
@@ -107,7 +107,9 @@ AS_IF([test "x$cuda_checked" != "xyes"],
          AS_IF([test "x$cuda_happy" = "xyes"],
                [AS_IF([test "x$with_nvcc_gencode" = "xdefault"],
                       [AS_IF([test $CUDA_MAJOR_VERSION -eq 11],
-                             [NVCC_ARCH="${ARCH7} ${ARCH8} ${ARCH9} ${ARCH10} ${ARCH11} ${ARCH12}"])],
+			     [NVCC_ARCH="${ARCH7} ${ARCH8} ${ARCH9} ${ARCH10} ${ARCH11}"],
+			     [test $CUDA_MAJOR_VERSION -eq 12],
```
zasdfgbnm, line 111: I think this should be CUDA 11.1, not CUDA 12. To double check, you can run
```
nvcc --version &&  touch test.cu && nvcc -gencode arch=compute_86,code=compute_86 test.cu
```
in
```
# CUDA 11.0
docker run --gpus all -it nvcr.io/nvidia/pytorch:20.08-py3
# CUDA 11.1
docker run --gpus all -it nvcr.io/nvidia/pytorch:20.12-py3
```

zasdfgbnm (changes requested):

config/m4/cuda.m4
kingchc: @zasdfgbnm - Thanks for the info, I likely made the wrong assumption. I just double-checked, the flag did exist in 11.4, but not in 11.0. Is it introduced in 11.1 I presume? I will update the PR to check the minor version then.

kingchc: @zasdfgbnm @Sergei-Lebedev - One side question, I thought `ARCH12` means new flags introduced by CUDA 12, thus made the wrong assumption. Looks like `sm_86` was introduced by CUDA 11.1, would `ARCH111` be a better name to avoid confusion? WDYT?

zasdfgbnm: I prefer the `ARCH111` naming scheme :)

zasdfgbnm (approved):

config/m4/cuda.m4
Sergei-Lebedev: @kingchc ARCH111 looks good to me, pls also rename ARCH11 to ARCH110 and others as @zasdfgbnm suggested

kingchc: Updated

Sergei-Lebedev (approved):

PR #646, opened 2022-10-05
EC/CUDA: opt persistent multi ops

Sergei-Lebedev:
## What
Optimize copy_multi and reduce_multi_dst operations for persistent executor in TL CUDA

## How ?
Multi operations are split into smaller subtasks to utilize multiple workers (blcoks) in persistent kernel


manjugv (approved):

manjugv:
@edgargabriel can you provide an extra review on this? Thanks.

bureddy (approved):

edgargabriel (approved):
I think it looks good to me

PR #647, opened 2022-10-05
DOCS: Updating NEWS for v1.1 (port to v1.1)

manjugv:


NEWS
```diff
@@ -3,7 +3,68 @@
...
+
+### GPU Collectives (CUDA, NCCL TL and RCCL TL)
+- Added support for CUDA TL (intranode collectives for NVIDIA GPUs)
+- Added multiring allgatherv, alltoall, reduce-scatter, and reduce-scatterv
+  multiring in CUDA TL
+- Added topo based ring construction in CUDA TL to maximize bandwidth
+- Added NCCL gather, scatter and its vector variant
+- Enable using multiple streams for collectives
+- Added support for RCCL gather (v), scatter (v), broadcast, allgather (v),
+  barrier, alltoall (v) and all reduce collectives
```
edgargabriel, line 54: maybe we should add RCCL in the TL section above instead, (e.g. added TL RCCL) since we added TL rccl  in this release, not just the operations themselves ( I think these are all operations anyway)

manjugv: Something like this "Added RCCL TL to support RCCL collectives" ?

edgargabriel: yes, I think this sounds good

vspetrov (approved):

PR #657, opened 2022-10-23
EC/CUDA: optimize reduce with unrolling

samnordmann:
## What
Various optimizations in reduce ad reduce_strided kernels. We get the following improvements on the BW (see also graphs below):
- triggered mode (aka persistent mode): BW improved by a factor of ~2
- interruptible mode, reduce collective: BW improved by a factor of ~10
- interruptible mode, reduce_strided collective: BW roughly unchanged

## How
- To optimize the triggered mode, we split and unroll the loops in the kernel. The Unroll factor has been hand-tuned to 7. Higher value could potentially lead to better performances but with the current default number of threads (=512), the kernel does not fit into the GPU (tested on Tesla V100 32GB)
-  We have to set `UCC_EC_CUDA_USE_COOPERATIVE_LAUNCH` to 0 otherwise the persistent kernel with the unrolled loop with unroll factor 7 is too large to fit in the GPU (tested on Tesla V100 32GB)
- to avoid code duplication, we use the same implementation for triggered and interruptible mode. Only the unrolling constant is tuned independently in the two cases
- For interruptible mode, the Unroll factor is set to 1. We then use a series of non-conventional optimizations which yield surprisingly good results (BW improved by a factor 10 for reduce collective). These optimizations are especially important in the interruptible mode when `UCC_EC_CUDA_REDUCE_NUM_BLOCKS=auto` which sets the number of blocks to the maximum possible. The optimizations can be seen for example on `src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h:90` where we use 
```
            for (j = 1; j < UCC_EE_EXECUTOR_NUM_BUFS; j++) {                    \
                if (j >= n_srcs) {                                              \
                    break;                                                      \
                }                                                               \
```
instead of
```
for (j = 1; j < n_srcs; j++) { 
```
another example of non-conventional optimization is on `src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h:83` where we use 
```
        memcpy(s, task.srcs, UCC_EE_EXECUTOR_NUM_BUFS * sizeof(_Type *));       \
```
instead of
```
        memcpy(s, task.srcs, n_srcs* sizeof(_Type *));       \
```
I personally don't understand the intuition behind these optimizations, I just notice that they lead to significantly improved performance. Any intuition or help to understand this would be much welcome.

Performance on A100 GPU:

reducedt_UNROLL=8_blocks=80_threads=auto_a100_FINAL

Performance on V100GPU:
reducedt_dt=float32_master_plot
reducedt_dt=float32_master-T_plot
reducedt_strided_dt=float32_master_plot
reducedt_strided_dt=float32_master-T_plot
SUMMARY_reducedt_dt=int32_master_plot
SUMMARY_reducedt_dt=int32_master-T_plot
SUMMARY_reducedt_strided_dt=int32_master_plot
SUMMARY_reducedt_strided_dt=int32_master-T_plot

### DIFF
--- a/src/components/ec/cuda/ec_cuda.c
+++ b/src/components/ec/cuda/ec_cuda.c
@@ -73,7 +73,7 @@ static ucc_config_field_t ucc_ec_cuda_config_table[] = {
      ucc_offsetof(ucc_ec_cuda_config_t, reduce_num_blocks),
      UCC_CONFIG_TYPE_ULUNITS},
 
-    {"USE_COOPERATIVE_LAUNCH", "1",
+    {"USE_COOPERATIVE_LAUNCH", "0",
      "whether to use cooperative launch in persistent kernel executor",
      ucc_offsetof(ucc_ec_cuda_config_t, use_cooperative_launch),
      UCC_CONFIG_TYPE_BOOL},
--- a/src/components/ec/cuda/ec_cuda.c
+++ b/src/components/ec/cuda/ec_cuda.c
@@ -49,12 +49,12 @@ static ucc_config_field_t ucc_ec_cuda_config_table[] = {
      UCC_CONFIG_TYPE_UINT},
 
     {"EXEC_NUM_WORKERS", "1",
-     "Number of thread blocks to use for cuda executor",
+     "Number of thread blocks to use for cuda persistent executor",
      ucc_offsetof(ucc_ec_cuda_config_t, exec_num_workers),
      UCC_CONFIG_TYPE_ULUNITS},
 
     {"EXEC_NUM_THREADS", "512",
-     "Number of thread per block to use for cuda executor",
+     "Number of threads per block to use for cuda persistent executor",
      ucc_offsetof(ucc_ec_cuda_config_t, exec_num_threads),
      UCC_CONFIG_TYPE_ULUNITS},
 
@@ -71,6 +71,12 @@ static ucc_config_field_t ucc_ec_cuda_config_table[] = {
     {"REDUCE_NUM_BLOCKS", "auto",
      "Number of thread blocks to use for reduction in interruptible mode",
      ucc_offsetof(ucc_ec_cuda_config_t, reduce_num_blocks),
+     UCC_CONFIG_TYPE_ULUNITS},
+
+    {"REDUCE_NUM_THREADS", "auto",
+     "Number of threads per block to use for reduction in interruptible "
+     "executor",
+     ucc_offsetof(ucc_ec_cuda_config_t, reduce_num_threads),
      UCC_CONFIG_TYPE_ULUNITS},
 
     {"USE_COOPERATIVE_LAUNCH", "0",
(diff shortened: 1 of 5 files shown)


Sergei-Lebedev (approved):

shimmybalsam (approved):

samnordmann:
bot:retest

samnordmann:
bot:retest

src/components/ec/cuda/ec_cuda.c
```diff
@@ -212,6 +218,27 @@ static ucc_status_t ucc_ec_cuda_post_driver_stream_task(uint32_t *status,
...
 
+static inline void ucc_ec_cuda_set_threads_nbr(int *nt, int maxThreadsPerBlock)
+{
+    if (*nt != UCC_ULUNITS_AUTO) {
+        if (maxThreadsPerBlock < *nt) {
+            ec_warn(
+                &ucc_ec_cuda.super,
+                "number of threads per block is too large, max supported is %d",
+                maxThreadsPerBlock);
+        } else if (*nt % WARP_SIZE != 0) {
```
vspetrov, line 229: plz take (*nt % WARP_SIZE) in ()

src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
```diff
@@ -53,114 +60,173 @@ cuFloatComplex operator* (const cuFloatComplex & first,
...
+                }                                                                 \
+                d[i] = _OP(d[i], s[j][i]);                                        \
+            }                                                                     \
+        }                                                                         \
+        if (flags & UCC_EEE_TASK_FLAG_REDUCE_WITH_ALPHA) {                        \
+            for (i = triggered                                                    \
+                         ? num_lines + threadIdx.x                                \
+                         : num_lines + threadIdx.x + blockIdx.x * blockDim.x;     \
+                 i < count;                                                       \
+                 i += triggered ? blockDim.x : blockDim.x * gridDim.x) {          \
```
vspetrov, line 133: i assume this is optimized by compiler anyways, but i personally would move "triggered ? blockDim.x : blockDim.x * gridDim.x" to the local vars.

src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
```diff
@@ -53,114 +60,173 @@ cuFloatComplex operator* (const cuFloatComplex & first,
...
+                                     : (blockDim.x * gridDim.x) / WARP_SIZE;    \
+        const int    idx       = threadIdx.x % WARP_SIZE;                       \
+        const size_t num_lines =                                                \
+            (count / (WARP_SIZE * UNROLL)) * (WARP_SIZE * UNROLL);              \
+        __shared__ uint16_t n_src2;                                             \
+        _Type               tmp1[UNROLL];                                       \
+        _Type               tmp2[UNROLL];                                       \
+        size_t              i, j;                                               \
+        n_src2 = task.n_src2;                                                   \
+        assert(task.stride % sizeof(_Type) == 0);                               \
```
vspetrov, line 170: why "assert" instead of "ucc_assert_system" ?

vspetrov (approved):
@Sergei-Lebedev @samnordmann  plz remind me what is oldest supported gpu for UCC? Will this LOOP_UNROLL = 6 be a problem for gpu's older than V100 (or we don't need them)?

src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
samnordmann: alright. I store it in the variable `step`. I also defined the variable `start` that stores
```
triggered? num_lines + threadIdx.x  : num_lines + threadIdx.x + blockIdx.x * blockDim.x;  
```

Sergei-Lebedev:
> @Sergei-Lebedev @samnordmann plz remind me what is oldest supported gpu for UCC? Will this LOOP_UNROLL = 6 be a problem for gpu's older than V100 (or we don't need them)?

We support Maxwell and newer. I don't think it's a problem, but maybe not optimal for Maxwell and Pascal.

PR #659, opened 2022-10-27
BUILD: fix UCX version detection

Sergei-Lebedev:
## What
Fixes build time UCX version detection

## Why ?
Need to always check UCX version during build time, before this PR it was only checking if user explicitly sets --with-ucx=ucxdir

### DIFF
--- a/config/m4/ucx.m4
+++ b/config/m4/ucx.m4
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2001-2020, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2001-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # See file LICENSE for terms.
 #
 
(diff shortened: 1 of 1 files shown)


config/m4/ucx.m4
```diff
@@ -67,21 +67,31 @@ AS_IF([test "x$ucx_checked" != "xyes"],[
         ],
         [])
 
+        AS_IF([test "x$ucx_happy" = "xyes"],
```
shimmybalsam, line 70: If statement duplicated in line 88, perhaps combine?

shimmybalsam (approved):

config/m4/ucx.m4
Sergei-Lebedev: it's not duplicate, during version check ucx_happy might become "no", then we don't need to do following checks
```
            AS_IF([test $ucx_major -lt ${UCX_MIN_REQUIRED_MAJOR} && test $ucx_minor -lt ${UCX_MIN_REQUIRED_MINOR}],
            [
                AC_MSG_ERROR([Required UCX version: ${UCX_MIN_REQUIRED_MAJOR}.${UCX_MIN_REQUIRED_MINOR}])
                ucx_happy="no"
            ], [])
```

manjugv (approved):

bureddy (approved):

PR #660, opened 2022-10-27
CORE: unify check for userdefined dt

Sergei-Lebedev:
## What
Add single function to check userdefined datatype. Return UCC_ERR_NOT_SUPPORTED if TL doesn't support userdefined datatype

### DIFF
--- a/src/components/tl/cuda/alltoall/alltoall.c
+++ b/src/components/tl/cuda/alltoall/alltoall.c
@@ -27,9 +27,9 @@ ucc_status_t ucc_tl_cuda_alltoall_init(ucc_base_coll_args_t *coll_args,
         return UCC_ERR_NOT_SUPPORTED;
     }
 
-    task = ucc_tl_cuda_task_init(coll_args, team);
-    if (ucc_unlikely(!task)) {
-        return UCC_ERR_NO_MEMORY;
+    status = ucc_tl_cuda_task_init(coll_args, team, &task);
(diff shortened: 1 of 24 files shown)


src/components/tl/cuda/alltoall/alltoall.c
```diff
@@ -27,6 +27,10 @@ ucc_status_t ucc_tl_cuda_alltoall_init(ucc_base_coll_args_t *coll_args,
         return UCC_ERR_NOT_SUPPORTED;
     }
 
+    if (!ucc_coll_args_is_predefined_dt(&coll_args->args, UCC_RANK_INVALID)) {
```
shimmybalsam, line 30: general comment for all cuda and ucp colls:
why not add this function into cuda_task_init() / ucp_task_init() , similar to how you did in nccl?

manjugv (approved):

manjugv:
@bureddy can you review this? Thanks.

bureddy (approved):

src/components/tl/cuda/alltoall/alltoall.c
Sergei-Lebedev: added for cuda colls, for ucp colls it doesn't make sense since tl ucp will perhaps be the only tl with not predefined dt support

shimmybalsam (approved):

Sergei-Lebedev:
bot:retest

PR #669, opened 2022-11-02
TL/CUDA: reduce scatter linear

Sergei-Lebedev:
Adding linear reduce scatter algorithm in TL CUDA. Each rank broadcasts its source buffer to each other rank following by local reduce operation. Supported only on NVSwtich topology.

### DIFF
--- a/src/components/tl/cuda/allgather/allgather_linear.c
+++ b/src/components/tl/cuda/allgather/allgather_linear.c
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) Mellanox Technologies Ltd. 2022.  ALL RIGHTS RESERVED.
+ * Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See file LICENSE for terms.
  */
@@ -18,6 +18,10 @@ ucc_status_t ucc_tl_cuda_allgather_linear_init(ucc_base_coll_args_t *coll_args,
     status = ucc_tl_cuda_task_init(coll_args, team, &task);
     if (ucc_unlikely(status != UCC_OK)) {
         return status;
+    }
+
+    if (ucc_unlikely(!task)) {
+        return UCC_ERR_NO_MEMORY;
     }
 
     task->allgatherv_linear.get_count  = ucc_tl_cuda_allgather_get_count;
--- a/src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
+++ b/src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
@@ -1,0 +1,49 @@
+/**
+ * Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "reduce_scatterv/reduce_scatterv.h"
+#include "reduce_scatter/reduce_scatter.h"
+
+ucc_status_t ucc_tl_cuda_reduce_scatter_linear_init(ucc_base_coll_args_t *coll_args,
+                                                    ucc_base_team_t *     tl_team,
+                                                    ucc_coll_task_t **    task_p)
+{
+    ucc_tl_cuda_team_t *team   = ucc_derived_of(tl_team, ucc_tl_cuda_team_t);
+    size_t              ssize  = UCC_TL_CUDA_TEAM_LIB(team)->cfg.scratch_size;
+    ucc_datatype_t      dt     = coll_args->args.dst.info.datatype;
+    size_t send_size, frag_size;
+    ucc_tl_cuda_task_t *task;
+    ucc_status_t status;
+
+    if (coll_args->args.op == UCC_OP_AVG) {
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    status = ucc_tl_cuda_task_init(coll_args, team, &task);
+    if (ucc_unlikely(status != UCC_OK)) {
+        return status;
+    }
+
+    task->reduce_scatterv_linear.get_count  = ucc_tl_cuda_reduce_scatter_get_count;
+    task->reduce_scatterv_linear.get_offset = ucc_tl_cuda_reduce_scatter_get_offset;
+    task->reduce_scatterv_linear.dt         = coll_args->args.dst.info.datatype;
+    task->reduce_scatterv_linear.rbuf       = coll_args->args.dst.info.buffer;
+
+    send_size = task->reduce_scatterv_linear.get_count(task, 0);
+    frag_size = ucc_min(ssize / ucc_dt_size(dt) / 2, send_size);
+
+    task->reduce_scatterv_linear.num_frags = ucc_div_round_up(send_size,
+                                                             frag_size);
+    task->super.flags          |= UCC_COLL_TASK_FLAG_EXECUTOR;
+    task->super.post           = ucc_tl_cuda_reduce_scatterv_linear_start;
+    task->super.triggered_post = ucc_triggered_post;
+    task->super.progress       = ucc_tl_cuda_reduce_scatterv_linear_progress;
+    task->super.finalize       = ucc_tl_cuda_reduce_scatterv_linear_finalize;
+    task->bar                  = TASK_BAR(task);
+
+    *task_p = &task->super;
+    return UCC_OK;
+}
--- a/src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
+++ b/src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
@@ -11,10 +11,7 @@ ucc_status_t ucc_tl_cuda_reduce_scatter_linear_init(ucc_base_coll_args_t *coll_a
                                                     ucc_base_team_t *     tl_team,
                                                     ucc_coll_task_t **    task_p)
 {
-    ucc_tl_cuda_team_t *team   = ucc_derived_of(tl_team, ucc_tl_cuda_team_t);
-    size_t              ssize  = UCC_TL_CUDA_TEAM_LIB(team)->cfg.scratch_size;
-    ucc_datatype_t      dt     = coll_args->args.dst.info.datatype;
-    size_t send_size, frag_size;
+    ucc_tl_cuda_team_t *team = ucc_derived_of(tl_team, ucc_tl_cuda_team_t);
     ucc_tl_cuda_task_t *task;
     ucc_status_t status;
 
@@ -27,16 +24,12 @@ ucc_status_t ucc_tl_cuda_reduce_scatter_linear_init(ucc_base_coll_args_t *coll_a
         return status;
     }
 
-    task->reduce_scatterv_linear.get_count  = ucc_tl_cuda_reduce_scatter_get_count;
-    task->reduce_scatterv_linear.get_offset = ucc_tl_cuda_reduce_scatter_get_offset;
+    task->reduce_scatterv_linear.get_count  =
+        ucc_tl_cuda_reduce_scatter_get_count;
+    task->reduce_scatterv_linear.get_offset =
+        ucc_tl_cuda_reduce_scatter_get_offset;
     task->reduce_scatterv_linear.dt         = coll_args->args.dst.info.datatype;
     task->reduce_scatterv_linear.rbuf       = coll_args->args.dst.info.buffer;
-
-    send_size = task->reduce_scatterv_linear.get_count(task, 0);
-    frag_size = ucc_min(ssize / ucc_dt_size(dt) / 2, send_size);
-
-    task->reduce_scatterv_linear.num_frags = ucc_div_round_up(send_size,
-                                                             frag_size);
     task->super.flags          |= UCC_COLL_TASK_FLAG_EXECUTOR;
     task->super.post           = ucc_tl_cuda_reduce_scatterv_linear_start;
     task->super.triggered_post = ucc_triggered_post;
--- a/src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
+++ b/src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
@@ -1,0 +1,449 @@
+/**
+ * Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "reduce_scatterv.h"
+#include "components/ec/ucc_ec.h"
+#include "tl_cuda_cache.h"
+#include "utils/arch/cpu.h"
+#include "utils/arch/cuda_def.h"
+
+/*
+ * fragmented buffered copy linear reduce scatterv algorithm
+ *
+ * Description:
+ *  Definitions:
+ *      blockI  - full send buffer at Rank I
+ *      fragI_J - fragment of send buffer at Rank I and step J
+ *      NS      - number of steps
+ *      NF      - number of fragments
+ *      N       - team size
+ *
+ *  Setup:
+ *      max_frag_size = ucc_min(ucc_max(block1, block2, ..., block N),
+ *                              scratch_size / 2 / N)
+ *      NF            = ucc_max(block1, block2, ..., block N) / max_frag_size
+ *      NS            = 1 + NF
+ *
+ *  Algorithm
+ *      for rank R
+ *      step 1:    copy fragR_1 to remote scratch buffers for all ranks
+ *
+ *      step 1:    reduce frag1_1, frag2_1, ..., fragN_1 from local scratch buffer
+ *                 to local dst buffer
+ *                 copy fragR_2 from local src buffer to remote scratch buffers
+ *                 for all ranks
+ *
+ *      step NS-1: reduce frag1_(NS-2), frag2_(NS-2), ..., fragN_(NS-2) from local
+ *                 scratch buffer to local dst buffer
+ *                 copy fragR_NS from local src buffer to remote scratch buffers
+ *                 for all ranks
+ *
+ *      step NS:   reduce frag1_(NS-1), frag2_(NS-1), ..., fragN_(NS-1) from local
+ *                 scratch buffer to local dst buffer
+ */
+
+enum
+{
+    STAGE_SYNC,    /*< Wait for free SYNC segment */
+    STAGE_SETUP,   /*< Wait for memhandle setup to finish */
+    STAGE_COPIES,  /*< Linear algorithm is running */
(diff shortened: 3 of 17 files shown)


manjugv:
@edgargabriel Can you please review?

manjugv:
Ping @edgargabriel

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+ *
+ *  Setup:
+ *      max_frag_size = ucc_min(ucc_max(block1, block2, ..., block N),
+ *                              scratch_size / 2 / N)
+ *      NF            = ucc_max(block1, block2, ..., block N) / max_frag_size
+ *      NS            = 1 + NF
+ *
+ *  Algorithm
+ *      for rank R
+ *      step 1:    copy fragR_1 to remote scratch buffers for all ranks
```
edgargabriel, line 32: I think based on the code below, this is step 0? Would help avoiding having two different 'step 1' in the description of the algorithm

edgargabriel (approved):
According to my understanding this looks good. I left two minor comments,  both comments are however mere suggestions.

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+            ucc_tl_cuda_reduce_scatterv_linear_reduce(task, exec, sbuf, rbuf,
+                step - 1, local_offset, &task->reduce_scatterv_linear.exec_task[1]);
+        } else {
+            ucc_tl_cuda_reduce_scatterv_linear_copy(task, exec, sbuf, step,
+                remote_offset, &task->reduce_scatterv_linear.exec_task[0]);
+            ucc_tl_cuda_reduce_scatterv_linear_reduce(task, exec, sbuf, rbuf,
+                step - 1, local_offset, &task->reduce_scatterv_linear.exec_task[1]);
+        }
+
+        if ((task->reduce_scatterv_linear.exec_task[0] == NULL) &&
```
samnordmann, line 317: If we remove the if condition on line 259-260 I think we can get rid of this block

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+ *      step 1:    copy fragR_1 to remote scratch buffers for all ranks
+ *
+ *      step 1:    reduce frag1_1, frag2_1, ..., fragN_1 from local scratch buffer
+ *                 to local dst buffer
+ *                 copy fragR_2 from local src buffer to remote scratch buffers
+ *                 for all ranks
+ *
+ *      step NS-1: reduce frag1_(NS-2), frag2_(NS-2), ..., fragN_(NS-2) from local
+ *                 scratch buffer to local dst buffer
+ *                 copy fragR_NS from local src buffer to remote scratch buffers
```
samnordmann, line 41: fix the indices `frag1_(NS-2)` and `frag1_NS`. they should differ by 1

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+        frag_size   = ucc_buffer_block_count(send_size, nfrags, step);
+        frag_offset = ucc_buffer_block_offset(send_size, nfrags, step);
+        rank_offset = task->reduce_scatterv_linear.get_offset(task, i);
+
+        if (frag_size == 0) {
+            continue;
+        }
+        eargs.copy_multi.src[nv]    = PTR_OFFSET(sbuf, (rank_offset + frag_offset) * dt_size);
+        eargs.copy_multi.counts[nv] = frag_size * dt_size;
+        if (trank < i) {
```
samnordmann, line 176: We could replace the whole block by
```
eargs.copy_multi.dst[nv] = PTR_OFFSET(TASK_SCRATCH(task, i), remote_offset + scratch_offset)
eargs.copy_multi.dst[nv] -= (trank < i)? 0 : get_scratch_stride(team, dt)
```

src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "reduce_scatterv/reduce_scatterv.h"
+#include "reduce_scatter/reduce_scatter.h"
+
+ucc_status_t ucc_tl_cuda_reduce_scatter_linear_init(ucc_base_coll_args_t *coll_args,
```
samnordmann, line 10: Is it possible to call directly `ucc_tl_cuda_reduce_scatterv_linear_init`? I have the feeling that this function is nothing but a duplication of what is done in `ucc_tl_cuda_reduce_scatterv_linear_init` and `ucc_tl_cuda_reduce_scatterv_linear_start`

samnordmann (approved):

samnordmann (approved):

src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
```diff
@@ -0,0 +1,43 @@
...
+                                                    ucc_coll_task_t **    task_p)
+{
+    ucc_tl_cuda_team_t *team   = ucc_derived_of(tl_team, ucc_tl_cuda_team_t);
+    ucc_tl_cuda_task_t *task   = ucc_tl_cuda_task_init(coll_args, team);
+    size_t              ssize  = UCC_TL_CUDA_TEAM_LIB(team)->cfg.scratch_size;
+    ucc_datatype_t      dt     = coll_args->args.dst.info.datatype;
+    size_t send_size, frag_size;
+
+    if (ucc_unlikely(!task)) {
+        return UCC_ERR_NO_MEMORY;
```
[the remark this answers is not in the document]
Sergei-Lebedev: you are right, fixed

src/components/tl/cuda/allgather/allgather_linear.c
```diff
@@ -14,6 +14,10 @@ ucc_status_t ucc_tl_cuda_allgather_linear_init(ucc_base_coll_args_t *coll_args,
     ucc_tl_cuda_team_t *team = ucc_derived_of(tl_team, ucc_tl_cuda_team_t);
     ucc_tl_cuda_task_t *task = ucc_tl_cuda_task_init(coll_args, team);
 
+    if (ucc_unlikely(!task)) {
+        return UCC_ERR_NO_MEMORY;
```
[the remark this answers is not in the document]
Sergei-Lebedev: it's already printed in ucc_tl_cuda_task_init()

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+ * (2 * team_size * datatype_size)
+ */
+static inline size_t get_scratch_size(ucc_tl_cuda_team_t *team,
+                                      ucc_datatype_t      dt)
+{
+    size_t     dt_size = ucc_dt_size(dt);
+    ucc_rank_t tsize   = UCC_TL_TEAM_SIZE(team);
+    size_t     div     = 2 * dt_size * tsize;
+
+    return (UCC_TL_CUDA_TEAM_LIB(team)->cfg.scratch_size / div) * div;
```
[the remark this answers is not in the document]
Sergei-Lebedev: do you suggest using actual mapped memory size instead?

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+            (task->reduce_scatterv_linear.exec_task[1] != NULL)) {
+            for (i = 0; i < 2; i++) {
+                etask = task->reduce_scatterv_linear.exec_task[i];
+                if (etask != NULL) {
+                    st = ucc_ee_executor_task_test(etask);
+                    if (st == UCC_OK) {
+                        ucc_ee_executor_task_finalize(etask);
+                        task->reduce_scatterv_linear.exec_task[i] = NULL;
+                    } else {
+                        if (ucc_likely(st > 0)) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: no, status might UCC_OPERATION_INITIALIZED

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+    st = ucc_coll_task_get_executor(&task->super, &exec);
+    if (ucc_unlikely(st != UCC_OK)) {
+        return st;
+    }
+
+    step = get_rank_step(task, trank, 0);
+    while (step < num_steps) {
+        if ((task->reduce_scatterv_linear.exec_task[0] != NULL) ||
+            (task->reduce_scatterv_linear.exec_task[1] != NULL)) {
+            for (i = 0; i < 2; i++) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: i tested it before, didn't see any difference. Actually, both tasks are progressing here, loop just sets order of finalize

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,443 @@
...
+ *
+ *  Algorithm
+ *      for rank R
+ *      step 1:    copy fragR_1 to remote scratch buffers for all ranks
+ *
+ *      step 1:    reduce frag1_1, frag2_1, ..., fragN_1 from local scratch buffer
+ *                 to local dst buffer
+ *                 copy fragR_2 from local src buffer to remote scratch buffers
+ *                 for all ranks
+ *
```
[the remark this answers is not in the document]
Sergei-Lebedev: i just added "..." between steps, hopefully it works :)

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
Sergei-Lebedev: can you pls elaborate what condition to use in lines 259-260 instead?

src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
Sergei-Lebedev: ucc_tl_cuda_reduce_scatter_linear_init sets different pointers for get_count and get_offset. I can remove it and check collective type in ucc_tl_cuda_reduce_scatterv_linear init instead, but I wanted to explicitly show that reduce_scatter exists in TL CUDA. Otherwise, it will be hiddent inside reduce_scatterv

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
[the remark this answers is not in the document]
samnordmann: Sure, I see that both are progressing, it's just that we could hide the latency of finalizing task 1 while task 0 is still being progressed. But if you already tried this and didn't see a difference then we can just discard it.

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
samnordmann: I suggest to completely remove the if condition (and use no condition at all). If both `task[0]` and `task[1]` are `NULL`, then we don't progress any worker and just do
```
step++;
set_rank_step(task, trank, step, 0);
```
so we don't need to handle this case as a special case as it is done currently
Am I correct ?

src/components/tl/cuda/reduce_scatter/reduce_scatter_linear.c
samnordmann: Ok, I see. So do you think it would be a good idea that both `ucc_tl_cuda_reduce_scatter_linear_init` and `ucc_tl_cuda_reduce_scatterv_linear_init` call a common helper function that takes `get_count` and `get_offset` as arguments ?

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,445 @@
...
+ *      step 1:    reduce frag1_0, frag2_0, ..., fragN_0 from local scratch buffer
+ *                 to local dst buffer
+ *                 copy fragR_1 from local src buffer to remote scratch buffers
+ *                 for all ranks
+ *      ...
+ *
+ *      step NS-2: reduce frag1_(NS-3), frag2_(NS-3), ..., fragN_(NS-3) from local
+ *                 scratch buffer to local dst buffer
+ *                 copy fragR_NS from local src buffer to remote scratch buffers
+ *                 for all ranks
```
samnordmann, line 47: I think there is still an index problem here. it should be
```
 copy fragR_(NS-2) from local src buffer to remote scratch buffers
 for all ranks
```

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
samnordmann: I think there is still an index problem here. it should be
```
 copy fragR_(NS-2) from local src buffer to remote scratch buffers
 for all ranks
```

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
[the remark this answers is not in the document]
samnordmann: I am not sure, it would probably require many workarounds for little benefit, so it is maybe simpler to leave it as is for now. What do you think?

samnordmann (approved):

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,450 @@
...
+{
+    ucc_tl_cuda_task_t *task = ucc_derived_of(coll_task, ucc_tl_cuda_task_t);
+
+    tl_trace(UCC_TASK_LIB(task), "finalizing task %p", task);
+    ucc_tl_cuda_task_put(task);
+    return UCC_OK;
+}
+
+/*
+ * get maximum size of scratch memory evenly divisble by
```
vspetrov, line 89: divisible

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,450 @@
...
+ * (2 * team_size * datatype_size)
+ */
+static inline size_t get_scratch_size(ucc_tl_cuda_team_t *team,
+                                      ucc_datatype_t      dt)
+{
+    size_t     dt_size = ucc_dt_size(dt);
+    ucc_rank_t tsize   = UCC_TL_TEAM_SIZE(team);
+    size_t     div     = 2 * dt_size * tsize;
```
vspetrov, line 97: i'd suggest to rename local var "div" since there is a C func with the same name. Call it, e.g., divisor

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,450 @@
...
+}
+
+/* get stride between consecutive rank slots in scratch memory */
+static inline size_t get_scratch_stride(ucc_tl_cuda_team_t *team,
+                                        ucc_datatype_t dt)
+{
+    size_t     step_ssize = get_scratch_size(team, dt) / 2;
+    ucc_rank_t tsize      = UCC_TL_TEAM_SIZE(team);
+
+    return ucc_buffer_block_count(step_ssize, tsize, 0);
```
vspetrov, line 109: why not just step_ssize / tsize? accourding to get_scratch_size, it returns value that is divisible by 2 * dt_size * tsize, hence by tsize as well.

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,450 @@
...
+}
+
+/* get offset of rank slot inside scratch  */
+static inline size_t get_scratch_offset(ucc_tl_cuda_team_t *team,
+                                        ucc_datatype_t dt, ucc_rank_t rank)
+{
+    size_t     step_ssize = get_scratch_size(team, dt) / 2;
+    ucc_rank_t tsize      = UCC_TL_TEAM_SIZE(team);
+
+    return ucc_buffer_block_offset(step_ssize, tsize, rank);
```
vspetrov, line 119: same: step_ssize / tsize  * rank.

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,450 @@
...
+    return UCC_OK;
+
+exit_err:
+    return status;
+}
+
+ucc_status_t
+ucc_tl_cuda_reduce_scatterv_linear_setup_test(ucc_tl_cuda_task_t *task)
+{
+    ucc_tl_cuda_team_t *team = TASK_TEAM(task);
```
vspetrov, line 145: empty line after local vars

src/components/tl/cuda/reduce_scatterv/reduce_scatterv_linear.c
```diff
@@ -0,0 +1,450 @@
...
+    for (i = 0, nv = 0; i < tsize; i++) {
+        if (i == trank) {
+            continue;
+        }
+        send_size   = task->reduce_scatterv_linear.get_count(task, i);
+        frag_size   = ucc_buffer_block_count(send_size, nfrags, step);
+        frag_offset = ucc_buffer_block_offset(send_size, nfrags, step);
+        rank_offset = task->reduce_scatterv_linear.get_offset(task, i);
+
+        if (frag_size == 0) {
```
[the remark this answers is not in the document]
Sergei-Lebedev: yes, simplest case: tesm_size = 2, src.info.count = 2, dst.info_v.counts = [2, 0]. What do you mean nv == tsize- 1? nv is incremented only if frag_size != 0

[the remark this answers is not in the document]
vspetrov: right, got it, i misread it.

vspetrov (approved):

bureddy (approved):

PR #678, opened 2022-11-07
TEST: add linter for TL CUDA

Sergei-Lebedev:
## What
Add new GH actions tests for CUDA. New workflow will run clang-tidy linter for all CUDA components in UCC including TL/CUDA, EC/CUDA, MC/CUDA. Also fixed found warning and errors.
New implementation of ucc_assert is needed to suppress false positive warnings, i.e. telling compiler that variable value within certain range.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,6 +1,7 @@
 #
-# Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
+# See file LICENSE for terms.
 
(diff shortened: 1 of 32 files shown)


src/Makefile.am
```diff
@@ -1,6 +1,7 @@
 #
-# Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
shimmybalsam, line 2: why did you remove 2022?

shimmybalsam (approved):

vspetrov (approved):

bureddy (approved):

PR #697, opened 2022-12-19
testing ubuntu2004 dockerfile

jeniaka:
## What
Adding dockerfile ubuntu 20.04

## Why ?
increase number of distros to run ci


jeniaka:
branch removed

PR #698, opened 2022-12-20
TL/UCP: ranks reordering

Sergei-Lebedev:
## What
Use topology information in TL UCP to reorder ranks so that ranks first grouped by node, by socket within node, by numa within socket. 

## Why ?
Permutation is used in TL UCP allgather to form better ring algorithm.
Performance data collected on N2 PPN48 (AMD EPYC 7443: 24 cores, 2 sockets, 8 NUMAs) 
ompi
ucc no topo
ucc with topo

### DIFF
--- a/src/components/tl/ucp/allgather/allgather.c
+++ b/src/components/tl/ucp/allgather/allgather.c
@@ -20,42 +20,7 @@ ucc_base_coll_alg_info_t
         [UCC_TL_UCP_ALLGATHER_ALG_LAST] = {
             .id = 0, .name = NULL, .desc = NULL}};
 
-static ucc_rank_t ucc_tl_ucp_allgather_ring_get_send_block(ucc_subset_t *subset,
-                                                           ucc_rank_t trank,
-                                                           ucc_rank_t tsize,
-                                                           int step)
-{
-    return ucc_ep_map_eval(subset->map, (trank - step + tsize) % tsize);
-}
-
-static ucc_rank_t ucc_tl_ucp_allgather_ring_get_recv_block(ucc_subset_t *subset,
-                                                           ucc_rank_t trank,
-                                                           ucc_rank_t tsize,
-                                                           int step)
-{
-    return ucc_ep_map_eval(subset->map, (trank - step - 1 + tsize) % tsize);
-}
-
 ucc_status_t ucc_tl_ucp_allgather_init(ucc_tl_ucp_task_t *task)
 {
-    ucc_tl_ucp_team_t *team = TASK_TEAM(task);
-    ucc_sbgp_t *sbgp;
-
-    if (!ucc_coll_args_is_predefined_dt(&TASK_ARGS(task), UCC_RANK_INVALID)) {
-        tl_error(UCC_TASK_LIB(task), "user defined datatype is not supported");
-        return UCC_ERR_NOT_SUPPORTED;
-    }
-
-    if (!(task->flags & UCC_TL_UCP_TASK_FLAG_SUBSET)) {
-        if (team->topo) {
-            sbgp = ucc_topo_get_sbgp(team->topo, UCC_SBGP_FULL_HOST_ORDERED);
-            task->subset.myrank = sbgp->group_rank;
-            task->subset.map    = sbgp->map;
-        }
-    }
-    task->allgather_ring.get_send_block = ucc_tl_ucp_allgather_ring_get_send_block;
-    task->allgather_ring.get_recv_block = ucc_tl_ucp_allgather_ring_get_recv_block;
-    task->super.post                    = ucc_tl_ucp_allgather_ring_start;
-    task->super.progress                = ucc_tl_ucp_allgather_ring_progress;
-    return UCC_OK;
+    return ucc_tl_ucp_allgather_ring_init_common(task);
 }
--- a/src/components/tl/ucp/allgather/allgather.c
+++ b/src/components/tl/ucp/allgather/allgather.c
@@ -20,14 +20,42 @@ ucc_base_coll_alg_info_t
         [UCC_TL_UCP_ALLGATHER_ALG_LAST] = {
             .id = 0, .name = NULL, .desc = NULL}};
 
+static ucc_rank_t ucc_tl_ucp_allgather_ring_get_send_block(ucc_subset_t *subset,
+                                                           ucc_rank_t trank,
+                                                           ucc_rank_t tsize,
+                                                           int step)
+{
+    return ucc_ep_map_eval(subset->map, (trank - step + tsize) % tsize);
+}
+
+static ucc_rank_t ucc_tl_ucp_allgather_ring_get_recv_block(ucc_subset_t *subset,
+                                                           ucc_rank_t trank,
+                                                           ucc_rank_t tsize,
+                                                           int step)
+{
+    return ucc_ep_map_eval(subset->map, (trank - step - 1 + tsize) % tsize);
+}
+
 ucc_status_t ucc_tl_ucp_allgather_init(ucc_tl_ucp_task_t *task)
 {
+    ucc_tl_ucp_team_t *team = TASK_TEAM(task);
+    ucc_sbgp_t *sbgp;
+
     if (!ucc_coll_args_is_predefined_dt(&TASK_ARGS(task), UCC_RANK_INVALID)) {
         tl_error(UCC_TASK_LIB(task), "user defined datatype is not supported");
         return UCC_ERR_NOT_SUPPORTED;
     }
 
-    task->super.post     = ucc_tl_ucp_allgather_ring_start;
-    task->super.progress = ucc_tl_ucp_allgather_ring_progress;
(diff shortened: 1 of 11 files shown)


src/components/topo/ucc_sbgp.c
```diff
@@ -386,6 +456,9 @@ ucc_status_t ucc_sbgp_create(ucc_topo_t *topo, ucc_sbgp_type_t type)
     case UCC_SBGP_FULL:
         status = sbgp_create_full(topo, sbgp);
         break;
+    case UCC_SBGP_FULL_HOST_ORDERED:
+        status = sbgp_create_full_ordered(topo, sbgp);
```
bureddy, line 460: is it host list order preserved when doing this ordering?

Sergei-Lebedev: it sorts hosts by hashes, for instance
**original mapping**:
rank0 - hostA (hash: 0xB) socket 0
rank1 - hostA (hash: 0xB) socket 1
rank2 - hostA (hash: 0xB) socket 0
rank3 - hostA (hash: 0xB) socket 1
rank4 - hostB (hash: 0xA) socket 0
rank5 - hostB (hash: 0xA) socket 1
rank6 - hostB (hash: 0xA) socket 0
rank7 - hostB (hash: 0xA) socket 1
**reordering**:
rank4, rank6, rank5, rank7, rank0, rank2, rank1, rank3

src/components/tl/ucp/allgather/allgather.c
```diff
@@ -16,14 +16,42 @@ ucc_base_coll_alg_info_t
...
     if (!ucc_coll_args_is_predefined_dt(&TASK_ARGS(task), UCC_RANK_INVALID)) {
         tl_error(UCC_TASK_LIB(task), "user defined datatype is not supported");
         return UCC_ERR_NOT_SUPPORTED;
     }
 
-    task->super.post     = ucc_tl_ucp_allgather_ring_start;
-    task->super.progress = ucc_tl_ucp_allgather_ring_progress;
+    if (!(task->flags & UCC_TL_UCP_TASK_FLAG_SUBSET)) {
+        if (team->topo) {
+            sbgp = ucc_topo_get_sbgp(team->topo, UCC_SBGP_FULL_HOST_ORDERED);
```
vspetrov, line 47: I would enable that functionality with variable (ON by default). Sometimes, user still wants to "force" by-node mapping and there should be an option to do so, imho.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -71,8 +71,15 @@ extern const char
         }                                                                      \
     } while (0)
 
+
+enum ucc_tl_ucp_task_flags {
+    /*indicates whether subset field of tl_ucp_task is set*/
+    UCC_TL_UCP_TASK_FLAG_SUBSET = UCC_BIT(0),
```
vspetrov, line 77: do we need this flag? the original idea was that subset is set to EP_MAP_FULL by default. So if collective supports it, then it always uses "subset" which evaluates to no-op for FULL. Otherwise works as expected.

src/components/topo/ucc_sbgp.c
```diff
@@ -370,6 +371,75 @@ static inline ucc_status_t sbgp_create_full(ucc_topo_t *topo, ucc_sbgp_t *sbgp)
...
+
+    qsort(sorted, gsize, sizeof(proc_info_id_t), ucc_compare_proc_info_id);
+    for (i = 0; i < gsize; i++) {
+        if (sorted[i].id == topo->set.myrank) {
+            sbgp->group_rank = i;
+        }
+        sbgp->rank_map[i] = sorted[i].id;
+    }
+
+    /*TODO: try to detect map by numa,socket,node and use UCC_EP_MAP_CB to save
```
vspetrov, line 435: you should use ucc_ep_map_from_array to try to pack the ep map (detect strided/flat). Alloc int *rank, fill in line 430, and then ucc_ep_map_from_array.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -143,6 +150,23 @@ typedef struct ucc_tl_ucp_task {
...
         } allgather_kn;
+        struct {
+            /*
+             * get send/recv block depends on subset type being used.
+             * For service allgather we need to get context endpoints but keep
+             * subset numbering.
+             * For regular allgather with rank reordering both endpoints
+             * and blocks permutation are necessary.
+             */
+            ucc_rank_t (*get_send_block)(ucc_subset_t *subset,
```
vspetrov, line 161: we could maybe try to remove that conditionals for services vs regular. IF we introduce tl_ucp_team->reorder_map explciitly rathar than updating subset. Then:
```
allgather_ring_init() {
     if (!team_is_single_node(tl_ucp_team) && cfg->allgather_use_host_ordering) {
             if (task->subset == FULL) {
                   sbgp = get_sbgp(HOST_ORDERED);
                   task->reorder_map = sbgp.map;
             } else {
                   task->reorder_map = FULL; // this is true for service 
             }
     }
```
then in the collective you would use double mapping team->subset->reorder_map ? Now that i think of it, we maybe even can udpate subset as ucc_ep_map_create_nested(subset, reorder_map) and it will work? (need to check)

src/components/tl/ucp/allgather/allgather.c
Sergei-Lebedev: added

src/components/tl/ucp/tl_ucp_coll.h
Sergei-Lebedev: it's used to identify if subset is already set and shouldn't be changed. For example, in service collective.

src/components/topo/ucc_sbgp.c
Sergei-Lebedev: yes, it's already used in ucc_sbgp_create, but it can't detect map by numa, socket or node because they are not really strided

src/components/tl/ucp/tl_ucp_coll.h
Sergei-Lebedev: not sure i understand, semantic is diffrrent for service allgather and regular allgather. Service allgather with subset looks to me like a lightweight allgather with new subteam and so new topo based permutation required in that case.

vspetrov: yeah.. i was thinking how to avoid that difference of flows between service and regular. At first i thought we can unify them, but looks like we can't. we need separate maps for "peer"  and "block" computations for the 2 cases. so basically i meant to be explcit: 1 thing is "subset", it running an algorithm over only a handful of procs out of the team; another thin is "reordering". 

Suppose we would introduce "reorder_map" default to FULL, optionally set to get_sbgp(HOST_ORDERED).map. Then in your allgather implementation you would have:
```
    sendto   = ucc_ep_map_eval(task->subset.map, (trank + 1) % tsize);
    recvfrom = ucc_ep_map_eval(task->subset.map, (trank - 1 + tsize) % tsize);
    sendto   = ucc_ep_map_eval(task->reorder.map, sendto);
    recvfrom = ucc_ep_map_eval(task->reorder.map, recvfrom);

   ...
   get_recv_block {
     ...
   ucc_ep_map_eval(task->reorder_map, (trank - step - 1 + tsize) % tsize);
   }
```

So, this way you don't have fn pointers for send/get block. the behavior is achieved by setting reorder_map = FULL for service. Then it would even could be possible to support reordering for subset (though i don't see a use case for that now. maybe if we would like to support arbitrary size active sets we could use re-ordering for them).

Anyways, this was my logic. And when i wrote this i quickly thought that maybe this even allows to build nested map (my comment above) - but that was incorrect.

manjugv (approved):

src/components/tl/ucp/tl_ucp_coll.h
Sergei-Lebedev: code snippet above is not correct, subset.map is actually reverse permutation that map from new numbering to old numbering, subset.myrank is rank number in new numbering. In ring algorithm we first need to find neighbors in new numbering and then translate it back to old numbering to know EPs. However, if we want to combine multiple permutations only reverse permutation is not enough, we also need forward permutation that translates from old numbering to new numbering.
Since we are not going to use subset and reordering simultaneously anyway, maybe add union to tl_ucp task structure:
```
union {
  ucc_subset_t subset;
  ucc_subset_t reordering;
} 
```
If algorithm supports both subset and reordering it will use subset, otherwise it will use reordering field. In those algorithms that support both we will need somehow to check if rank permutation is enough, or we also need block permutation. This can be done by providing function pointers to blocks call functions or adding flag that says to permute ranks only and don't touch blocks.

manjugv:
@vspetrov are you ok with the changes?

manjugv:
ping @vspetrov

vspetrov (approved):

Sergei-Lebedev:
bot:retest

PR #700, opened 2022-12-22
EC/CUDA: memops supported by default in cuda 12.0

bureddy:
## Why ?
CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS removed in cuda-12.0 and memops supported by default.


Sergei-Lebedev (commented):
need to fix cuda.m4 to use correct gencode flags for CUDA12

Sergei-Lebedev (approved):

vspetrov (approved):

samnordmann (approved):

PR #702, opened 2022-12-28
TL/UCP: get range from team cfg instead of lib

shimmybalsam:
## What
Get allreduce radix range from team cfg instead of lib cfg

## Why ?
lib cfg will have either default value or user value from cmd. Team cfg handles config file values to be applied as well.

## How ?
In team creation if config file is available and tuning turned on, team cfg will be updated with config file values.


shimmybalsam:
Depeneds on #701

bureddy (approved):

Sergei-Lebedev (approved):

PR #703, opened 2022-12-28
EC/CUDA: unify reduce and reduce_strided kernels

samnordmann:
Simplify and unifies the implementation of the reduce kernels in EC/CUDA.
- reduce and reduce_strided kernels are unified into one, at the cost of an extra template parameter
- the part with the unrolled loop and the remainder are unified into one macro `CUDA_REDUCE_WITH_OP_CHUNK` that we call twice
- No perf degradation, cf graphs

reducedt_float32_a100_plot
reducedt_float32-T_a100_plot
reducedt_float64_a100_plot
reducedt_float64-T_a100_plot
reducedt_int32_a100_plot
reducedt_int32-T_a100_plot
reducedt_strided_float32_a100_plot
reducedt_strided_float32-T_a100_plot
reducedt_strided_float64_a100_plot
reducedt_strided_float64-T_a100_plot
reducedt_strided_int32_a100_plot
reducedt_strided_int32-T_a100_plot

### DIFF
--- a/src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
+++ b/src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
@@ -59,102 +59,101 @@ cuFloatComplex operator* (const cuFloatComplex & first,
                                cuCimagf(first) * second);
 }
 
-#define CUDA_REDUCE_WITH_OP_CHUNK(offset, unroll, warp_size, _OP)\
-    do{\
-        const int    warp =                                                     \
-            triggered ? threadIdx.x / warp_size                                 \
-                        : (threadIdx.x + blockIdx.x * blockDim.x) / warp_size; \
-        const int    num_warps = triggered                                      \
-                                    ? blockDim.x / warp_size                   \
-                                    : (blockDim.x * gridDim.x) / warp_size;    \
-        const int    idx       = threadIdx.x % warp_size;                       \
-        const size_t num_lines =                                                \
-            (count / (warp_size * unroll)) * (warp_size * unroll);              \
-        _Type     tmp1[unroll];                                                 \
-        _Type     tmp2[unroll];                                                 \
-        for (line = offset + warp * warp_size * unroll + idx; line < num_lines;   \
-            line += num_warps * warp_size * unroll) {                          \
-            _Pragma("unroll") for (i = 0; i < unroll; i++)                      \
-            {                                                                   \
-                tmp1[i] = s1[line + warp_size * i];                           \
-            }                                                                   \
-            for (j = 0; j < MAXSRCS; j++) {                                     \
-                if (j >= n_src2) {                                              \
-                    break;                                                      \
-                }                                                               \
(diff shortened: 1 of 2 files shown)


src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
```diff
@@ -15,7 +15,6 @@ extern "C" {
 #include "ec_cuda_half_sm52.h"
 #include <cuda_bf16.h>
 #include <cuComplex.h>
-#include <assert.h>
```
shimmybalsam, line 18: is this not needed for line 134?

src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
```diff
@@ -60,165 +59,101 @@ cuFloatComplex operator* (const cuFloatComplex & first,
...
+            s2     = (_Type *)task.src2;                                         \
+            ld     = task.stride / sizeof(_Type);                                \
+            ucc_assert_system(task.stride % sizeof(_Type) == 0);                 \
+        } else {                                                                 \
+            memcpy(s, task.srcs, UCC_EE_EXECUTOR_NUM_BUFS * sizeof(_Type *));    \
+            n_src2 = task.n_srcs - 1;                                            \
+            s1     = s[0];                                                       \
+        }                                                                        \
+        CUDA_REDUCE_WITH_OP_CHUNK(0, UNROLL, WARP_SIZE, _OP);                    \
+        CUDA_REDUCE_WITH_OP_CHUNK(                                               \
```
shimmybalsam, line 141: perhaps add comment between the two calls of CUDA_REDUCE_WITH_OP_CHUNK:
 /* second call for data remainder */ 
 
 Also, is there always remainder? should second CUDA_REDUCE_WITH_OP_CHUNK always be called?

src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
samnordmann: the header is already included in ucc_assert.h

src/components/ec/cuda/kernel/ec_cuda_reduce_ops.h
samnordmann: Ok I add a comment.
There is not always a remainder (no remainder if `count` is a multiple of `UNROLL*WARP_SIZE`). However, there is no harm in calling `CUDA_REDUCE_WITH_OP_CHUNK` two times anyway because if there is no remainder we will just not enter the for loop.

shimmybalsam: Ok, possibly just a matter of personal taste, but you can precompute offset before second call and only do second call if offset == 0. 
Not sure which way is more "costly" - adding an extra if branch or adding an extra call to the stack.

vspetrov (approved):

Sergei-Lebedev (approved):

PR #704, opened 2022-12-29
CORE: check tl is reachable

Sergei-Lebedev:
## What
Check if TL context exists in every rank before creating tl team.

## Why ?
Because TL contexts list might not be the same for every rank it's possible to have allgather collision when different ranks trying to create different teams e.g.
```
tl_ucp_coll.c:79   TL_UCP ERROR failure in send completion Message truncated
```

## How ?
During address exchange stage in team create or context create collect list of TLs available for each team rank.

### DIFF
--- a/src/core/ucc_context.c
+++ b/src/core/ucc_context.c
@@ -530,8 +530,11 @@ poll:
     {
         /* Compute storage rank and check proc info uniqeness */
         ucc_rank_t r = UCC_RANK_MAX;
-        ucc_context_addr_header_t *h, *h2;
+        int j;
+        ucc_context_addr_header_t *h, *h0;
 
+        addr_storage->flags = UCC_ADDR_STORAGE_FLAG_TLS_SYMMETRIC;
+        h0 = UCC_ADDR_STORAGE_RANK_HEADER(addr_storage, 0);
         for (i = 0; i < addr_storage->size; i++) {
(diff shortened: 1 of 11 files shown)


shimmybalsam (approved):

manjugv (approved):

bureddy (approved):

src/core/ucc_context.c
```diff
@@ -544,10 +541,27 @@ ucc_status_t ucc_core_addr_exchange(ucc_context_t          *context,
...
                     return UCC_ERR_NO_MESSAGE;
                 }
-                r = (ucc_rank_t)i;
+                r = i;
             }
         }
         addr_storage->rank = r;
+        /*check if TLs arrays is symmetric*/
+        addr_storage->flags = UCC_ADDR_STORAGE_FLAG_TLS_SYMMETRIC;
+        h = UCC_ADDR_STORAGE_RANK_HEADER(addr_storage, 0);
```
vspetrov, line 550: do you really need to iterate over addr_storage 2nd time? could you just do the same in the loop above (adding if (i > 1)) ?

vspetrov (approved):

src/core/ucc_context.c
Sergei-Lebedev: merged two loops

PR #707, opened 2023-01-04
TEST: add configurable dt for non-reduction colls

samnordmann:
## What
Add datatypes for non-reduction collectives in mpi_test

## Why
This patch is necessary for testing different datatype as the user would do, even if the collectives do not involve reduction

### DIFF
--- a/test/mpi/test_alltoall.cc
+++ b/test/mpi/test_alltoall.cc
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See file LICENSE for terms.
  */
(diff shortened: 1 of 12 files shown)


test/mpi/test_alltoall.cc
```diff
@@ -54,21 +53,21 @@ TestAlltoall::TestAlltoall(ucc_test_team_t &_team, TestCaseParams &params) :
...
+    args.dst.info.datatype    = dt;
     args.dst.info.mem_type    = mem_type;
     UCC_CHECK(set_input());
     UCC_CHECK_SKIP(ucc_collective_init(&args, &req, team.team), test_skip);
 }
 
 ucc_status_t TestAlltoall::set_input(int iter_persistent)
 {
-    size_t dt_size = ucc_dt_size(TEST_DT);
+    size_t      dt_size           = ucc_dt_size(dt);
```
shimmybalsam, line 70: alignment

test/mpi/test_mpi.h
```diff
@@ -380,13 +383,16 @@ class UccTestMpi {
 };
 
 class TestAllgather : public TestCase {
-public:
+    ucc_datatype_t dt;
```
shimmybalsam, line 386: not needed as is protected in parent class. same for all collectives.

shimmybalsam (changes requested):

test/mpi/test_mpi.h
```diff
@@ -176,6 +176,7 @@ void set_gpu_device(test_set_gpu_device_t set_device);
 #endif
 int ucc_coll_inplace_supported(ucc_coll_type_t c);
 int ucc_coll_is_rooted(ucc_coll_type_t c);
+bool ucc_coll_has_datatype(ucc_coll_type_t c);
```
shimmybalsam, line 179: perhaps should be changed to int similar to inplace_supported and is_rooted?

shimmybalsam (approved):

vspetrov (approved):

Sergei-Lebedev (approved):

PR #709, opened 2023-01-10
TL/CUDA: use pci bdf to build topo

Sergei-Lebedev:
## What
Use GPU PCI address instead of GPU ID to collect topology information in TL/CUDA

## Why ?
GPU ID (from cudaGetDevice() )  is not guaranteed to be unique within single node, for instance if user sets CUDA_VISIBLE_DEVICES=$OMPI_COMM_WORLD_LOCAL_RANK every rank will see one GPU with ID 0.

### DIFF
--- a/src/components/tl/cuda/tl_cuda_team_topo.c
+++ b/src/components/tl/cuda/tl_cuda_team_topo.c
@@ -236,6 +236,7 @@ ucc_tl_cuda_team_topo_init_proxies(const ucc_tl_cuda_team_t *team,
     float *data;
     float score, min_score;
     ucc_status_t status;
+    char pci_str[2][MAX_PCI_BUS_ID_STR];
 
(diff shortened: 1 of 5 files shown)


src/components/tl/cuda/tl_cuda_team_topo.c
```diff
@@ -236,6 +236,7 @@ ucc_tl_cuda_team_topo_init_proxies(const ucc_tl_cuda_team_t *team,
     float *data;
     float score, min_score;
     ucc_status_t status;
+    char pci_str[2][MAX_PCI_BUS_ID_STR];
```
samnordmann, line 239: copyright

samnordmann (approved):

manjugv (approved):

bureddy (approved):

PR #727, opened 2023-02-07
DOCS: aligning conf file

shimmybalsam:
## What
Aligning ucc.conf to be similar to that in ucc_private.

## Why ?
For future users who use this file as a template/example for their configurations. This way they will see up to date multiple architectures tuning.

### DIFF
--- a/contrib/ucc.conf
+++ b/contrib/ucc.conf
@@ -21,6 +21,8 @@ UCC_CL_HIER_NET_SBGP_TLS=^sharp,nccl,cuda
 # with ucp sbgp on top
 UCC_CL_HIER_FULL_SBGP_TLS=ucp
 
+# Tuning sections, currently only supports TL/UCP
+#Intel Broadwell:
 [vendor=intel model=broadwell team_size=28 ppn=28 nnodes=1]
 UCC_TL_UCP_ALLREDUCE_KN_RADIX=2
 UCC_TL_UCP_ALLREDUCE_SRA_KN_RADIX=7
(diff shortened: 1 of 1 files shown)


manjugv:
@shimmybalsam should we call this ucc_ompi.conf since we have OMPI variables?

shimmybalsam:
> @shimmybalsam should we call this ucc_ompi.conf since we have OMPI variables?

@manjugv  I don't think so. I think it should be a unified example file for any user.
The OMPI prefix was added in ucc_private due to interference of those specific variables in certain cases (for example pytorch). 
For this file I can remove OMPI from the variables so that they will start with UCC. What do you think?

manjugv:
> > @shimmybalsam should we call this ucc_ompi.conf since we have OMPI variables?
> 
> @manjugv I don't think so. I think it should be a unified example file for any user. The OMPI prefix was added in ucc_private due to interference of those specific variables in certain cases (for example pytorch). For this file I can remove OMPI from the variables so that they will start with UCC. What do you think?

Yes, that sounds better.

shimmybalsam:
> > > @shimmybalsam should we call this ucc_ompi.conf since we have OMPI variables?
> > 
> > 
> > @manjugv I don't think so. I think it should be a unified example file for any user. The OMPI prefix was added in ucc_private due to interference of those specific variables in certain cases (for example pytorch). For this file I can remove OMPI from the variables so that they will start with UCC. What do you think?
> 
> Yes, that sounds better.

OK fixed.
@manjugv @Sergei-Lebedev please rereview thanks.

Sergei-Lebedev (approved):

manjugv (approved):

shimmybalsam:
bot:retest

shimmybalsam:
bot:retest

shimmybalsam:
bot:retest

PR #729, opened 2023-02-13
TL/UCP: add knomial allgather

Sergei-Lebedev:
## What
Add knomial allgather algorithm to TL/UCP

## Why ?
Gives better performance for small message  comparing to existing ring algorithm in TL UCP

## How ?
Porting changes from old PR https://github.com/openucx/ucc/pull/361, initially implemented by @vspetrov


vspetrov (approved):
Would be nice to add some small example data of perf.

manjugv (approved):

PR #731, opened 2023-02-15
API: Clarify team create semantics

manjugv:
The change is clarifying the semantics of ucc_team_create_post. No changes to the API, it is agreeing with the existing semantics (as captured in the docs).

https://openucx.github.io/ucc/api/v1.0/pdf/ucc.pdf


vspetrov:
@Sergei-Lebedev what about "lazy ompi comm creation" ? do you need to allow multiple concurrent team creations in case when the lazy ucc team post is triggered from non-blocking coll?

Sergei-Lebedev:
> @Sergei-Lebedev what about "lazy ompi comm creation" ? do you need to allow multiple concurrent team creations in case when the lazy ucc team post is triggered from non-blocking coll?

yes, this is what we discussed on yesterday meeting, we will continue discussion next week

Sergei-Lebedev (approved):

vspetrov (approved):

PR #741, opened 2023-03-03
TEST: return error if all tests skipped

edgargabriel:
## What
return an error if all tests have been skipped, this is most likely not the expected behavior and indicates a problem.

## Why ?
This issue came up when analyzing how the dlopen error with rccl tl has been missed by our CI. The reason turned out to be that the CI still returned SUCCESS despite of the rccl tl not being able to dlopen and execute correctly, but since no test formally 'failed' but all were skipped, no error code was returned.


Sergei-Lebedev (approved):

Sergei-Lebedev:
@edgargabriel please add copyright header

manjugv (approved):

PR #743, opened 2023-03-06
TL/UCP: remove memcpy in last SRA step

Sergei-Lebedev:
## What
Avoid memory copy in last step of reduce scatter.

## How ?
In last step of reduce scatter local reduce operation is split into strided reduce of received data and reduce of result of that operation and local data.

### DIFF
--- a/src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
+++ b/src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See file LICENSE for terms.
  */
@@ -15,6 +15,23 @@
         task->reduce_scatter_kn.phase = _phase;                                \
     } while (0)
 
+static inline void get_sbuf_rbuf(ucc_knomial_pattern_t *p, ucc_coll_args_t *args,
+                                 void *scratch, size_t block_count,
+                                 void **sbuf, void **rbuf)
+{
+    uint8_t node_type = p->node_type;
+    size_t  dt_size   = ucc_dt_size(args->dst.info.datatype);
+
+    if (ucc_knomial_pattern_loop_first_iteration(p)) {
+        *sbuf = (KN_NODE_PROXY == node_type || UCC_IS_INPLACE(*args))
+                 ? args->dst.info.buffer: args->src.info.buffer;
+        *rbuf = scratch;
+    } else {
+        *sbuf = scratch;
+        *rbuf = PTR_OFFSET(*sbuf, block_count * dt_size);
+    }
+}
+
 void ucc_tl_ucp_reduce_scatter_knomial_progress(ucc_coll_task_t *coll_task)
 {
     ucc_tl_ucp_task_t          *task       = ucc_derived_of(coll_task,
@@ -23,7 +40,7 @@ void ucc_tl_ucp_reduce_scatter_knomial_progress(ucc_coll_task_t *coll_task)
     ucc_tl_ucp_team_t          *team       = TASK_TEAM(task);
     ucc_kn_radix_t              radix      = task->reduce_scatter_kn.p.radix;
     int                         avg_pre_op =
-        UCC_TL_UCP_TEAM_LIB(team)->cfg.reduce_avg_pre_op;
+    UCC_TL_UCP_TEAM_LIB(team)->cfg.reduce_avg_pre_op;
     uint8_t                     node_type  =
         task->reduce_scatter_kn.p.node_type;
     ucc_knomial_pattern_t      *p          = &task->reduce_scatter_kn.p;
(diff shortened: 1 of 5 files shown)


src/utils/ucc_dt_reduce.h
```diff
@@ -33,7 +33,7 @@ ucc_dt_reduce_strided(void *src1, void *src2, void *dst, size_t n_vectors,
 {
     ucc_ee_executor_task_args_t eargs;
 
-    if (count == 0) {
+    if (count == 0 || n_vectors == 0) {
```
shimmybalsam, line 36: why?

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
```diff
@@ -180,25 +178,42 @@ void ucc_tl_ucp_reduce_scatter_knomial_progress(ucc_coll_task_t *coll_task)
...
+
         ucc_sra_kn_get_offset_and_seglen(count, dt_size, rank, size, radix,
                                          &offset, &local_seg_count);
-        eargs.task_type = UCC_EE_EXECUTOR_TASK_COPY;
-        eargs.copy.dst  = PTR_OFFSET(args->dst.info.buffer, offset);
-        eargs.copy.src  = task->reduce_scatter_kn.scratch;
-        eargs.copy.len  = local_seg_count * dt_size;
-        status = ucc_ee_executor_task_post(task->reduce_scatter_kn.executor, &eargs,
-                                           &task->reduce_scatter_kn.etask);
+        status = ucc_dt_reduce(local_data, rbuf,
```
vspetrov, line 203: I seem to misunderstand the change. So, in case of !scratch (full pow radix case + no INPLACE) on the last reduction you do it in 2 steps: first you add all the buffers received during last step, place result in the beginning of the "rbuf" (which is dst.buffer). Then you do 1 more step in the end: reduce local data with that intermediate result and place the answer right back rbuf. Right?

How is that different from doing the entire reduce_strided with step_radix - 1?

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
```diff
@@ -180,25 +178,42 @@ void ucc_tl_ucp_reduce_scatter_knomial_progress(ucc_coll_task_t *coll_task)
...
-        ucc_knomial_pattern_next_iteration(p);
+        ucc_kn_rs_pattern_next_iter(p);
     }
 
     if (!task->reduce_scatter_kn.scratch_mc_header) {
+        ucc_knomial_pattern_prev_iteration(p);
+        get_sbuf_rbuf(p, args, task->reduce_scatter_kn.scratch, block_count,
+                      &sbuf, &rbuf);
+
+        step_radix = ucc_kn_compute_step_radix(p);
```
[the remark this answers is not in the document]
Sergei-Lebedev: no, block_count is different

src/utils/ucc_dt_reduce.h
Sergei-Lebedev: if n_vectors is 0 we have no data to reduce

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
```diff
@@ -71,105 +87,87 @@ void ucc_tl_ucp_reduce_scatter_knomial_progress(ucc_coll_task_t *coll_task)
...
-                                                 &offset, &local_seg_count);
-                reduce_data = PTR_OFFSET(args->dst.info.buffer, offset);
+                status = ucc_dt_reduce_strided(
+                    rbuf, PTR_OFFSET(rbuf, local_seg_count * dt_size), rbuf,
+                    step_radix - 2, local_seg_count, local_seg_count * dt_size,
+                    dt, args, 0, 0, task->reduce_scatter_kn.executor,
+                    &task->reduce_scatter_kn.etask);
+
+            } else {
+                if (task->reduce_scatter_kn.scratch_mc_header &&
```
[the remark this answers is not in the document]
Sergei-Lebedev: maybe i missing something, there 3 different cases 
* last iteration and don't have scratch - partial reduction
* last iteration and have scratch  - full reduction since receive data is in scratch
* not last iteration - full reduction

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
Sergei-Lebedev: yes, it's correct, the difference is that in this PR we do step_radix -1 reductions while in prev version it was step_radix -1 reduction plus memory copy
i see big improvement for cases where this memory copy takes a lot of time for example 
image

vspetrov:
Discussed over phone with Sergey. He is to check if the change works correctly in the case when count % comm_size != 0.

Sergei-Lebedev:
> Discussed over phone with Sergey. He is to check if the change works correctly in the case when count % comm_size != 0.

@vspterov, I've changed the way offsets are calculated please have a look

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
```diff
@@ -15,14 +15,57 @@
...
+            correct destination directly
+        */
+        local_seg_offset /= dt_size;
+        if (local_seg_offset <= block_count || local_seg_count == 0) {
+            *sbuf = scratch;
+            *rbuf = PTR_OFFSET(*sbuf, block_count * dt_size);
+        } else {
+            offset = (local_seg_offset - block_count) % local_seg_count;
+            *sbuf = scratch;
+            *rbuf = PTR_OFFSET(*sbuf, (block_count + offset) * dt_size);
```
vspetrov, line 53: looks good to me. just one question: if scratch is actual user rbuf (no extra + no_inplace case) do we always have enough space after "block_count" to recv segement. I think so but haven't proved on paper. Maybe add assert, like:
block_ount + offset + local_seg_count <= count.

vspetrov (approved):

src/components/tl/ucp/reduce_scatter/reduce_scatter_knomial.c
Sergei-Lebedev: I think I checked all the corner cases and in every case where rbuf is used we have enough space. I will merge this in, let's see if any tests will fail until next release

PR #746, opened 2023-03-06
TL/SHARP: SHARP OOB fixes

bureddy:
- hide sharp lib errors
  - disable lazy init by default
  - removed hard coded sharp IB device.

## Why ?
- users are seeing sharp lib errors related to the application trying to contact sharp_am service but sharp is not enabled in the setup.
- sharp groups are limited resources and sharp team creation needs to know resource availability to avoid failure during the collective post.

### DIFF
--- a/src/components/tl/sharp/tl_sharp.c
+++ b/src/components/tl/sharp/tl_sharp.c
@@ -30,7 +30,7 @@ static ucc_config_field_t ucc_tl_sharp_context_config_table[] = {
     {"", "", NULL, ucc_offsetof(ucc_tl_sharp_context_config_t, super),
      UCC_CONFIG_TYPE_TABLE(ucc_tl_context_config_table)},
 
-    {"DEVICES", "mlx5_0:1",
+    {"DEVICES", "",
      "SHARP device list",
      ucc_offsetof(ucc_tl_sharp_context_config_t, dev_list),
      UCC_CONFIG_TYPE_STRING},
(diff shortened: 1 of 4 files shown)


src/components/tl/sharp/tl_sharp_context.c
```diff
@@ -300,6 +300,17 @@ ucc_status_t ucc_tl_sharp_context_init(ucc_tl_sharp_context_t *sharp_ctx,
...
+#if HAVE_DECL_SHARP_COLL_HIDE_ERRORS
+    if (lib->super.super.log_component.log_level < UCC_LOG_LEVEL_DEBUG) {
+        init_spec.config.flags               |= SHARP_COLL_HIDE_ERRORS;
+    }
+#endif
+#if HAVE_DECL_SHARP_COLL_DISABLE_LAZY_GROUP_RESOURCE_ALLOC
+    if(!sharp_ctx->cfg.enable_lazy_group_alloc) {
+        init_spec.config.flags               |= SHARP_COLL_DISABLE_LAZY_GROUP_RESOURCE_ALLOC;
+    }
+#endif
```
Sergei-Lebedev, line 312: print warning if disabling lazy group alloc is not supported by sharp and user has changed default value

src/components/tl/sharp/tl_sharp.c
```diff
@@ -30,7 +30,7 @@ static ucc_config_field_t ucc_tl_sharp_context_config_table[] = {
     {"", "", NULL, ucc_offsetof(ucc_tl_sharp_context_config_t, super),
      UCC_CONFIG_TYPE_TABLE(ucc_tl_context_config_table)},
 
-    {"DEVICES", "mlx5_0:1",
+    {"DEVICES", "",
```
Sergei-Lebedev, line 33: how device will be selected for previous SHARP releases?

Sergei-Lebedev (approved):

src/components/tl/sharp/tl_sharp_context.c
bureddy: maybe we can expose the user option also only when flag is there

src/components/tl/sharp/tl_sharp.c
Sergei-Lebedev: right, this is my question, if one builds UCC 1.2 with SHARP 3.0 how it will work without providing device?

bureddy: the default selection was fixed in libsharp in the last HPCX release  i. it will fail if we use older than that. that should be ok

Sergei-Lebedev (approved):

manjugv (approved):

PR #766, opened 2023-04-10
TL/UCP: opt radix alg

shimmybalsam:
## What
Adding auto (sub) optimal radix calculation to be used for Allreduce SRA and Bcast SAG inside TL/UCP.

## Why ?
To improve performance when degredation caused by extra ranks.

## How ?
Calculates the radix which minimizes extra ranks per team size.
Additionally makes sure that radix wont be larger then socket size (across all nodes), to also minimize cross socket communication.

### DIFF
--- a/src/coll_patterns/recursive_knomial.h
+++ b/src/coll_patterns/recursive_knomial.h
@@ -8,6 +8,7 @@
 #define RECURSIVE_KNOMIAL_H_
 
 #define UCC_KN_PEER_NULL ((ucc_rank_t)-1)
+#define UCC_KN_MIN_RADIX 2
 typedef uint16_t ucc_kn_radix_t;
 
 enum {
@@ -228,6 +229,40 @@ ucc_knomial_calc_recv_dist(ucc_rank_t team_size, ucc_rank_t rank,
         dist *= radix;
     }
     return dist;
+}
+
+/* Calculates (sub) opt radix for Allreduce SRA and Bcast SAG,
+   by minimizing n_extra ranks */
+static inline ucc_rank_t ucc_kn_get_opt_radix(ucc_rank_t team_size,
+                                              ucc_rank_t max_radix)
+{
+    ucc_rank_t remainder = 0, n_trees = 0, min_val = 0,
+               min_i = UCC_KN_MIN_RADIX;
+    ucc_rank_t min_trees, r, fs;
+
+    max_radix = (max_radix >= UCC_KN_MIN_RADIX) ? max_radix : UCC_KN_MIN_RADIX;
+    min_trees = max_radix;
+
+    for (r = UCC_KN_MIN_RADIX; r <= max_radix; r++) {
+        fs = r;
+        while (fs < team_size) {
+            fs = fs * r;
+        }
+        fs        = (fs == team_size) ? fs : fs / r;
+        n_trees   = team_size / fs;
+        remainder = team_size - (team_size / fs) * fs;
+        if (remainder == 0) {
+            return r;
+        }
+        if (r == UCC_KN_MIN_RADIX || (r > UCC_KN_MIN_RADIX &&
+            (remainder < min_val ||
+            (remainder == min_val && n_trees < min_trees)))) {
+            min_val   = remainder;
+            min_trees = n_trees;
+            min_i     = r;
+        }
+    }
+    return min_i;
 }
 
 /* A set of convenience macros used to implement sw based progress
--- a/src/coll_patterns/recursive_knomial.h
+++ b/src/coll_patterns/recursive_knomial.h
@@ -236,31 +236,24 @@ ucc_knomial_calc_recv_dist(ucc_rank_t team_size, ucc_rank_t rank,
 static inline ucc_rank_t ucc_kn_get_opt_radix(ucc_rank_t team_size,
                                               ucc_kn_radix_t max_radix)
 {
-    ucc_rank_t     remainder = 0, n_trees = 0, min_val = 0;
-    ucc_kn_radix_t min_i     = UCC_KN_MIN_RADIX;
+    ucc_rank_t     n_extra = 0, min_val = team_size;
+    ucc_kn_radix_t min_i   = UCC_KN_MIN_RADIX;
+    ucc_kn_radix_t max_r   = ucc_max(max_radix, UCC_KN_MIN_RADIX);
     ucc_kn_radix_t r, fs;
-    ucc_rank_t     min_trees;
 
-    max_radix = ucc_max(max_radix, UCC_KN_MIN_RADIX);
-    min_trees = max_radix;
-
-    for (r = UCC_KN_MIN_RADIX; r <= max_radix; r++) {
+    for (r = UCC_KN_MIN_RADIX; r <= max_r; r++) {
         fs = r;
         while (fs < team_size) {
             fs = fs * r;
         }
-        fs        = (fs == team_size) ? fs : fs / r;
-        n_trees   = team_size / fs;
-        remainder = team_size - (team_size / fs) * fs;
-        if (remainder == 0) {
+        fs      = (fs == team_size) ? fs : fs / r;
+        n_extra = team_size - (team_size / fs) * fs;
+        if (n_extra == 0) {
             return r;
         }
-        if (r == UCC_KN_MIN_RADIX || (r > UCC_KN_MIN_RADIX &&
-            (remainder < min_val ||
-            (remainder == min_val && n_trees < min_trees)))) {
-            min_val   = remainder;
-            min_trees = n_trees;
-            min_i     = r;
+        if (n_extra < min_val) {
+            min_val = n_extra;
+            min_i   = r;
         }
     }
     return min_i;
--- a/src/coll_patterns/recursive_knomial.h
+++ b/src/coll_patterns/recursive_knomial.h
@@ -234,13 +234,14 @@ ucc_knomial_calc_recv_dist(ucc_rank_t team_size, ucc_rank_t rank,
 /* Calculates (sub) opt radix for Allreduce SRA and Bcast SAG,
    by minimizing n_extra ranks */
 static inline ucc_rank_t ucc_kn_get_opt_radix(ucc_rank_t team_size,
-                                              ucc_rank_t max_radix)
+                                              ucc_kn_radix_t max_radix)
 {
-    ucc_rank_t remainder = 0, n_trees = 0, min_val = 0,
-               min_i = UCC_KN_MIN_RADIX;
-    ucc_rank_t min_trees, r, fs;
+    ucc_rank_t     remainder = 0, n_trees = 0, min_val = 0;
+    ucc_kn_radix_t min_i     = UCC_KN_MIN_RADIX;
+    ucc_kn_radix_t r, fs;
+    ucc_rank_t     min_trees;
 
-    max_radix = (max_radix >= UCC_KN_MIN_RADIX) ? max_radix : UCC_KN_MIN_RADIX;
+    max_radix = ucc_max(max_radix, UCC_KN_MIN_RADIX);
     min_trees = max_radix;
 
     for (r = UCC_KN_MIN_RADIX; r <= max_radix; r++) {
--- a/src/components/tl/ucp/tl_ucp_coll.h
+++ b/src/components/tl/ucp/tl_ucp_coll.h
@@ -409,14 +409,15 @@ static inline unsigned
 ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
                                 size_t             msgsize,
                                 ucc_memory_type_t  mem_type,
-                                ucc_mrange_uint_t *p)
+                                ucc_mrange_uint_t *p,
+                                ucc_rank_t         default_value)
 {
     unsigned radix;
 
     radix = ucc_mrange_uint_get(p, msgsize, mem_type);
 
     if (UCC_UUNITS_AUTO == radix) {
-        return team->opt_radix;
+        return default_value;
     }
     return radix;
 }
--- a/src/components/tl/ucp/tl_ucp_coll.h
+++ b/src/components/tl/ucp/tl_ucp_coll.h
@@ -417,6 +417,10 @@ ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
 
     if (UCC_UUNITS_AUTO == radix) {
         /* auto selection based on team configuration */
+        if (UCC_TL_UCP_TEAM_CTX(team)->topo_required &&
+            !IS_SERVICE_TEAM(team) && team->cfg.calc_opt_radix) {
+            return team->opt_radix;
+        }
         return UCC_UUNITS_AUTO_RADIX;
     }
     return radix;
--- a/src/components/tl/ucp/tl_ucp_coll.h
+++ b/src/components/tl/ucp/tl_ucp_coll.h
@@ -416,12 +416,7 @@ ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
     radix = ucc_mrange_uint_get(p, msgsize, mem_type);
 
     if (UCC_UUNITS_AUTO == radix) {
-        /* auto selection based on team configuration */
-        if (team->topo && team->topo->topo->sock_bound &&
-            !IS_SERVICE_TEAM(team)) {
-            return team->opt_radix;
-        }
-        return UCC_UUNITS_AUTO_RADIX;
+        return team->opt_radix;
     }
     return radix;
 }
--- a/src/components/tl/ucp/tl_ucp_coll.h
+++ b/src/components/tl/ucp/tl_ucp_coll.h
@@ -417,8 +417,8 @@ ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
 
     if (UCC_UUNITS_AUTO == radix) {
         /* auto selection based on team configuration */
-        if (UCC_TL_UCP_TEAM_CTX(team)->topo_required &&
-            !IS_SERVICE_TEAM(team) && team->cfg.calc_opt_radix) {
+        if (team->topo && team->topo->topo->sock_bound &&
+            !IS_SERVICE_TEAM(team)) {
             return team->opt_radix;
         }
         return UCC_UUNITS_AUTO_RADIX;
--- a/src/components/tl/ucp/tl_ucp_team.c
+++ b/src/components/tl/ucp/tl_ucp_team.c
@@ -46,8 +46,8 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_team_t, ucc_base_context_t *tl_context,
 {
     ucc_tl_ucp_context_t *ctx =
         ucc_derived_of(tl_context, ucc_tl_ucp_context_t);
-    ucc_rank_t   max_radix;
-    ucc_status_t status;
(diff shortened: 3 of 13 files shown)


src/components/topo/ucc_sbgp.c
```diff
@@ -504,6 +532,9 @@ ucc_status_t ucc_sbgp_create(ucc_topo_t *topo, ucc_sbgp_type_t type)
         sbgp->map = ucc_ep_map_from_array(&sbgp->rank_map, sbgp->group_size,
                                           ucc_subset_size(&topo->set), 1);
     }
+    if (sbgp->rank_map && sbgp->status == UCC_SBGP_NOT_EXISTS) {
+        ucc_free(sbgp->rank_map);
```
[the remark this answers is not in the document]
shimmybalsam: Not related to this PR specifically - but found issue during debug, map wasn't freed in this edge case, even though it should be.

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -416,6 +416,10 @@ ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
 
     if (UCC_UUNITS_AUTO == radix) {
         /* auto selection based on team configuration */
+        if (UCC_TL_UCP_TEAM_CTX(team)->topo_required &&
+            !IS_SERVICE_TEAM(team) && team->cfg.calc_opt_radix) {
```
Sergei-Lebedev, line 420: do we still need service team?

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -86,6 +121,19 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_team_t, ucc_base_context_t *tl_context,
...
     }
+    if (self->topo && !IS_SERVICE_TEAM(self) && self->cfg.calc_opt_radix) {
+        status = ucc_sbgp_create(self->topo, UCC_SBGP_NODE_LEADERS);
+        if (UCC_OK != status) {
+            ucc_ep_map_destroy_nested(&self->ctx_map);
+            ucc_topo_cleanup(self->topo);
+            return status;
+        }
+        max_radix       = ucc_min(UCC_TL_TEAM_SIZE(self),
+                                  self->topo->min_socket_size);
```
Sergei-Lebedev, line 132: don't access it directly instead pls add function ucc_topo_min_socket_size similar to ucc_topo_min_ppn

src/components/topo/ucc_sbgp.c
```diff
@@ -203,18 +205,33 @@ static ucc_status_t sbgp_create_node_leaders(ucc_topo_t *topo, ucc_sbgp_t *sbgp,
         return UCC_ERR_NO_MEMORY;
     }
 
+    nl_array_3 = ucc_malloc(comm_size * nnodes * sizeof(ucc_rank_t),
```
Sergei-Lebedev, line 208: seems like too much memory, we cannot have number of sockets more than comm_size

src/components/topo/ucc_sbgp.c
```diff
@@ -203,18 +205,33 @@ static ucc_status_t sbgp_create_node_leaders(ucc_topo_t *topo, ucc_sbgp_t *sbgp,
...
+    for (i = 0; i < nnodes * comm_size; i++) {
+        nl_array_3[i] = 0;
+    }
 
     for (i = 0; i < comm_size; i++) {
-        ucc_rank_t    ctx_rank = ucc_ep_map_eval(set->map, i);
-        ucc_host_id_t host_id  = topo->topo->procs[ctx_rank].host_id;
+        ucc_rank_t      ctx_rank  = ucc_ep_map_eval(set->map, i);
+        ucc_host_id_t   host_id   = topo->topo->procs[ctx_rank].host_id;
+        ucc_socket_id_t socket_id = topo->topo->procs[ctx_rank].socket_id;
```
Sergei-Lebedev, line 229: how do we handle case when rank is not socket bound?

shimmybalsam: fixed.

samnordmann (approved):

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -416,6 +416,10 @@ ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
 
     if (UCC_UUNITS_AUTO == radix) {
         /* auto selection based on team configuration */
+        if (team->topo && team->topo->topo->sock_bound &&
+            !IS_SERVICE_TEAM(team)) {
+            return team->opt_radix;
+        }
         return UCC_UUNITS_AUTO_RADIX;
```
Sergei-Lebedev, line 423: maybe just
```suggestion
        return team->opt_radix;
```

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -86,6 +89,15 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_team_t, ucc_base_context_t *tl_context,
                  "topo is not available, disabling ranks reordering");
         self->cfg.use_reordering = 0;
     }
+
+    if (self->topo && !IS_SERVICE_TEAM(self) && self->topo->topo->sock_bound) {
+        max_radix       = ucc_min(UCC_TL_TEAM_SIZE(self),
```
Sergei-Lebedev, line 94: don't you need to check if ucc_topo_min_socket_size is valid?

shimmybalsam:
bot:retest

shimmybalsam:
bot:retest

Sergei-Lebedev (approved):

src/components/topo/ucc_topo.h
```diff
@@ -126,6 +134,58 @@ static inline int ucc_topo_isoppn(ucc_topo_t *topo)
     return ucc_topo_max_ppn(topo) == ucc_topo_min_ppn(topo);
 }
 
+/* Returns min socket size across the nodes */
+static inline ucc_rank_t ucc_topo_min_socket_size(ucc_topo_t *topo)
+{
+    ucc_sbgp_t *sbgp = ucc_topo_get_sbgp(topo, UCC_SBGP_NODE_LEADERS);
+
+    if (sbgp->status == UCC_SBGP_NOT_INIT) {
+        ucc_assert(topo->topo->sock_bound);
```
vspetrov, line 143: why is this assertion here? what if node leaders sbgp is successfully created but proc is not sock/numa bound ? will this functino return topo->min_socket_size = 0? should the caller of the fn treat return 0 as non correct value?

src/components/tl/ucp/tl_ucp_coll.h
```diff
@@ -415,8 +415,7 @@ ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
     radix = ucc_mrange_uint_get(p, msgsize, mem_type);
 
     if (UCC_UUNITS_AUTO == radix) {
-        /* auto selection based on team configuration */
-        return UCC_UUNITS_AUTO_RADIX;
+        return team->opt_radix;
```
vspetrov, line 418: this assumes that we have single opt_radix for all potential usages of any mrange_uint radix parameter (for any algorithm). I don't think this assumption is true in all cases. originally the fn would return AUTO_RADIX flag that would mean for the caller to use special logic - potentially specific for algorithm for example. that was more clean imho.

src/components/topo/ucc_topo.h
shimmybalsam: This function - ucc_topo_min_socket_size() - should only be called if sock bound. there are similar functions for numa as well.
If user calls ucc_topo_get_sbgp(topo, UCC_SBGP_NODE_LEADERS) outside of this function then he wont reach this assertion.
In any case  topo->min_socket_size is set by default to UCC_MAX_RADIX (not to 0), if there is an error along the way it is set to UCC_RANK_INVALID. User should check. In our case we take in team ucc_min() of the result and of team size, team size will always be smaller then UCC_MAX_RADIX/UCC_RANK_INVALID so there won't be an issue and no need to directly check.

vspetrov: i see. i'd say it is convoluted and a not obvious. lets at least document this in a comment. also, if we say that user always has to call this fn with sock/numa bound then assertion should be moved outside of "if (stats == NOT_INIT)" to the top of the fn. correct?

src/components/tl/ucp/tl_ucp_coll.h
vspetrov: Yes, but what i'm saying is different. This interface does not allow specifying different optimal radixes for say Bcast and Allreduce.

src/coll_patterns/recursive_knomial.h
```diff
@@ -230,6 +231,41 @@ ucc_knomial_calc_recv_dist(ucc_rank_t team_size, ucc_rank_t rank,
...
+        while (fs < team_size) {
+            fs = fs * r;
+        }
+        fs        = (fs == team_size) ? fs : fs / r;
+        n_trees   = team_size / fs;
+        remainder = team_size - (team_size / fs) * fs;
+        if (remainder == 0) {
+            return r;
+        }
+        if (r == UCC_KN_MIN_RADIX || (r > UCC_KN_MIN_RADIX &&
```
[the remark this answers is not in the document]
shimmybalsam: you're right second condition of minimizing trees is NOT needed I will remove.

src/components/tl/ucp/tl_ucp_coll.h
shimmybalsam: Can you elaborate why? 
From what I understand, if if statement in line 417 is true that UCC_UUNITS_AUTO == radix that means that irrelevant to which coll, radix was not set by user in cmd flags or in ucc.conf, so it can only now have been set by opt radix or not at all in which case team->opt_radix will either be set to something specific or will be UCC_UUNITS_AUTO_RADIX. 
What is the difference if we check which case, and if not set return UCC_UUNITS_AUTO_RADIX directly?

vspetrov: Right now you can only compute single value of for opt_radix for tl/ucp team. This opt radix is based currently on your minimizing alg for radix. it is the same for allreduce/bcast etc. So if user keeps AUTO for radix variable, then for all collectives it will use the same team->opt_radix value.

Now assume that you have 2 different algorithms for opt radix calcuation : 1 for allreduce and another for bcast. Then what you could do with the former implementation:
you call ucc_tl_ucp_get_radix_from_range from bcast/allreduce, and get UUNITS_AUTO_RADIX as output. Then you understand (at tl/ucp team level) that you need auto selectin. Then you call custom functions: e.g. tl_ucp_bcast_opt_radix() ro tl_ucp_allreduce_opt_radix() - and they can return different values.

In your PR you can't do that because now ucc_tl_ucp_get_radix_from_range  always returns only 1 possible value team->opt_radix

shimmybalsam: Fixed. Let me know if you have further comments on this.

src/coll_patterns/recursive_knomial.h
```diff
@@ -230,6 +231,35 @@ ucc_knomial_calc_recv_dist(ucc_rank_t team_size, ucc_rank_t rank,
...
+        fs = r;
+        while (fs < team_size) {
+            fs = fs * r;
+        }
+        fs        = (fs == team_size) ? fs : fs / r;
+        remainder = team_size - (team_size / fs) * fs;
+        if (remainder == 0) {
+            return r;
+        }
+        if (r == UCC_KN_MIN_RADIX || remainder < min_val) {
```
vspetrov, line 255: i still don't see why not setting min_val = team_size at line 239, and then just keep this condition:
```
if (remainder < min_val) {
    min_val = remainder;
    min_i     = r;
}
```
and again, for me n_extra is better name than remainder, though i'm not insisting.

src/components/tl/ucp/tl_ucp_team.c
```diff
@@ -86,6 +97,15 @@ UCC_CLASS_INIT_FUNC(ucc_tl_ucp_team_t, ucc_base_context_t *tl_context,
...
         self->cfg.use_reordering = 0;
     }
+    if (self->topo && !IS_SERVICE_TEAM(self) && self->topo->topo->sock_bound) {
+        max_radix       = ucc_min(UCC_TL_TEAM_SIZE(self),
+                                  ucc_topo_min_socket_size(self->topo));
+
+        self->opt_radix = ucc_kn_get_opt_radix(UCC_TL_TEAM_SIZE(self),
+                                               max_radix);
+    }
+    ucc_tl_ucp_switch_auto_radix_in_range(self);
```
vspetrov, line 107: i think this is bad solution.

I think your change now just breaks the parameter setting option for user. If user sets UCC_TL_UCP_SRA_RADIX=3, then during parsing mranged value would get "default_value=3" and here you just overwrite it to your opt_radix value ignoring the fact that user explictely set it.

What i suggested was: don't modify mrage parameter interface, neither alter its content after user it was parsed.
Instead modify ucc_tl_ucp_get_radix_from_range function. Let take the last argument "default_radix" that it will return for AUTO case, smthe like:
```
static inline unsigned
ucc_tl_ucp_get_radix_from_range(ucc_tl_ucp_team_t *team,
                                size_t             msgsize,
                                ucc_memory_type_t  mem_type,
                                ucc_kn_radix_t       default_radix,
                                ucc_mrange_uint_t *p)
{
    unsigned radix;

    radix = ucc_mrange_uint_get(p, msgsize, mem_type);

    if (UCC_UUNITS_AUTO == radix) {
           return default_radix;
    }
    return radix;
}
```
then in ucc_tl_ucp_allreduce_sra_knomial_frag_init you can do:
```
cfg_radix = ucc_tl_ucp_get_radix_from_range(tl_team,
                                                count * ucc_dt_size(dtype),
                                                mem_type, team->opt_radix, p);
```
same for bcast. later you can have custom opt radixes for different collectives: team->sra_allreduce_opt_radix, team->bcast_sag_opt_radix, etc.

For kn allreduce you can just keep default 4 rthrough parser interface or extend same way as proposed here.

shimmybalsam: My current implementation doesn't break user param setting, since it will update default value only for auto case (also if user sets UCC_TL_UCP_ALLREDUCE_SRA_KN_RADIX=0-64k:3,auto , then it will only use default value for this auto case.

But, I will say that it is more correct to use your suggestion since this way we get this updated value when only coll needs it as apposed to my solution which will set it in team create and then again fetch it during coll. So I will change to your solution.

src/coll_patterns/recursive_knomial.h
shimmybalsam: You're correct, changed (also to n_extra :) )

src/components/tl/ucp/tl_ucp_team.c
vspetrov: have you verified you statement above?
If add this line:
```
(base) vpetrov@NB-1996-wsl:~/workspace/ucc$ git diff
diff --git a/src/components/tl/ucp/allreduce/allreduce_sra_knomial.c b/src/components/tl/ucp/allreduce/allreduce_sra_knomial.c
index 1e95409..ee4abcb 100644
--- a/src/components/tl/ucp/allreduce/allreduce_sra_knomial.c
+++ b/src/components/tl/ucp/allreduce/allreduce_sra_knomial.c
@@ -119,6 +119,7 @@ static ucc_status_t ucc_tl_ucp_allreduce_sra_knomial_frag_init(
                                                   UCC_TL_TEAM_SIZE(tl_team),
                                                   count);

+printf("starting SRA with radix: %d\n", radix);
     /* 1st step of allreduce: knomial reduce_scatter */
     UCC_CHECK_GOTO(
         ucc_tl_ucp_reduce_scatter_knomial_init_r(&args, team, &task, radix),
```
Then run with current master:
```
(base) vpetrov@NB-1996-wsl:~/workspace/ucc$ mpirun --bind-to none -x xUCC_TL_UCP_ALLREDUCE_SRA_KN_RADIX=2  -np 3 ./install/bin/ucc_test_mpi -c allreduce -I 0 -d float32 -t world -O 0 -m 65536:65536 -o sum

===== UCC MPI TEST INFO =======
   seed        : 12970
===============================

starting SRA with radix: 3
starting SRA with radix: 3
starting SRA with radix: 3

===== UCC MPI TEST REPORT =====
   total tests : 1
   passed      : 1
   skipped     : 0
   failed      : 0
   elapsed     : 0s
(base) vpetrov@NB-1996-wsl:~/workspace/ucc$ mpirun --bind-to none -x UCC_TL_UCP_ALLREDUCE_SRA_KN_RADIX=2  -np 3 ./install/bin/ucc_test_mpi -c allreduce -I 0 -d float32 -t world -O 0 -m 65536:65536 -o sum

===== UCC MPI TEST INFO =======
   seed        : 12976
===============================

starting SRA with radix: 2
starting SRA with radix: 2
starting SRA with radix: 2

===== UCC MPI TEST REPORT =====
   total tests : 1
   passed      : 1
   skipped     : 0
   failed      : 0
   elapsed     : 0s
```
Note how setting RADIX=2 actually sets it.
Now running with your PR:
```
(base) vpetrov@NB-1996-wsl:~/workspace/ucc$ git show HEAD --quiet
commit f14ca7259a9d796601466bd6d65853070638330c (HEAD -> shimmy)
Author: Shimmy Balsam <sbalsam@nvidia.com>
Date:   Tue Jun 13 21:13:43 2023 +0300

    REVIEW: third review fixes
(base) vpetrov@NB-1996-wsl:~/workspace/ucc$ mpirun --bind-to none -x UCC_TL_UCP_ALLREDUCE_SRA_KN_RADIX=2  -np 3 ./install/bin/ucc_test_mpi -c allreduce -I 0 -d float32 -t world -O 0 -m 65536:65536 -o sum

===== UCC MPI TEST INFO =======
   seed        : 13064
===============================

starting SRA with radix: 3
starting SRA with radix: 3
starting SRA with radix: 3

===== UCC MPI TEST REPORT =====
   total tests : 1
   passed      : 1
   skipped     : 0
   failed      : 0
   elapsed     : 0s
```
Note, how setting RADIX=2 is ignored.

I don't see how " it will update default value only for auto case" this is true. You call  ucc_tl_ucp_switch_auto_radix_in_range unconditionally. So mranged param "default" value will get overwritten either to UCC_UUINTS_AUTO_RADIX which is 4, or to your team->opt_radix.

src/components/topo/ucc_topo.h
```diff
@@ -126,6 +134,74 @@ static inline int ucc_topo_isoppn(ucc_topo_t *topo)
...
+
+/* Returns max socket size across the nodes.
+ * If not set will return 0,
+ * in case of error will return UCC_RANK_INVALID.
+ */
+static inline ucc_rank_t ucc_topo_max_socket_size(ucc_topo_t *topo)
+{
+    ucc_assert(topo->topo->sock_bound);
+
+     ucc_sbgp_t *sbgp = ucc_topo_get_sbgp(topo, UCC_SBGP_NODE_LEADERS);
```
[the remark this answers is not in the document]
shimmybalsam: added declaration to top but kept assert before calling ucc_topo_get_sbgp().

src/components/tl/ucp/tl_ucp_team.c
shimmybalsam: switched to your solution

vspetrov (approved):

shimmybalsam:
bot:retest

shimmybalsam:
bot:retest

PR #768, opened 2023-04-17
TL/UCP: use pairwise pattern in knomial step

Sergei-Lebedev:
## What
Iterate over peers in knomial groups with pairwise style loop

## Why ?
Better matching of send/recv pairs inside knomial group


Sergei-Lebedev:
bot:retest

manjugv (approved):

bureddy (approved):

PR #771, opened 2023-05-08
TL/UCP: avoid copy in knomial scatter

Sergei-Lebedev:
## What
Adding configuration parameter for knomial scatter algorithm (part of scatter allgather broadcast) to receive data directly into destination buffer and avoid memory copy as a last step of the algorithm

## Why ?
Improves performance of large message broadcast. By default enabled only for GPU collective

### DIFF
--- a/src/components/tl/ucp/scatter/scatter_knomial.c
+++ b/src/components/tl/ucp/scatter/scatter_knomial.c
@@ -45,11 +45,12 @@ void ucc_tl_ucp_scatter_knomial_progress(ucc_coll_task_t *coll_task)
     ucc_rank_t             peer, vroot, vpeer, peer_recv_dist;
     ucc_rank_t             step_radix, peer_seg_index, local_seg_index;
     ptrdiff_t              peer_seg_offset, offset;
-    ucc_status_t           status;
     ucc_kn_radix_t         loop_step;
     size_t                 block_count, peer_seg_count, local_seg_count;
+    ucc_status_t           status;
 
     root = VRANK(root, root, size);
+    sbuf = (rank == root) ? args->src.info.buffer : args->dst.info.buffer;
 
     if (task->scatter_kn.phase == UCC_SCATTER_KN_PHASE_LOOP) {
         goto UCC_SCATTER_KN_PHASE_LOOP;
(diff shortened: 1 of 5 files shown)


src/components/tl/ucp/scatter/scatter_knomial.c
```diff
@@ -62,13 +63,7 @@ void ucc_tl_ucp_scatter_knomial_progress(ucc_coll_task_t *coll_task)
     while (!ucc_knomial_pattern_loop_done(p)) {
         step_radix  = ucc_kn_compute_step_radix(p);
         block_count = ucc_sra_kn_compute_block_count(count, rank, p);
-        sbuf        = (rank == root)
-                           ? args->src.info.buffer : args->dst.info.buffer;
-        rbuf        = args->dst.info.buffer;
```
shimmybalsam, line 67: where are you setting rbuf instead?

src/components/tl/ucp/scatter/scatter_knomial.c
```diff
@@ -174,6 +174,16 @@ ucc_status_t ucc_tl_ucp_scatter_knomial_start(ucc_coll_task_t *coll_task)
     task->scatter_kn.recv_dist = ucc_knomial_calc_recv_dist(size - p->n_extra,
                                                             vrank, p->radix,
                                                             vroot);
+    task->scatter_kn.recv_offset = 0;
+    is_zcopy = UCC_TL_UCP_TEAM_LIB(team)->cfg.scatter_kn_enable_recv_zcopy;
+    if (((is_zcopy == UCC_CONFIG_AUTO) &&
+         (args->src.info.mem_type != UCC_MEMORY_TYPE_HOST)) ||
```
shimmybalsam, line 180: why not on host?

src/components/tl/ucp/scatter/scatter_knomial.c
Sergei-Lebedev: can you elaborate pls, i don't understand the question

src/components/tl/ucp/scatter/scatter_knomial.c
Sergei-Lebedev: this tricky, depending on CPU vendor I see different results. For AMD it's better to have extra copy and for Intel it's better to receive with correct offset. For GPUs it's always better to avoid copy

src/components/tl/ucp/scatter/scatter_knomial.c
shimmybalsam: nevermind, i see it is already set in the begining of progress and this is redundant.

shimmybalsam (approved):

vspetrov (approved):

PR #773, opened 2023-05-11
TEST: check op and dt compatibility

Sergei-Lebedev:
## What
Check operation is supported for given datatype in ucc_test_mpi

### DIFF
--- a/test/mpi/test_mpi.cc
+++ b/test/mpi/test_mpi.cc
@@ -308,7 +308,7 @@ int ucc_coll_reduce_supported(ucc_reduction_op_t op, ucc_datatype_t dt)
     case UCC_DT_BFLOAT16:
     case UCC_DT_FLOAT128:
         return (op == UCC_OP_SUM || op == UCC_OP_PROD || op == UCC_OP_MAX ||
-                op == UCC_OP_MIN);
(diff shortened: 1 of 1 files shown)


manjugv (approved):

test/mpi/test_mpi.cc
```diff
@@ -288,6 +288,36 @@ void UccTestMpi::set_ops(std::vector<ucc_reduction_op_t> &_ops)
...
+    case UCC_DT_UINT64:
+    case UCC_DT_UINT128:
+        return (op != UCC_OP_AVG);
+    case UCC_DT_FLOAT16:
+    case UCC_DT_FLOAT32:
+    case UCC_DT_FLOAT64:
+    case UCC_DT_BFLOAT16:
+    case UCC_DT_FLOAT128:
+        return (op == UCC_OP_SUM || op == UCC_OP_PROD || op == UCC_OP_MAX ||
+                op == UCC_OP_MIN);
```
marsaev, line 311: does that mean that `UCC_OP_AVG` for float types is disabled?

bureddy (approved):

test/mpi/test_mpi.cc
Sergei-Lebedev: you are right, fixed, thanks

PR #156, opened 2021-04-09
TL collective plugin interface

vspetrov:
## What
PR adds the interface for the custom/closed/vendor plugins inside TL, i.e. plugin at the collective algorithm implementation level

## Why ?
The use case is: given the existing open source TL (TL/UCP is the most obvious example) provide a way for 3rd party implement an algorithm re-using TL/UCP resources and functionality and distribute this algorithm as a closed plug-in.

## How ?
The tl level iface is provided. Added a necessary logic to the build process that searches for the plugins. The build of plugins can be enabled with "--with-tlcp", where tlcp stands for "tl collective plugin".
Example reference implementation of a TL/UCP plugin is added in components/tl/ucp/coll_plugins/example.

The development flow for the 3rd party: fork UCC repo, add a "git submodule" with the code of the plugin. The code base of a plugin can be stored separately. The sync with the main ucc becomes easy and smooth: git pull. Since plugin code is in a separate folder (submodule) no merge issues. 

The integration of a plugin algorithm into score-based selection logic is made naturally. Plugin provides "get_scores" interface and thus reports to the TL dering team creation. This allows vendor to define which collective, which msg range, which mem type thier plugin will run on. And this can be altered in runtime with the same parameter "SCORE". For the example plugin above it is: UCC_TL_UCP_TLCP_EXAMPLE_SCORE.

Plugin may implement several algorithms for different collectives if needed. So, no more than 1 plugin from single vendor is required.

PR #786, opened 2023-05-29
TL/UCP: extended collective plugin

Sergei-Lebedev:
## What
Extending collective plugin interface (https://github.com/openucx/ucc/pull/156) with functions to create and destroy custom plugin context.
Added fully function example of knomial allreduce using active messages of UCX.


Sergei-Lebedev:
cc @gvallee @BrodyWilliams

BrodyWilliams (approved):

PR #798, opened 2023-06-21
TL/NCCL make ncclGroupEnd nb

shimmybalsam:
## What
check nccl status after using ncclGroupEnd , as well as for ncclSend/ncclRecv if they are called outside of ncclGroupStart/End.

## Why ?
To fit nb change from using ncclCommInitRankConfig from PR #772

### DIFF
--- a/src/components/tl/nccl/tl_nccl.c
+++ b/src/components/tl/nccl/tl_nccl.c
@@ -39,8 +39,9 @@ static ucs_config_field_t ucc_tl_nccl_context_config_table[] = {
      UCS_CONFIG_TYPE_ENUM(ucc_tl_nccl_completion_sync_names)
     },
 
-    {"NCCL_CFG_BLOCKING", "1",
-     "If set to 0 will use non-blocking mode, if set to 1 will use blocking",
+    {"BLOCKING", "1",
+     "If set to 0 will use non-blocking mode communicator behavior, "
+     "if set to 1 will use blocking mode",
      ucs_offsetof(ucc_tl_nccl_context_config_t, nccl_cfg_blocking),
      UCS_CONFIG_TYPE_BOOL},
 
--- a/src/components/tl/nccl/tl_nccl.c
+++ b/src/components/tl/nccl/tl_nccl.c
@@ -39,6 +39,11 @@ static ucs_config_field_t ucc_tl_nccl_context_config_table[] = {
      UCS_CONFIG_TYPE_ENUM(ucc_tl_nccl_completion_sync_names)
     },
 
+    {"NCCL_CFG_BLOCKING", "1",
+     "If set to 0 will use non-blocking mode, if set to 1 will use blocking",
+     ucs_offsetof(ucc_tl_nccl_context_config_t, nccl_cfg_blocking),
+     UCS_CONFIG_TYPE_BOOL},
+
     {NULL}};
 
 UCC_CLASS_DEFINE_NEW_FUNC(ucc_tl_nccl_lib_t, ucc_base_lib_t,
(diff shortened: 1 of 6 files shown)


shimmybalsam:
bot:retest

src/components/tl/nccl/tl_nccl_context.c
```diff
@@ -11,10 +11,37 @@
 #include "core/ucc_ee.h"
 #include "utils/arch/cpu.h"
 
+#if NCCL_USE_NON_BLOCKING
+static ucc_status_t ucc_tl_nccl_nb_progress(ucc_tl_nccl_task_t *task) {
```
samnordmann, line 15: I think it would be better to do something in the spirit of:
```
static ucc_status_t ucc_tl_nccl_progress(ucc_tl_nccl_task_t *task) {
#if NCCL_USE_NON_BLOCKING
   return ucc_tl_nccl_nb_progress(ucc_tl_nccl_task_t *task);
#else
    return UCC_OK;
#endif
}
```
This way you can avoid many branches, here on line 14 but also each time you call ucc_tl_nccl_nb_progress

shimmybalsam:
> Can we abstract it better than having ifdef's all around the code?

@manjugv I abstracted/removed most of the branching, let me know what you think, thanks.

manjugv (approved):

shimmybalsam:
bot:retest

src/components/tl/nccl/tl_nccl.h
```diff
@@ -122,17 +126,41 @@ typedef struct ucc_tl_nccl_task {
...
+        if (ncclInProgress == *nccl_status) {
+            *task_st = UCC_INPROGRESS;
+            return UCC_INPROGRESS;
+        }
+    }
+#endif
+    return UCC_OK;
+}
+
+#define NCCLCHECK_GOTO(_cmd, _label, _st, _lib, _task_st, _comm, _check_nb)    \
```
Sergei-Lebedev, line 149: i would rather do something like this

```
do {
    ncclResult_t e, st;
    
    /* check nccl function status */
    e = _cmd;  
    if (ucc_unlikely(e != ncclSuccess && e != ncclInProgress))
    {
        _st = UCC_ERR_NO_MESSAGE;
        goto label;
    }

    /* check nccl comm status */
    st = ncclCommGetAsyncError(nccl_commm, &e);
    if (ucc_unlikely(st != ncclSuccess)) {
        _st = UCC_ERR_NO_MESSAGE;
        goto label;
    }

    if (e == ncclInProgress) {
        _st = UCC_INPROGRESS;
    } else if (e == ncclSuccess) {
        _st = UCC_OK;
    } else {
        _st = UCC_ERR_NO_MESSAGE;
        goto label;
    }
} while(0);
```

src/components/tl/nccl/tl_nccl.c
```diff
@@ -39,6 +39,11 @@ static ucs_config_field_t ucc_tl_nccl_context_config_table[] = {
      UCS_CONFIG_TYPE_ENUM(ucc_tl_nccl_completion_sync_names)
     },
 
+    {"NCCL_CFG_BLOCKING", "1",
```
Sergei-Lebedev, line 42: just "BLOCKING" to match NCCL config

src/components/tl/nccl/tl_nccl.c
```diff
@@ -39,6 +39,11 @@ static ucs_config_field_t ucc_tl_nccl_context_config_table[] = {
...
+     "If set to 0 will use non-blocking mode, if set to 1 will use blocking",
```
Sergei-Lebedev, line 43: If set to 0 will use non-blocking mode communicator behavior, if set to 1 will use blocking mode

src/components/tl/nccl/tl_nccl.h
shimmybalsam: we can't check if e != ncclInProgress or call ncclCommGetAsyncError unless we are in nccl version that supports nb. This is why I put these checks in seperate function, which if doesn't support nb then will "do nothing" and then back in the macro check if e != ncclSuccess

Sergei-Lebedev (approved):

PR #801, opened 2023-06-27
MC/CUDA: add mc managed and tests

shimmybalsam:
## What
Adding support for memory type cudaManaged across UCC.

### DIFF
--- a/src/components/mc/cuda/mc_cuda.c
+++ b/src/components/mc/cuda/mc_cuda.c
@@ -116,7 +116,8 @@ static ucc_status_t ucc_mc_cuda_get_attr(ucc_mc_attr_t *mc_attr)
 }
 
 static ucc_status_t ucc_mc_cuda_mem_alloc(ucc_mc_buffer_header_t **h_ptr,
-                                          size_t                   size)
+                                          size_t                   size,
+                                          ucc_memory_type_t        mt)
 {
     cudaError_t             st;
     ucc_mc_buffer_header_t *h =
@@ -126,7 +127,9 @@ static ucc_status_t ucc_mc_cuda_mem_alloc(ucc_mc_buffer_header_t **h_ptr,
                  sizeof(ucc_mc_buffer_header_t));
         return UCC_ERR_NO_MEMORY;
     }
-    st = cudaMalloc(&h->addr, size);
+    st = (mt == UCC_MEMORY_TYPE_CUDA) ? cudaMalloc(&h->addr, size) :
+                                        cudaMallocManaged(&h->addr, size,
+                                                          cudaMemAttachGlobal);
     if (ucc_unlikely(st != cudaSuccess)) {
         cudaGetLastError();
         mc_error(&ucc_mc_cuda.super,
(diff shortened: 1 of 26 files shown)


shimmybalsam:
@edgargabriel please see small changes made in TL/ROCM

src/components/mc/cuda/mc_cuda.c
```diff
@@ -139,20 +142,23 @@ static ucc_status_t ucc_mc_cuda_mem_alloc(ucc_mc_buffer_header_t **h_ptr,
     h->from_pool = 0;
     h->mt        = UCC_MEMORY_TYPE_CUDA;
```
samnordmann, line 143: shouldn't we use `mt` here ?

src/components/mc/cuda/mc_cuda.c
```diff
@@ -139,20 +142,23 @@ static ucc_status_t ucc_mc_cuda_mem_alloc(ucc_mc_buffer_header_t **h_ptr,
...
 
 static ucc_status_t ucc_mc_cuda_mem_pool_alloc(ucc_mc_buffer_header_t **h_ptr,
-                                               size_t                   size)
+                                               size_t                   size,
+                                               ucc_memory_type_t        mt)
 {
     ucc_mc_buffer_header_t *h = NULL;
-    if (size <= MC_CUDA_CONFIG->mpool_elem_size) {
+    if (size <= MC_CUDA_CONFIG->mpool_elem_size &&
+        mt != UCC_MEMORY_TYPE_CUDA_MANAGED) {
```
samnordmann, line 156: If I understand correctly we choose not to allocate a mpool for cuda managed ?

src/core/ucc_coll.c
```diff
@@ -219,6 +219,9 @@ UCC_CORE_PROFILE_FUNC(ucc_status_t, ucc_collective_init,
         case UCC_MEMORY_TYPE_CUDA:
```
samnordmann, line 219: nit: just add the case cuda managed under the case cuda:
```
case UCC_MEMORY_TYPE_CUDA:
case UCC_MEMORY_TYPE_CUDA_MANAGED:
     coll_ee_type = UCC_EE_CUDA_STREAM;
     break;
```

test/gtest/core/test_mc_reduce.cc
```diff
@@ -57,6 +57,9 @@ class test_mc_reduce : public testing::Test {
             }
 #endif
             break;
+        case UCC_MEMORY_TYPE_CUDA_MANAGED:
```
samnordmann, line 60: why do we treat Cuda managed differently than Cuda here ? What with the triggered case ?
If there is not difference then you can simply add `case UCC_MEMORY_TYPE_CUDA_MANAGED:` below `case UCC_MEMORY_TYPE_CUDA:`

edgargabriel:
> @edgargabriel please see small changes made in TL/ROCM

@shimmybalsam thanks for the heads up. I tested your branch and things still work for ROCm with your change.

src/components/mc/cuda/mc_cuda.c
shimmybalsam: no, there are certain cases (free) for example that we check what h->mt is and based on that choose what function to call, and besides alloc we want all other mc functions to treat cudaManages as cuda.

src/components/mc/cuda/mc_cuda.c
shimmybalsam: correct

test/gtest/core/test_mc_reduce.cc
shimmybalsam: we don't support triggered with cudaManaged

samnordmann (approved):

Sergei-Lebedev (approved):

src/components/mc/cuda/mc_cuda.c
```diff
@@ -126,7 +127,9 @@ static ucc_status_t ucc_mc_cuda_mem_alloc(ucc_mc_buffer_header_t **h_ptr,
                  sizeof(ucc_mc_buffer_header_t));
         return UCC_ERR_NO_MEMORY;
     }
-    st = cudaMalloc(&h->addr, size);
+    st = (mt == UCC_MEMORY_TYPE_CUDA) ? cudaMalloc(&h->addr, size) :
+                                        cudaMallocManaged(&h->addr, size,
+                                                          cudaMemAttachGlobal);
```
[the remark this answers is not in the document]
shimmybalsam: Needed for compilation - It is supposed to be this way by default but I think it is compiler specific

PR #805, opened 2023-06-30
TEST: disable cuda_ipc in ucx

Sergei-Lebedev:
## What
Disable UCX cuda_ipc transport for some tests.

## Why ?
There is known issue when TL CUDA and cuda_ipc transport used together. They might try to register same memory twice and as a result either UCX or UCC aborts.
see for more details https://github.com/openucx/ucx/issues/3192


manjugv:
@Sergei-Lebedev We should update this (https://github.com/openucx/ucc#compile-ucx) accordingly. Is there a better place to capture ?

Sergei-Lebedev:
> @Sergei-Lebedev We should update this (https://github.com/openucx/ucc#compile-ucx) accordingly. Is there a better place to capture ?

we can mention it in FAQ, but I don't see why we should update compilation instructions

PR #810, opened 2023-07-18
TL/MLX5: complete the bcast mcast design

MamziB:
TL/MLX5: complete the bcast mcast design

@jladd-mlnx @janjust @manjugv

### DIFF
--- a/src/components/tl/mlx5/mcast/tl_mlx5_mcast_context.c
+++ b/src/components/tl/mlx5/mcast/tl_mlx5_mcast_context.c
@@ -6,13 +6,256 @@
 
 #include <inttypes.h>
 #include "tl_mlx5_mcast.h"
+#include "tl_mlx5_mcast_helper.h"
+#include "tl_mlx5_mcast_rcache.h"
 #include "utils/arch/cpu.h"
 #include <ucs/sys/string.h>
 #include "core/ucc_service_coll.h"
 #include "tl_mlx5.h"
 
-ucc_status_t ucc_tl_mlx5_mcast_context_init(ucc_tl_mlx5_mcast_context_t    *context, /* NOLINT */
-                                            ucc_tl_mlx5_mcast_ctx_params_t *mcast_ctx_conf /* NOLINT */)
+
+static ucc_status_t ucc_tl_mlx5_mcast_service_bcast(void *arg, void *buf, int size, int root)
 {
+    ucc_tl_mlx5_mcast_oob_p2p_context_t *ctx    = (ucc_tl_mlx5_mcast_oob_p2p_context_t *)arg;
+    ucc_status_t                         status = UCC_OK;
+    ucc_team_t                          *team   = ctx->base_team;
+    ucc_subset_t                         subset = ctx->subset;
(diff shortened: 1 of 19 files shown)


manjugv:
@MamziB CI error: 
```
mcast/tl_mlx5_mcast_progress.c:107:19: error: unused function 'psn2root' [-Werror,-Wunused-function]
[523](https://github.com/openucx/ucc/actions/runs/5590396224/job/15141427772?pr=810#step:9:524)
static inline int psn2root(ucc_tl_mlx5_mcast_coll_req_t *req, int psn)
```

MamziB:
@samnordmann can you please take a look at this PR when you have time?

src/components/tl/mlx5/mcast/tl_mlx5_mcast_context.c
```diff
@@ -6,13 +6,256 @@
...
+
+    if (0 != rdma_resolve_addr(ctx->id, (struct sockaddr *)&ip_oib_addr,
+                (struct sockaddr *) &dst_addr, 1000)) {
+        tl_error(lib, "TL/MCAST: Failed to resolve rdma addr, errno %d", errno);
+        status = UCC_ERR_NO_RESOURCE;
+        goto error;
+    }
+
+    if (rdma_get_cm_event(ctx->channel, &revent) < 0 ||
+        revent->event != RDMA_CM_EVENT_ADDR_RESOLVED) {
```
samnordmann, line 161: I might be wrong but in the case where `rdma_get_cm_event` is successfull and
`revent->event != RDMA_CM_EVENT_ADDR_RESOLVED` do we need to free `revent`?

samnordmann (commented):
Hey @MamziB, thanks for this PR! 
I started reviewing the context initialization and I'm posting here some general remarks. I'd be happy to discuss it with you.

1) Since the PR deals with a lot of different components, I think it could help the quality and pace of the review it was divided into several PRs or commits. On the other hand I'm aware it would require extra work for you.
What do you think about it ?

2) About the service bcast and barrier, If I'm not mistaken those functions are called during team creation, therefore I think that they have to be made non-blocking (because UCC ensures team creation is not blocking). Could you also explain why the collectives are performed on the `ctx->base_team` and not on the team?

3) I'm getting a compilation error with gcc 13.1.0. I can send you the full log if you need
```
src/components/tl/mlx5/mcast/tl_mlx5_mcast_progress.c:62:14: error: conflicting types for ‘ucc_tl_mlx5_mcast_check_nack_requests_all’ due to enum/integer mismatch; have ‘ucc_status_t(ucc_tl_mlx5_mcast_coll_comm_t *)’ {aka ‘ucc_status_t(struct ucc_tl_mlx5_mcast_coll_comm *)’} [-Werror=enum-int-mismatch]
```

Sergei-Lebedev:
@MamziB please split the PR into multiple smaller parts. You can extract control path such as QP and CQ as well as rcache etc.
It's hard to review it

MamziB:
@Sergei-Lebedev Sure will do. Let me close this one then.

PR #818, opened 2023-08-07
CORE: fix error handling for ctx create epilog

samnordmann:
## What
Fix some bugs in the core dealing with context create epilog error handling

### DIFF
--- a/src/core/ucc_context.c
+++ b/src/core/ucc_context.c
@@ -417,21 +417,6 @@ static ucc_status_t ucc_create_tl_contexts(ucc_context_t *ctx,
         status = UCC_ERR_NOT_FOUND;
         goto err;
     }
-    /* build the list of names of all available tl contexts.
-       This is a convenience data struct for CLs */
-    ctx->all_tls.count = ctx->n_tl_ctx;
-    ctx->all_tls.names = ucc_malloc(sizeof(char*) * ctx->n_tl_ctx, "all_tls");
-    if (!ctx->all_tls.names) {
-        ucc_error("failed to allocate %zd bytes for all_tls names",
-                  sizeof(char*) * ctx->n_tl_ctx);
-        status = UCC_ERR_NO_MEMORY;
-        goto err;
-
-    }
-    for (i = 0; i < ctx->n_tl_ctx; i++) {
-        ctx->all_tls.names[i] = (char*)ucc_derived_of(ctx->tl_ctx[i]->super.lib,
-                                               ucc_tl_lib_t)->iface->super.name;
-    }
     return UCC_OK;
 err:
     for (i = 0; i < ctx->n_tl_ctx; i++) {
@@ -602,7 +587,7 @@ ucc_status_t ucc_context_create_proc_info(ucc_lib_h                   lib,
     ucc_tl_lib_t              *tl_lib;
     ucc_context_t             *ctx;
     ucc_status_t               status;
-    uint64_t                   i, j;
+    uint64_t                   i, j, n_tl_ctx, created_ctx_counter = 0;
     int                        num_cls;
 
     num_cls = config->n_cl_cfg;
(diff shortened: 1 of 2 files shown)


shimmybalsam (approved):

src/core/ucc_context.c
```diff
@@ -793,20 +778,22 @@ ucc_status_t ucc_context_create_proc_info(ucc_lib_h                   lib,
...
+                created_ctx_counter++;
+            } else {
                 if (ucc_tl_is_required(lib, tl_lib->iface, 1)) {
                     ucc_error("ctx create epilog for %s failed: %s",
                               tl_lib->iface->super.name, ucc_status_string(status));
                     goto error_ctx_create;
                 } else {
                     ucc_debug("ctx create epilog for %s failed: %s",
                               tl_lib->iface->super.name, ucc_status_string(status));
-                    tl_lib->iface->context.destroy(&tl_ctx->super);
```
Sergei-Lebedev, line 809: why we don't destroy context in this case?

src/core/ucc_context.c
```diff
@@ -417,21 +417,6 @@ static ucc_status_t ucc_create_tl_contexts(ucc_context_t *ctx,
         status = UCC_ERR_NOT_FOUND;
         goto err;
     }
-    /* build the list of names of all available tl contexts.
-       This is a convenience data struct for CLs */
-    ctx->all_tls.count = ctx->n_tl_ctx;
-    ctx->all_tls.names = ucc_malloc(sizeof(char*) * ctx->n_tl_ctx, "all_tls");
```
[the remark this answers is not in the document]
samnordmann: --> I reversed this change
Actually `ctx->all_tls` is used for cl ctx creation that happens between tl ctx create and tl ctx create epilog. So this object needs to be filled at tl ctx create, as it is done currently.

src/core/ucc_context.c
```diff
@@ -816,17 +803,38 @@ ucc_status_t ucc_context_create_proc_info(ucc_lib_h                   lib,
...
 
     ucc_debug("created ucc context %p for lib %s", ctx, lib->full_prefix);
     *context = ctx;
     return UCC_OK;
 
 error_ctx_create:
-    for (i = 0; i < ctx->n_cl_ctx; i++) {
-        config->cl_cfgs[i]->cl_lib->iface->context.destroy(
-            &ctx->cl_ctx[i]->super);
+    for (j = 0; j < created_ctx_counter; j++) {
```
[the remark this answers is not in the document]
samnordmann: right, I got confused, done

src/core/ucc_context.c
samnordmann: In my opinion we shouldn't call ctx->destroy at this point because the ctx is not fully setup (because this happens before ctx create epilog).
The implementation of ctx create epilog itself should take care of the cleanup in case of an error that happens inside it.

samnordmann:
Additions / Issues:
- I fixed tl/sharp ctx epilog cleanup.
- if ctx create epilog fails, we need to call  `ucc_tl_context_t` (i.e. the base class) cleanup function. Do you have an idea how to properly do that? Should it be in core or in each tl's ctx create epilog function ?
- We should also remove context destroy in case of an error inside `ucc_create_tl_contexts`. But I don't how to handle an error in this case. We should call a function that cleans up only what has been allocated by ctx create (and not ctx create epilog) https://github.com/openucx/ucc/blob/6bdb758bc3d7467a29fc86f9cc80dc636c8ae82c/src/core/ucc_context.c#L439C16-L439C16

src/core/ucc_context.c
samnordmann: Actually we might want/need here and elsewhere (see [comment](https://github.com/openucx/ucc/pull/818#issuecomment-1669372142)) a function that cleans up only whats been allocated by ctx init (and not init epilog)...

Sergei-Lebedev: ok, please add short comment to context create epilog description that TL is responsible to cleanup resources if context create epilog fails

src/core/ucc_context.c
```diff
@@ -816,13 +819,26 @@ ucc_status_t ucc_context_create_proc_info(ucc_lib_h                   lib,
                                              tl_ctx);
                 }
             }
+        } else {
+            created_ctx_counter++;
         }
     }
+    if (0 == created_ctx_counter) {
+        ucc_error("no TL context created");
+        status = UCC_ERR_NO_MESSAGE;
```
Sergei-Lebedev, line 828: status = UCC_ERR_NO_RESOURCE

src/core/ucc_context.c
samnordmann: I put back the `destroy` at this line
see https://github.com/openucx/ucc/pull/818#issuecomment-1695843289

Sergei-Lebedev (approved):

manjugv (approved):

PR #820, opened 2023-08-15
TL/UCP: add reorder ranks to reduce_scatter

shimmybalsam:
## What
Add reordering ranks to reduce_scatter ring and reduce_scatterv ring 

## Why ?
Better performance

## How ?
Mapping ranks to first send messages within the same node
**Note:** in case of bidirectional, we need 2 use both mappings of reverse and reorder consecutively.

### DIFF
--- a/src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
+++ b/src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
@@ -12,6 +12,8 @@
 #include "utils/ucc_coll_utils.h"
 #include "utils/ucc_dt_reduce.h"
 
+#define REVERSED_FRAG 1
+
 static inline void send_completion_common(void *request, ucs_status_t status,
                                           void *user_data)
 {
@@ -112,7 +114,10 @@ static void ucc_tl_ucp_reduce_scatter_ring_progress(ucc_coll_task_t *coll_task)
 
     sendto   = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, sendto);
     recvfrom = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, recvfrom);
-
+    if (team->cfg.use_reordering) {
+        sendto   = ucc_ep_map_eval(task->subset.map, sendto);
+        recvfrom = ucc_ep_map_eval(task->subset.map, recvfrom);
+    }
     max_block_size = task->reduce_scatter_ring.max_block_count * dt_size;
     busy           = task->reduce_scatter_ring.s_scratch_busy;
     r_scratch      = task->reduce_scatter_ring.scratch;
@@ -129,8 +134,11 @@ static void ucc_tl_ucp_reduce_scatter_ring_progress(ucc_coll_task_t *coll_task)
         reduce_target = s_scratch[id];
         step          = task->tagged.send_posted;
         prevblock     = (rank - 1 - step + size) % size;
-        prevblock =
+        prevblock     =
             ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, prevblock);
+        if (team->cfg.use_reordering) {
+            prevblock     = ucc_ep_map_eval(task->subset.map, prevblock);
+        }
         /* reduction */
         ucc_assert(task->tagged.recv_posted == task->tagged.recv_completed);
         ucc_assert(task->tagged.recv_posted < size);
@@ -172,7 +180,9 @@ static void ucc_tl_ucp_reduce_scatter_ring_progress(ucc_coll_task_t *coll_task)
         recv_data_from = (rank - 2 - step + size) % size;
         recv_data_from =
             ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, recv_data_from);
-
+        if (team->cfg.use_reordering) {
+            recv_data_from = ucc_ep_map_eval(task->subset.map, recv_data_from);
+        }
         ucc_ring_frag_count(task, count, recv_data_from, &frag_count);
 
         UCPCHECK_GOTO(ucc_tl_ucp_recv_nb(r_scratch, frag_count * dt_size, mem_type,
@@ -199,17 +209,18 @@ ucc_tl_ucp_reduce_scatter_ring_start(ucc_coll_task_t *coll_task)
     ucc_tl_ucp_team_t *team     = TASK_TEAM(task);
     ucc_rank_t         size     = task->subset.map.ep_num;
     ucc_rank_t         rank     = task->subset.myrank;
-    ucc_rank_t         sendto   = (rank + 1) % size;
-    ucc_rank_t         recvfrom = (rank - 1 + size) % size;
     size_t             count    = args->dst.info.count * size;
     ucc_datatype_t     dt       = args->dst.info.datatype;
     size_t             dt_size  = ucc_dt_size(dt);
     ucc_memory_type_t  mem_type = args->dst.info.mem_type;
     void *             sbuf     = args->src.info.buffer;
     int                step     = 0;
+    ucc_rank_t         sendto   = (rank + 1) % size;
+    ucc_rank_t         recvfrom = (rank - 1 + size) % size;
+    ucc_rank_t         recv_block = (rank - 2 - step + size) % size;
+    ucc_rank_t         send_block = (rank - 1 - step + size) % size;
     size_t             block_offset, frag_count, frag_offset;
     void              *r_scratch;
-    ucc_rank_t         send_block, recv_block;
     ucc_status_t       status;
 
     ucc_tl_ucp_task_reset(task, UCC_INPROGRESS);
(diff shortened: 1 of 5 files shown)


shimmybalsam:
image

shimmybalsam:
image

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -112,7 +114,10 @@ static void ucc_tl_ucp_reduce_scatter_ring_progress(ucc_coll_task_t *coll_task)
 
     sendto   = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, sendto);
     recvfrom = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, recvfrom);
-
+    if (team->cfg.use_reordering) {
```
samnordmann, line 117: should we put if/else branching here ?

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -129,8 +134,11 @@ static void ucc_tl_ucp_reduce_scatter_ring_progress(ucc_coll_task_t *coll_task)
         reduce_target = s_scratch[id];
         step          = task->tagged.send_posted;
         prevblock     = (rank - 1 - step + size) % size;
-        prevblock =
+        prevblock     =
             ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, prevblock);
+        if (team->cfg.use_reordering) {
```
samnordmann, line 139: we could use the inline form here
```
prevblock = ucc_ep_map_eval(team->cfg.use_reordering ?  task->reduce_scatter_ring.inv_map : task->subset.map, prevblock) 
```

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -223,13 +234,17 @@ ucc_tl_ucp_reduce_scatter_ring_start(ucc_coll_task_t *coll_task)
...
 
+    r_scratch  = task->reduce_scatter_ring.scratch;
     sendto     = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, sendto);
     recvfrom   = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, recvfrom);
-    r_scratch  = task->reduce_scatter_ring.scratch;
-    recv_block = (rank - 2 - step + size) % size;
-    send_block = (rank - 1 - step + size) % size;
     recv_block = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, recv_block);
     send_block = ucc_ep_map_eval(task->reduce_scatter_ring.inv_map, send_block);
+    if (team->cfg.use_reordering) {
```
samnordmann, line 242: Here and elswhere we could simply define a map
```
ucc_ep_map_t map = team->cfg.use_reordering? task->subset.map : task->reduce_scatter_ring.inv_map)
```
and simply use this map instead of branching
What do you think ?

src/utils/ucc_coll_utils.c
```diff
@@ -631,20 +631,22 @@ ucc_ep_map_t ucc_ep_map_create_reverse(ucc_rank_t size)
     return map;
 }
 
-static inline int ucc_ep_map_is_reverse(ucc_ep_map_t *map)
+static inline int ucc_ep_map_is_reverse(ucc_ep_map_t *map,
+                                        int reversed_reordered_flag)
```
samnordmann, line 635: Would it be clearer to name this variable `use_reordering` ?
I'm honestly not sure, it's just that the name `reversed_reordered_flag` is not so clear to me but I let you decide what you think's best

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -12,6 +12,8 @@
 #include "utils/ucc_coll_utils.h"
 #include "utils/ucc_dt_reduce.h"
 
+#define REVERSED_FRAG 1
```
samnordmann, line 15: Can you explain what is the point of such a variable ? Do we want to make this flag configurable ?

test/gtest/utils/test_ep_map.cc
```diff
@@ -131,7 +131,7 @@ class test_ep_map_inv : public test_ep_map {
     void check_inv(EpMap map)
     {
         ucc_ep_map_t inv;
-        EXPECT_EQ(UCC_OK, ucc_ep_map_create_inverse(map.map, &inv));
+        EXPECT_EQ(UCC_OK, ucc_ep_map_create_inverse(map.map, &inv, 0));
```
samnordmann, line 134: what about adding test for the case where reordering is used ?

samnordmann (commented):
Hi @shimmybalsam , Thanks for the PR. It looks good to me, I only left minor suggestions

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
shimmybalsam: no its not configurable I added this more for readability. For example when we check if (frag == 1) this ight be confusing and it is better understood what this if statement is checking when using if (frag == REVERSED_FRAG).

This is my opinion, but if you still think this should be removed let me know,.

src/utils/ucc_coll_utils.c
shimmybalsam: This flag should be true only for the case of both reordering and reverse (reverse meaning the second frag when using bidirectional), so not for just reordering.

test/gtest/utils/test_ep_map.cc
shimmybalsam: reordering is used by default on the collective level, so gtests running ring alg are testing it. 
In this test we are just checking that an inverse map works, adding reordering flag is not really relevant here in my opinion as it won't test any additional functionality of the map.

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
shimmybalsam: what do you mean? what should be in else?

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
shimmybalsam: no, it is not one or the other, if reordering is being used then we want both mappings to happen one after the other.

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
shimmybalsam: no, in the case of bidirectional and reordering both maps are needed to be used one after the other.

samnordmann: ok, got it

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
samnordmann: Thanks for the explanation
In my opinion, it doesn't make it clearer because it introduces a new variable, and `if (frag) {...}` is just enough
But it's not critical, so please do what you think is best

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
samnordmann: nevermind

samnordmann (approved):

src/components/tl/ucp/reduce_scatterv/reduce_scatterv_ring.c
```diff
@@ -204,37 +213,42 @@ ucc_tl_ucp_reduce_scatterv_ring_start(ucc_coll_task_t *coll_task)
...
     size_t             dt_size  = ucc_dt_size(dt);
     ucc_memory_type_t  mem_type = args->dst.info_v.mem_type;
     void *             sbuf     = args->src.info.buffer;
     int                step     = 0;
+    ucc_rank_t         sendto   = (rank + 1) % size;
+    ucc_rank_t         recvfrom = (rank - 1 + size) % size;
+    ucc_rank_t         recv_block = (rank - 2 - step + size) % size;
+    ucc_rank_t         send_block = (rank - 1 - step + size) % size;
     size_t             block_offset, frag_count, frag_offset;
     void *             r_scratch;
```
Sergei-Lebedev, line 226:
```suggestion
    void              *r_scratch;
```

src/components/tl/ucp/reduce_scatterv/reduce_scatterv_ring.c
```diff
@@ -341,6 +361,7 @@ ucc_tl_ucp_reduce_scatterv_ring_init(ucc_base_coll_args_t *coll_args,
     ucc_tl_ucp_schedule_t *tl_schedule;
     ucc_schedule_t *       schedule;
     ucc_coll_task_t *      ctask;
+    ucc_sbgp_t *           sbgp;
```
Sergei-Lebedev, line 364: alignment

src/utils/ucc_coll_utils.h
```diff
@@ -230,7 +230,8 @@ void ucc_coll_str(const ucc_coll_task_t *task, char *str, size_t len,
 ucc_ep_map_t ucc_ep_map_create_reverse(ucc_rank_t size);
 
 /* Creates an inverse mapping for a given map */
-ucc_status_t ucc_ep_map_create_inverse(ucc_ep_map_t map, ucc_ep_map_t *inv_map);
+ucc_status_t ucc_ep_map_create_inverse(ucc_ep_map_t map, ucc_ep_map_t *inv_map,
```
Sergei-Lebedev, line 233: please add description for the parameters

src/components/tl/ucp/reduce_scatter/reduce_scatter_ring.c
```diff
@@ -353,13 +374,21 @@ ucc_tl_ucp_reduce_scatter_ring_init(ucc_base_coll_args_t *coll_args,
...
-    s[1].myrank = ucc_ep_map_eval(s[1].map, UCC_TL_TEAM_RANK(tl_team));
+        s[1].map    = ucc_ep_map_create_reverse(UCC_TL_TEAM_SIZE(tl_team));
+        s[1].myrank = ucc_ep_map_eval(s[1].map, s[0].myrank);
+    } else {
+        s[0].myrank     = UCC_TL_TEAM_RANK(tl_team);
+        s[0].map.type   = UCC_EP_MAP_FULL;
+        s[0].map.ep_num = UCC_TL_TEAM_SIZE(tl_team);
 
+        s[1].map    = ucc_ep_map_create_reverse(UCC_TL_TEAM_SIZE(tl_team));
+        s[1].myrank = ucc_ep_map_eval(s[1].map, UCC_TL_TEAM_RANK(tl_team));
```
Sergei-Lebedev, line 390: move out of if statement

Sergei-Lebedev (approved):

PR #821, opened 2023-08-15
TEST: fix summary calculation in test_mpi

Sergei-Lebedev:
## What
Fixes summary calc for ucc_test_mpi

## Why ?
Different ranks might execute different number of tests, e.g. if odd_even team split is used and initial world size is odd number.
internal issue https://redmine.mellanox.com/issues/3477397

### DIFF
--- a/test/mpi/main.cc
+++ b/test/mpi/main.cc
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * Copyright (c) Advanced Micro Devices, Inc. 2023. ALL RIGHTS RESERVED.
  *
  * See file LICENSE for terms.
@@ -7,8 +7,8 @@
 
 #include <getopt.h>
 #include <sstream>
-#include "test_mpi.h"
 #include <chrono>
+#include "test_mpi.h"
 
 int test_rand_seed = -1;
 static size_t test_max_size = TEST_UCC_RANK_BUF_SIZE_MAX;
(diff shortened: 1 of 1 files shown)


test/mpi/main.cc
```diff
@@ -505,14 +518,15 @@ void ProcessArgs(int argc, char** argv)
...
-    int failed = 0;
+    int failed                       = 0;
+    int total_done_skipped_failed[4] = {0};
+    std::chrono::steady_clock::time_point begin;
     int size, required, provided, completed, rank;
     UccTestMpi *test;
     MPI_Request req;
     std::string err;
 
+    begin = std::chrono::steady_clock::now();
```
samnordmann, line 529: why do we not assign it when it is declared ?

test/mpi/main.cc
```diff
@@ -596,50 +610,51 @@ int main(int argc, char *argv[])
...
+        case UCC_ERR_LAST:
+            total_done_skipped_failed[2]++;
+            break;
+        default:
+            total_done_skipped_failed[3]++;
+        }
+    }
+    MPI_Iallreduce(MPI_IN_PLACE, total_done_skipped_failed,
+                   sizeof(total_done_skipped_failed)/sizeof(int),
+                   MPI_INT, MPI_MAX, MPI_COMM_WORLD, &req);
```
samnordmann, line 631: why not just a "reduce" rooted in rank=0?

test/mpi/main.cc
```diff
@@ -596,50 +610,51 @@ int main(int argc, char *argv[])
...
                   << "s" << std::endl;
 
-	if (skipped == test->results.size()) {
-            std::cout << "\n All tests have been skipped, indicating most likely "
-                         "a problem\n";
-            failed = 1;
-	}
+        /* check if all tests have been skipped */
+        if (total_done_skipped_failed[0] == total_done_skipped_failed[2]) {
+                std::cout <<
```
samnordmann, line 651: indentation

test/mpi/main.cc
Sergei-Lebedev: just because alignment looks bad if move it to declaration block

test/mpi/main.cc
Sergei-Lebedev: essentially will have the same effect, it was allreduce here before so I keep it as allreduce. Can replace with reduce if you think it's important

samnordmann (approved):

manjugv (approved):

PR #822, opened 2023-08-16
The Neighbor Exchange Algorithm for Allgather

ikryukov:
## What
Implementation of Neighbor Exchange Allgather algorithm described in this paper: https://ieeexplore.ieee.org/document/1592302
Based on original implementation from OpenMPI: https://github.com/open-mpi/ompi/blob/8514e717215a42ddb6db065c3baae3a2779a072a/ompi/mca/coll/base/coll_base_allgather.c#L624

## Why ?
According to mentioned paper: ` For the TCP/IP protocol, an algorithm that incorporates pair-wise exchange is the best for minimizing TCP/IP traffic ` and this algorithm require **N / 2** steps instead of **N - 1** like in ring.

### DIFF
--- a/src/components/tl/ucp/allgather/allgather.c
+++ b/src/components/tl/ucp/allgather/allgather.c
@@ -17,6 +17,10 @@ ucc_base_coll_alg_info_t
             {.id   = UCC_TL_UCP_ALLGATHER_ALG_RING,
              .name = "ring",
              .desc = "O(N) Ring"},
+        [UCC_TL_UCP_ALLGATHER_ALG_NEIGHBOR] =
+            {.id   = UCC_TL_UCP_ALLGATHER_ALG_NEIGHBOR,
+             .name = "neighbor",
+             .desc = "O(N) Neighbor Exchange N/2 steps"},
         [UCC_TL_UCP_ALLGATHER_ALG_LAST] = {
             .id = 0, .name = NULL, .desc = NULL}};
 
--- a/src/components/tl/ucp/allgather/allgather.c
+++ b/src/components/tl/ucp/allgather/allgather.c
@@ -7,6 +7,8 @@
 #include "tl_ucp.h"
 #include "allgather.h"
 
+#define ALLGATHER_MAX_PATTERN_SIZE (sizeof(UCC_TL_UCP_ALLGATHER_DEFAULT_ALG_SELECT_STR))
+
 ucc_base_coll_alg_info_t
     ucc_tl_ucp_allgather_algs[UCC_TL_UCP_ALLGATHER_ALG_LAST + 1] = {
         [UCC_TL_UCP_ALLGATHER_ALG_KNOMIAL] =
@@ -27,4 +29,17 @@ ucc_base_coll_alg_info_t
 ucc_status_t ucc_tl_ucp_allgather_init(ucc_tl_ucp_task_t *task)
 {
     return ucc_tl_ucp_allgather_ring_init_common(task);
+}
+
+char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
+{
+    int   max_size = ALLGATHER_MAX_PATTERN_SIZE;
+    int   algo_num = UCC_TL_TEAM_SIZE(team) % 2
+                         ? UCC_TL_UCP_ALLGATHER_ALG_RING
+                         : UCC_TL_UCP_ALLGATHER_ALG_NEIGHBOR;
+    char *str      = ucc_malloc(max_size * sizeof(char));
+
+    ucc_snprintf_safe(str, max_size,
+                      UCC_TL_UCP_ALLGATHER_DEFAULT_ALG_SELECT_STR, algo_num);
+    return str;
 }
--- a/src/components/tl/ucp/allgather/allgather_neighbor.c
+++ b/src/components/tl/ucp/allgather/allgather_neighbor.c
@@ -66,12 +66,19 @@ ucc_status_t ucc_tl_ucp_allgather_neighbor_init(ucc_base_coll_args_t *coll_args,
 
 ucc_status_t ucc_tl_ucp_allgather_neighbor_init_common(ucc_tl_ucp_task_t *task)
 {
-    ucc_tl_ucp_team_t *team      = TASK_TEAM(task);
+    ucc_tl_ucp_team_t *team       = TASK_TEAM(task);
+    ucc_rank_t         tsize      = (ucc_rank_t)task->subset.map.ep_num;
     ucc_sbgp_t *sbgp;
 
     if (!ucc_coll_args_is_predefined_dt(&TASK_ARGS(task), UCC_RANK_INVALID)) {
         tl_error(UCC_TASK_LIB(task), "user defined datatype is not supported");
         return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    if (tsize % 2)
+    {
+        tl_warn(UCC_TASK_LIB(task), "odd team size is not supported, switching to ring");
+        return ucc_tl_ucp_allgather_ring_init_common(task);
     }
     
     if (!(task->flags & UCC_TL_UCP_TASK_FLAG_SUBSET)) {
--- a/src/components/tl/ucp/allgather/allgather_neighbor.c
+++ b/src/components/tl/ucp/allgather/allgather_neighbor.c
@@ -9,25 +9,24 @@
 
 static ucc_rank_t get_recv_from_rank(ucc_rank_t rank, ucc_rank_t size, int i)
 {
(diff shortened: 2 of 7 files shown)


src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,190 @@
...
+        neighbor[0]       = (rank - 1 + size) % size;
+        neighbor[1]       = (rank + 1) % size;
+        recv_data_from[0] = neighbor[0];
+        recv_data_from[1] = neighbor[0];
+        offset_at_step[0] = (-2);
+        offset_at_step[1] = (+2);
+    }
+    const int i_parity = i % 2;
+    return (recv_data_from[i_parity] +
+            offset_at_step[i_parity] * ((i + 1) / 2) + size) %
```
Sergei-Lebedev, line 32: ucc_div_round_up(i, 2) instead of ((i + 1) / 2)

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,190 @@
...
+        if (ucc_unlikely(UCC_OK != status)) {
+            return status;
+        }
+    }
+
+    if (!(trank % 2)) {
+        neighbor = (trank + 1) % tsize;
+    } else {
+        neighbor = (trank - 1 + tsize) % tsize;
+    }
```
Sergei-Lebedev, line 176: minor
```suggestion
    if (trank % 2) {
        neighbor = (trank - 1 + tsize) % tsize;
    } else {
        neighbor = (trank + 1) % tsize;
    }
```

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,190 @@
...
+
+ucc_status_t ucc_tl_ucp_allgather_neighbor_init(ucc_base_coll_args_t *coll_args,
+                                                ucc_base_team_t      *team,
+                                                ucc_coll_task_t     **task_h)
+{
+    ucc_tl_ucp_task_t *task;
+    ucc_status_t       status;
+
+    task   = ucc_tl_ucp_init_task(coll_args, team);
+    status = ucc_tl_ucp_allgather_neighbor_init_common(task);
```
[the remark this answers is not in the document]
ikryukov: merged

Sergei-Lebedev:
bot:retest

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,176 @@
...
+    ucc_tl_ucp_task_t *task      = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_tl_ucp_team_t *team      = TASK_TEAM(task);
+    ucc_rank_t         trank     = UCC_TL_TEAM_RANK(team);
+    ucc_rank_t         tsize     = UCC_TL_TEAM_SIZE(team);
+    void              *rbuf      = TASK_ARGS(task).dst.info.buffer;
+    ucc_memory_type_t  rmem      = TASK_ARGS(task).dst.info.mem_type;
+    ucc_datatype_t     dt        = TASK_ARGS(task).dst.info.datatype;
+    size_t             count     = TASK_ARGS(task).dst.info.count;
+    size_t             data_size = (count / tsize) * ucc_dt_size(dt);
+    ucc_rank_t         neighbor[2], i;
```
shimmybalsam, line 82: perhaps call the array neighbors instead of neighbor?

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,176 @@
+#include "config.h"
+#include "tl_ucp.h"
+#include "allgather.h"
+#include "core/ucc_progress_queue.h"
+#include "tl_ucp_sendrecv.h"
+#include "utils/ucc_math.h"
+#include "utils/ucc_coll_utils.h"
+#include "components/mc/ucc_mc.h"
+
+static ucc_rank_t get_recv_from_rank(ucc_rank_t rank, ucc_rank_t size, int i)
```
shimmybalsam, line 10: add pointer to neighbors array as a param of function so not to recalculate

shimmybalsam (approved):

shimmybalsam:
@Sergei-Lebedev should we add here ranks reordering? I assume it would improve performance substantially since each neighbor with default mapping will always cross cores...

Sergei-Lebedev:
> @Sergei-Lebedev should we add here ranks reordering? I assume it would improve performance substantially since each neighbor with default mapping will always cross cores...

this algorithm won't work with reordering, if need reordering we should fallback to ring

src/components/tl/ucp/allgather/allgather.c
```diff
@@ -17,10 +19,27 @@ ucc_base_coll_alg_info_t
...
 
 ucc_status_t ucc_tl_ucp_allgather_init(ucc_tl_ucp_task_t *task)
 {
     return ucc_tl_ucp_allgather_ring_init_common(task);
 }
+
+char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
+{
+    int   max_size = ALLGATHER_MAX_PATTERN_SIZE;
+    int   algo_num = UCC_TL_TEAM_SIZE(team) % 2
```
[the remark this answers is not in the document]
Sergei-Lebedev: we changing default to neighbor exchange here for even size communicators

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,176 @@
...
+    if (!ucc_coll_args_is_predefined_dt(&TASK_ARGS(task), UCC_RANK_INVALID)) {
+        tl_error(UCC_TASK_LIB(task), "user defined datatype is not supported");
+        status = UCC_ERR_NOT_SUPPORTED;
+        goto out;
+    }
+
+    if (UCC_TL_TEAM_SIZE(ucp_team) % 2) {
+        tl_warn(UCC_TASK_LIB(task),
+                "odd team size is not supported, switching to ring");
+        status = ucc_tl_ucp_allgather_ring_init_common(task);
```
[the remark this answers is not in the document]
Sergei-Lebedev: core will be able to fallback to other TL only, fallbacks within TL managed by TL itself

src/components/tl/ucp/allgather/allgather.c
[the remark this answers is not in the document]
samnordmann: yes
I just wanted to make sure that we indeed want to change the default in this case

src/components/tl/ucp/allgather/allgather_neighbor.c
ikryukov: ok, I'm not native speaker :+1:

src/components/tl/ucp/allgather/allgather_neighbor.c
ikryukov: just simplified it

samnordmann: what about defining an auxiliary function like that:

```
static inline ucc_rank_t offset_at_step(int i, ucc_rank_t rank) {
   return 2 * ((i % 2) ^^ (rank % 2) ? 1 : -1);
}
```

ikryukov: nice, but this is task for compiler to optimize this expression in such way, I would prefer to leave as is to be more human readable. (don't think it is hotspot anyway)

samnordmann (commented):
Thanks @ikryukov for the revisions! It looks good to me. Can you just fix the last comment please?

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -0,0 +1,176 @@
...
+    } else {
+        task->super.post     = ucc_tl_ucp_allgather_neighbor_start;
+        task->super.progress = ucc_tl_ucp_allgather_neighbor_progress;
+    }
+
+out:
+    if (status != UCC_OK) {
+        ucc_tl_ucp_put_task(task);
```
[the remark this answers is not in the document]
ikryukov: Sure, I can fix that, but I did it like in ring algo: https://github.com/openucx/ucc/blob/5548f08db523a9d5625ecee073b0a3e576f22cdb/src/components/tl/ucp/allgather/allgather_ring.c#L145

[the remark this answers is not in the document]
samnordmann: patch:
https://github.com/openucx/ucc/pull/837

samnordmann (approved):

Sergei-Lebedev (approved):

PR #824, opened 2023-08-17
TL/MLX5: [HW Multicast] adding --with-rdmacm configure option

MamziB:
TL/MLX5: [HW Multicast] adding --with-rdmacm configure option


janjust:
@Sergei-Lebedev Here we go - we're going to try and bbreak this down into a small of units as possible. The next set of PRs are critical for @MamziB upcoming Allgather feature.

Sergei-Lebedev (approved):

manjugv (approved):

samnordmann (approved):

PR #830, opened 2023-08-29
TL/UCP: add reorder ranks in allreduce kn

shimmybalsam:
## What
Add reorder ranks in allreduce kn

## How
ranks/peers already using basic 1:1 mapping in default task->subset.map, so no change needed in start/progress, only setting task->subset.map to correct reordering map in init is needed.

### DIFF
--- a/src/components/tl/ucp/allreduce/allreduce_knomial.c
+++ b/src/components/tl/ucp/allreduce/allreduce_knomial.c
@@ -218,8 +218,9 @@ ucc_status_t ucc_tl_ucp_allreduce_knomial_init_common(ucc_tl_ucp_task_t *task)
     size_t             count     = TASK_ARGS(task).dst.info.count;
     ucc_datatype_t     dt        = TASK_ARGS(task).dst.info.datatype;
     size_t             data_size = count * ucc_dt_size(dt);
-    ucc_rank_t         size      = (ucc_rank_t)task->subset.map.ep_num;
     ucc_mrange_uint_t *p         = &team->cfg.allreduce_kn_radix;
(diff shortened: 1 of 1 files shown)


shimmybalsam:
image

samnordmann (approved):

Sergei-Lebedev:
@shimmybalsam can you please run on a larger scale and up to 128MB

shimmybalsam:
> @shimmybalsam can you please run on a larger scale and up to 128MB

Sure but just out of general interest - why? We only use allreduce_knomial for small msgsize.

Sergei-Lebedev:
> > @shimmybalsam can you please run on a larger scale and up to 128MB
> 
> Sure but just out of general interest - why? We only use allreduce_knomial for small msgsize.

it's not clear from the graph that we got any improvements

shimmybalsam:
> > > @shimmybalsam can you please run on a larger scale and up to 128MB
> > 
> > 
> > Sure but just out of general interest - why? We only use allreduce_knomial for small msgsize.
> 
> it's not clear from the graph that we got any improvements

@Sergei-Lebedev I ran up to 4MB and there seems to be no improvement:

image


I also checked on 4 and 8 nodes and also on rock (2 4 and 8 nodes) , 1 ppn and full ppn (28 on jazz , 64 on rock) - still no imporvemnt.

But, when running non full ppn (but thats larger then 1) there is a positive effect.
For example here are the results for 20 ppn on 4 and 8 nodes on jazz and 34 ppn on 4 and 8 nodes on rock:

image

shimmybalsam:
bot:retest

Sergei-Lebedev (approved):

PR #834, opened 2023-08-31
TL/UCP: reorder ranks for SRA

shimmybalsam:
## What
Add reorder ranks in for Allreduce SRA

## Why ?
Performance benfit, especialy when n_extra ranks > 0, in which case we do a complete allreduce between "proxy" and "extra" ranks and if they are on the same node this significantly improves perf.

## How ?
By reordring ranks in each step:  reduce_scatter_knomial, allgather_knomial.

**Note:** allgather_knomial is used also for small message allgather and large message bcast (bcast SAG). In these cases reorder ranks will not be used, it is currently only implemented for the usage within allreduce SRA.

### DIFF
--- a/src/components/tl/ucp/allgather/allgather_knomial.c
+++ b/src/components/tl/ucp/allgather/allgather_knomial.c
@@ -45,10 +45,10 @@ void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
     size_t                 count     = args->dst.info.count;
     size_t                 dt_size   = ucc_dt_size(args->dst.info.datatype);
     size_t                 data_size = count * dt_size;
-    ucc_rank_t             size      = UCC_TL_TEAM_SIZE(team);
+    ucc_rank_t             size      = task->subset.map.ep_num;
     ucc_rank_t             broot     = args->coll_type == UCC_COLL_TYPE_BCAST ?
                                        args->root : 0;
-    ucc_rank_t             rank      = VRANK(UCC_TL_TEAM_RANK(team), broot, size);
+    ucc_rank_t             rank      = VRANK(task->subset.myrank, broot, size);
     size_t                 local     = GET_LOCAL_COUNT(args, size, rank);
     void                  *sbuf;
     ptrdiff_t              peer_seg_offset, local_seg_offset;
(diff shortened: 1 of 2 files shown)


shimmybalsam:
**Without extra ranks:**

image


**With extra ranks:**

image

shimmybalsam:
bot:retest

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -230,14 +238,22 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
     ucc_base_coll_args_t *coll_args, ucc_base_team_t *team,
     ucc_coll_task_t **task_h, ucc_kn_radix_t radix)
 {
+    ucc_tl_ucp_team_t *tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
     ucc_tl_ucp_task_t *task;
+    ucc_sbgp_t        *sbgp;
 
     task = ucc_tl_ucp_init_task(coll_args, team);
+    if (tl_team->cfg.use_reordering &&
+        coll_args->args.coll_type == UCC_COLL_TYPE_ALLREDUCE) {
```
samnordmann, line 247: Can you explain why we only enable it for ALLREDUCE?

samnordmann (approved):

src/components/tl/ucp/allgather/allgather_knomial.c
shimmybalsam: allgather_knomial is used also for small message allgather and large message bcast (bcast SAG). In these cases reorder ranks will not be used yet - additional changes will need to be made.
Currently in this PR, the only usage of reorder ranks in allgather that is implemented is as part of allreduce SRA.

Sergei-Lebedev (commented):
can we instead make map part of knomial pattern. That way we don't need to repeat ucc_ep_map_eval every time we get peer.
In future we can also add ring patter to do the same for ring algorithms

shimmybalsam:
> can we instead make map part of knomial pattern. That way we don't need to repeat ucc_ep_map_eval every time we get peer. In future we can also add ring patter to do the same for ring algorithms

As discussed on call, in my opinion we should merge this PR as is as first step, since it substantially improves performance of SRA. In future we can rewrite knomial pattern functions (which will effect all knomial algorithms in TL/UCP).

Meanwhile here is some more supportive data:

image

shimmybalsam:
bot:retest

Sergei-Lebedev (approved):

PR #836, opened 2023-08-31
TEST: fix one-sided tests always running

wfaderhold21:
## What
Currently, when running mpi test, onesided tests will be performed regardless of the user specifically setting a non-onesided test to perform. For example,

`mpirun -np 2 ucc_test_mpi -c allreduce`

Will perform both allreduce tests and onesided alltoall tests, which is not correct.

### DIFF
--- a/test/mpi/main.cc
+++ b/test/mpi/main.cc
@@ -606,6 +606,9 @@ int main(int argc, char *argv[])
         std::vector<ucc_coll_type_t>           os_colls(onesided_colls.size());
         std::vector<ucc_coll_type_t>::iterator it_start;
 
+        std::sort(colls.begin(), colls.end());
+        std::sort(onesided_colls.begin(), onesided_colls.end());
+
         it_start = std::set_intersection(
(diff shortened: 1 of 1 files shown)


Sergei-Lebedev:
ok to test

test/mpi/main.cc
```diff
@@ -602,7 +603,14 @@ int main(int argc, char *argv[])
     }
 
     if (has_onesided) {
-        test->set_colls(onesided_colls);
+        std::vector<ucc_coll_type_t>           os_colls(onesided_colls.size());
+        std::vector<ucc_coll_type_t>::iterator it_start;
+
+        it_start = std::set_intersection(
```
Sergei-Lebedev, line 609: set_intersection requires both ranges to be sorted, I'm not sure this is always the case

wfaderhold21: You're right. I've added in sorting of the list of collectives to test.

Sergei-Lebedev (approved):

samnordmann (approved):

PR #804, opened 2023-06-29
EC/ROCM and TL/RCCL: fix clang rocm5.6 errors

edgargabriel:
## What

this commit fixes two issues exposed by the UCC Linter-ROCM script. 

First, there was a mistake on the configure test checking for the rocm version. 
Second, the hip_fp16.h  header file in rocm-5.6 generates some warnings with clang-12 that are not ignored with the -Werror flag being set. Silence that particular warning for clang.

PR #844, opened 2023-09-21
EC/CPU: fix error message

Sergei-Lebedev:
## What
Fix typo in error message

### DIFF
--- a/src/components/ec/cpu/ec_cpu_reduce.c
+++ b/src/components/ec/cpu/ec_cpu_reduce.c
@@ -114,7 +114,7 @@
             break;                                                             \
         default:                                                               \
             ec_error(&ucc_ec_cpu.super,                                        \
-                     "float dtype does not support "                           \
+                     "int dtype does not support "                             \
                      "requested reduce op: %s",                                \
                      ucc_reduction_op_str(_op));                               \
             return UCC_ERR_NOT_SUPPORTED;                                      \


samnordmann (approved):

edgargabriel:
@Sergei-Lebedev it seems to me that we are running into the same issue with UCX that we had with UCC and was fixed in PR https://github.com/openucx/ucc/pull/804 . The hip header file generates a warning when compiled with clang, which in itself is harmless, but because  all warnings are treated as errors, the compilation fails. I will see what we can do about it. One option would be to compile UCX with gcc instead of clang (since we are not truly interested in running clang-tidy on ucx in this CI, but just on UCC)

edgargabriel:
@Sergei-Lebedev it is more complicated: compiling UCX with gcc instead of clang fixes the first issue. There is however another issue in that we need clang-14 to compile UCC  for the rocm/rccl components. With clang-12 we ran into another issue in the hip header files where a flag to disable some warnings is not recognized (in the hip header files) and generates a warning again:

```
In file included from /opt/rocm/include/hip/hip_runtime_api.h:483:
/opt/rocm/include/hip/surface_types.h:30:34: error: unknown warning group '-Wreserved-identifier', ignored [-Werror,-Wunknown-warning-option]
#pragma clang diagnostic ignored "-Wreserved-identifier"
```

With clang-14 this error is not observed, and UCC compiles correctly. However clang-tidy with clang-14 fails, see 

https://github.com/openucx/ucc/actions/runs/6264142779/job/17010173739?pr=846

Not sure what our options are. If you want I can convert Clang-tidy-rocm into a rocm compilation only (without the clang-tidy part), which would also still have some value to ensure that a pr is compiled also against the rocm components. Let me know what you suggest.

edgargabriel:
the problem is of course triggered by the release of rocm-5.7 a few days ago. The script fetches the most recent rocm version. I can also explore whether I could fetch a fixed rocm version instead, hence go back to rocm-5.6 for example.

edgargabriel:
https://github.com/openucx/ucc/pull/846 fixes the ROCM-linter compilation issue, by using rocm 5.6.1 (instead of the newest version), and compiling ucx with gcc instead of clang.

manjugv (approved):

PR #846, opened 2023-09-21
TEST: use gcc to compile UCX in clang-tidy-rocm

edgargabriel:
## What
The hip header file generates a warning when compiled with clang, which in itself is harmless, but because all warnings are treated as errors, the compilation fails in the UCX step. Compile UCX with gcc instead of clang (since we are not truly interested in running clang-tidy on ucx in this CI, but just on UCC)

Furthermore, force using rocm 5.6.1 for now instead of the most recent rocm release, since rocm-5.7 would require clang-14


Sergei-Lebedev (approved):

Sergei-Lebedev:
looks good to me, thanks for debugging the issue

manjugv (approved):

Issue #341, opened 2021-11-03
Docs refinement


Sergei-Lebedev:
- ucc_team_create_post returns UCC_OK or error, it doesn't return UCC_INPROGRESS if team is not created yet
- If ucc_team_create_post returns UCC_OK but later ucc_team_create_test returns error user is responsible to call ucc_team_destroy to free allocated resources
- If ucc_collective_init returns UCC_OK but later ucc_collective_post or ucc_collective_triggered_post or ucc_collective_test returns error user is responsbile to call ucc_collective_finalize to free allocated resources
- add ucc_error_type_t to ucc_team_params_t and ucc_context_params_t

PR #851, opened 2023-10-04
TL/NCCL: lazy init nccl comm

Sergei-Lebedev:
## What
Lazily initialize TL NCCL on first CUDA collective.

## Why ?
Both NCCL and CUDA require CUDA devices to be set before team create. In MPI workloads it's not always possible since MPI_Init creates UCC team and to set device we need to know rank and local rank.

replaces #758


src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -69,15 +75,17 @@ ucc_status_t ucc_tl_nccl_team_destroy(ucc_base_team_t *tl_team)
 #if NCCL_USE_NON_BLOCKING
     ncclResult_t nccl_status, st;
 
-    if (team->nccl_comm && team->comm_state == UCC_INPROGRESS) {
+    if (team->comm_state == TL_NCCL_COMM_STATE_DESTROY_COMM) {
```
samnordmann, line 78: can we merge this if with the one below ? Maybe avoid the goto?
I am not sure to fully understand the flow of this function.
Shouldn't we free the Cuda stream even if `team->comm_state == TL_NCCL_COMM_STATE_DESTROY_COMM` ?

src/components/tl/nccl/tl_nccl_team.c
```diff
@@ -101,95 +109,126 @@ ucc_status_t ucc_tl_nccl_team_destroy(ucc_base_team_t *tl_team)
...
+        if (!memcmp(&errorid, team->unique_id, sizeof(errorid))) {
+            tl_error(tl_team->context->lib, "incorrect unique id");
+            return status;
+        }
+
+        team->comm_state = TL_NCCL_COMM_STATE_INIT_TEAM;
+    }
+
+    // int debug = 1;
+    // while(debug);
```
samnordmann, line 224: remove this ?

src/components/tl/nccl/tl_nccl_team.c
Sergei-Lebedev: TL_NCCL_COMM_STATE_DESTROY_COMM means cleanup function was already called and currently inprogress. comm_destroy is nonblocking operation just like team create. 

>Shouldn't we free the Cuda stream even if team->comm_state == TL_NCCL_COMM_STATE_DESTROY_COMM

no, stream is already destroyed if state is TL_NCCL_COMM_STATE_DESTROY_COMM 

>can we merge this if with the one below ? Maybe avoid the goto?

goto is used to go to comm destroy progress

src/components/tl/ucc_tl.c
```diff
@@ -242,6 +242,11 @@ ucc_status_t ucc_tl_team_create_multiple(ucc_team_multiple_req_t *req)
     }
     req->descs[*id].status = UCC_TL_CTX_IFACE(req->descs[*id].ctx)
                                ->team.create_test(&req->descs[*id].team->super);
+    if (req->descs[*id].status < 0) {
+        /* if team create failed in team create test need to clenup resources */
+        UCC_TL_CTX_IFACE(req->descs[*id].ctx)->team.destroy(
```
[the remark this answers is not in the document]
Sergei-Lebedev: >Why isn't the responsability of team_create_test to cleanup the team in case of failure ?

This our internal agreement, see here https://github.com/openucx/ucc/issues/341

samnordmann (approved):

src/components/tl/nccl/tl_nccl.c
```diff
@@ -39,12 +39,17 @@ static ucs_config_field_t ucc_tl_nccl_context_config_table[] = {
      UCS_CONFIG_TYPE_ENUM(ucc_tl_nccl_completion_sync_names)
     },
 
-    {"BLOCKING", "1",
-     "If set to 0 will use non-blocking mode communicator behavior, "
-     "if set to 1 will use blocking mode",
+    {"BLOCKING", "yes",
+     "If set to yes will use non-blocking mode communicator behavior, "
```
bureddy, line 43: should it "if set to "no" for non-blocking

src/components/tl/nccl/tl_nccl_coll.c
```diff
@@ -143,6 +144,11 @@ ucc_status_t ucc_tl_nccl_init_task(ucc_base_coll_args_t *coll_args,
         return UCC_ERR_NOT_SUPPORTED;
     }
 
+    status = ucc_tl_nccl_comm_init(nccl_team);
```
bureddy, line 147: maybe we can call this func only if (team->comm_state != TL_NCCL_COMM_STATE_READY)  to avoid unnecessary for every task init.

Sergei-Lebedev: added

bureddy (approved):

PR #858, opened 2023-10-16
MC/ROCM: ROCm compatibility fixes

nileshnegi:
## What
- Adds condition based on HIP_VERSION for using `type` instead of `memoryType` in `hipPointerAttribute_t`.
- Adds switch case for `hipMemoryTypeManaged`

## Why ?
- From ROCm 5.5 onwards, `memoryType` is deprecated and `type` is used instead in `hipPointerAttribute_t`.
- From ROCm 5.7.1 onwards, `hipMallocManaged` pointers return `hipMemoryTypeManaged` instead of `hipMemoryTypeHost`.

### DIFF
--- a/src/components/mc/rocm/mc_rocm.c
+++ b/src/components/mc/rocm/mc_rocm.c
@@ -9,6 +9,7 @@
 #include "utils/ucc_malloc.h"
 #include "utils/arch/cpu.h"
 #include <hip/hip_runtime.h>
+#include <hip/hip_version.h>
 #include <hsa/hsa.h>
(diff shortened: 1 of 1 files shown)


edgargabriel:
> Looks good to me. Regarding managed memory support, I think proper memory allocation function is needed to support ROCM_MANAGED for algorithms with scratch buffer e.g. allreduce in TL/UCP

@Sergei-Lebedev thank you, yes, you are correct, it is something that we need to add to our to-do list for the near future.

samnordmann (approved):

PR #862, opened 2023-10-30
Sliding Window Allreduce

nsarka:
This PR implements Sliding Window Allreduce, to be used by the UROM UCC worker. The algorithm gets passed all of the data it needs (all host source bufs, destination bufs, src rkeys, dst rkeys, ucp workers, endpoints, ...) via the global work buffer. It's set up to use one ucp worker per thread. All ucp setup is done by the UROM UCC worker outside of this UCC change.

The algorithm itself is meant to run on the DPU.

### DIFF
--- a/src/components/tl/ucp/allreduce/allreduce.c
+++ b/src/components/tl/ucp/allreduce/allreduce.c
@@ -13,13 +13,17 @@ ucc_base_coll_alg_info_t
         [UCC_TL_UCP_ALLREDUCE_ALG_KNOMIAL] =
             {.id   = UCC_TL_UCP_ALLREDUCE_ALG_KNOMIAL,
              .name = "knomial",
-             .desc =
-                 "recursive knomial with arbitrary radix (optimized for latency)"},
+             .desc = "recursive knomial with arbitrary radix (optimized for "
+                     "latency)"},
         [UCC_TL_UCP_ALLREDUCE_ALG_SRA_KNOMIAL] =
             {.id   = UCC_TL_UCP_ALLREDUCE_ALG_SRA_KNOMIAL,
              .name = "sra_knomial",
              .desc = "recursive knomial scatter-reduce followed by knomial "
                      "allgather (optimized for BW)"},
+        [UCC_TL_UCP_ALLREDUCE_ALG_SLIDING_WINDOW] =
+            {.id   = UCC_TL_UCP_ALLREDUCE_ALG_SLIDING_WINDOW,
+             .name = "sliding_window",
+             .desc = "sliding window allreduce (optimized for running on DPU)"},
         [UCC_TL_UCP_ALLREDUCE_ALG_LAST] = {
             .id = 0, .name = NULL, .desc = NULL}};
 
@@ -45,4 +49,12 @@ ucc_status_t ucc_tl_ucp_allreduce_knomial_init(ucc_base_coll_args_t *coll_args,
     status = ucc_tl_ucp_allreduce_knomial_init_common(task);
 out:
     return status;
+}
+
+ucc_status_t
+ucc_tl_ucp_allreduce_sliding_window_init(ucc_base_coll_args_t *coll_args,
+                                         ucc_base_team_t *     team,
+                                         ucc_coll_task_t **    task_h)
+{
+    return UCC_OK;
 }
--- a/src/components/tl/ucp/allreduce/allreduce.c
+++ b/src/components/tl/ucp/allreduce/allreduce.c
@@ -56,5 +56,34 @@ ucc_tl_ucp_allreduce_sliding_window_init(ucc_base_coll_args_t *coll_args,
                                          ucc_base_team_t *     team,
                                          ucc_coll_task_t **    task_h)
 {
-    return UCC_OK;
+    ucc_status_t             status  = UCC_OK;
+    ucc_tl_ucp_team_t *      tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
+    ucc_tl_ucp_task_t *      task;
+    ucc_ee_executor_params_t params;
+
+    ALLREDUCE_TASK_CHECK(coll_args->args, tl_team);
+
+    task = ucc_tl_ucp_init_task(coll_args, team);
+    if (ucc_unlikely(!task)) {
+        ucc_error("couldnt allocate task");
+        return UCC_ERR_NO_MEMORY;
+    }
+    *task_h              = &task->super;
+    task->super.post     = ucc_tl_ucp_allreduce_sliding_window_start;
+    task->super.progress = ucc_tl_ucp_allreduce_sliding_window_progress;
+    task->super.finalize = ucc_tl_ucp_allreduce_sliding_window_finalize;
+
+    ucc_tl_ucp_allreduce_sliding_window_task_init(coll_args, team, task);
+
+    params.mask    = UCC_EE_EXECUTOR_PARAM_FIELD_TYPE;
+    params.ee_type = UCC_EE_CPU_THREAD;
+    status =
+        ucc_ee_executor_init(&params, &task->allreduce_sliding_window.executor);
+
+    if (UCC_OK != status) {
+        ucc_error("failed to init executor: %s", ucc_status_string(status));
+    }
+
+out:
+    return status;
 }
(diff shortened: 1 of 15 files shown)


manjugv:
@nsarka Does this run (optimally or sub optimally) on Hosts without DPUs?

nsarka:
> @nsarka Does this run (optimally or sub optimally) on Hosts without DPUs?

Yes, it should work suboptimally on Hosts without DPUs, so long as the global work buffer info struct is populated with the correct ucp info.

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -0,0 +1,323 @@
...
+                                + pipe->my_offset;
+            data_size       = count * dt_size;
+            src_rank        = pipe->src_rank;
+            getbuf          = accbuf->state == FREE ? 
+                                accbuf : &pipe->getbuf[get_idx];
+            src_addr        = task->allreduce_sliding_window.sbufs[src_rank]
+                                + get_offset;
+            dst_addr        = getbuf->buf;
+
+            assert(getbuf->state == FREE);
```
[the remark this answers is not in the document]
samnordmann: Is there a way to ensure that the buffer is free ? I see that `pipe->avail_buffs > 0` but is it sufficient to ensure that this precise buffer is free ?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -0,0 +1,323 @@
...
+            remaining_elems = pipe->my_count - pipe->count_received;
+            get_idx         = pipe->get_idx % pipe->num_buffers;
+            count           = ucc_min(pipe->buffer_size/dt_size,
+                                      remaining_elems);
+            get_offset      = pipe->count_received * dt_size
+                                + pipe->my_offset;
+            data_size       = count * dt_size;
+            src_rank        = pipe->src_rank;
+            getbuf          = accbuf->state == FREE ? 
+                                accbuf : &pipe->getbuf[get_idx];
```
samnordmann, line 182: What do you think about using a memory pool (see mpool) to dynamically request buffers from, instead of doing manually like here?

samnordmann: Sorry, can you please point me to where `pipe->getbuf` is allocated?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -0,0 +1,323 @@
...
+            src_addr        = task->allreduce_sliding_window.sbufs[src_rank]
+                                + get_offset;
+            dst_addr        = getbuf->buf;
+
+            assert(getbuf->state == FREE);
+
+            ucp_worker_fence(task->allreduce_sliding_window.worker);
```
samnordmann, line 189: is this fence really needed ?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -0,0 +1,323 @@
...
+            }
+
+            if (pipe->done_put == host_team_size) {
+                assert(pipe->avail_buffs == pipe->num_buffers);
+                assert(pipe->done_get == host_team_size);
+                assert(pipe->done_red == host_team_size-1);
+                assert(pipe->done_put == host_team_size);
+
+                pipe->count_serviced += accbuf->count;
+                accbuf->state         = FREE;
```
samnordmann, line 305: Why not using `ucc_tl_ucp_allreduce_sliding_window_reset_buf` here?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -0,0 +1,323 @@
...
+        {
+            remaining_elems = pipe->my_count - pipe->count_received;
+            get_idx         = pipe->get_idx % pipe->num_buffers;
+            count           = ucc_min(pipe->buffer_size/dt_size,
+                                      remaining_elems);
+            get_offset      = pipe->count_received * dt_size
+                                + pipe->my_offset;
+            data_size       = count * dt_size;
+            src_rank        = pipe->src_rank;
+            getbuf          = accbuf->state == FREE ?
```
samnordmann, line 181: would it be a good idea to merge accbuf into getbuf, and set accbuf to be the first entry of getbug ? Or do you prefer to keep them separated?

samnordmann (commented):
Hi Nicolas! Thank you for this PR which looks really great. I left a bunch of remarks, most of them are really minor and only deal with the codestyle. Let me stress here a couple of items:
- The collective needs to be non-blocking; however I fear that the progress function `ucc_tl_ucp_allreduce_sliding_window_req_test` is blocking and needs to be fixed
- I may be wrong but there might be an issue with `ucc_tl_ucp_allreduce_sliding_window_reduction`, which seems to only post the reduction but does not poll completion
- Can you please update the tests to test this algorithm as well?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
nsarka: They are allocated this way inside of the UROM ucc worker. There is one big allocation and getbuf is set to the 2nd buffer in the allocation, like this: 
```
accbuf = malloc(buffer_size * num_bufs);
getbufs = accbuf + buffer_size;
```

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
nsarka: The buffers are allocated in the UROM ucc worker here: https://gitlab-master.nvidia.com/hpc_sw_arch/urom/-/merge_requests/148/diffs#c81a79030f411e05cc204356423a4277e4216c04_0_58

Then, the pointers to the buffers are set and passed in via the global work buffer argument on the coll_args.

I am open to the idea of using a memory pool for these buffers. What would be the advantage? Right now, the UROM ucc worker allocates the buffers. It is reconfigurable--with a UROM command it can free and then allocate a different count of buffers with another size.

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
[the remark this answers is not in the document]
nsarka: If the get buf is not free, it is because it is either recv'ing or reducing. It cannot be recv'ing, because avail_bufs > 0 and get_idx is incremented after the buffer is set to recv'ing. It cannot be reducing, because red_idx is always behind get_idx.

src/components/tl/ucp/allreduce/allreduce.c
```diff
@@ -46,3 +50,65 @@ ucc_status_t ucc_tl_ucp_allreduce_knomial_init(ucc_base_coll_args_t *coll_args,
...
+    task->allreduce_sliding_window.pipe         = &gwbi_p->pipes[gwbi_p->tid];
+    task->allreduce_sliding_window.num_get_bufs = gwbi_p->num_bufs - 1;
+    task->allreduce_sliding_window.window_size  = gwbi_p->window_size;
+    task->allreduce_sliding_window.tid          = gwbi_p->tid;
+    task->allreduce_sliding_window.nthreads     = gwbi_p->nthreads;
+    task->allreduce_sliding_window.put_requests = gwbi_p->pipes[gwbi_p->tid].put_requests;
+
+    params.mask    = UCC_EE_EXECUTOR_PARAM_FIELD_TYPE;
+    params.ee_type = UCC_EE_CPU_THREAD;
+    status         = ucc_ee_executor_init(&params,
```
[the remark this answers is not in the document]
nsarka: If I don't initialize an executor here, then there is a segfault in the reduction that indicates it's trying to use an executor that doesn't exist. I have not debugged the issue further since initializing one here seemed to fix the problem. If I remember correctly, the reduction would call a function that looks for an executor on the task.super object, going upwards until it finds one. In this case, it never found one.

janjust:
> @nsarka Does this run (optimally or sub optimally) on Hosts without DPUs?

@manjugv We have not tested this on the host, moreover, We don't anticipate this performing better than SRA for example at scale.

nsarka:
> Hi Nicolas! Thank you for this PR which looks really great. I left a bunch of remarks, most of them are really minor and only deal with the codestyle. Let me stress here a couple of items:
> 
> * The collective needs to be non-blocking; however I fear that the progress function `ucc_tl_ucp_allreduce_sliding_window_req_test` is blocking and needs to be fixed
> * I may be wrong but there might be an issue with `ucc_tl_ucp_allreduce_sliding_window_reduction`, which seems to only post the reduction but does not poll completion
> * Can you please update the tests to test this algorithm as well?

Hi Sam, I have taken care of all of the above items. Please let me know if it looks good to you.

PR #872, opened 2023-11-07
TOOLS: use separate cuda alloc in perftest

Sergei-Lebedev:
## What
Add CUDA memory allocation function to ucc_perftest

## Why ?
Result of perftest might be biased due to use of internal memory pool in UCC.

### DIFF
--- a/tools/perf/ucc_pt_coll.cc
+++ b/tools/perf/ucc_pt_coll.cc
@@ -5,11 +5,29 @@
  */
 
 #include "ucc_pt_coll.h"
+#include "ucc_pt_cuda.h"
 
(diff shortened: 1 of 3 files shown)


samnordmann (approved):
Copyrights needs to be updated.

Can you explain why it is different for host mem type ?

Sergei-Lebedev:
> Copyrights needs to be updated.
> 
> Can you explain why it is different for host mem type ?

It's not different for host mem type, but host memory is less sensitive to memory pools. Update PR with host and cuda managed memory

bureddy (approved):

PR #876, opened 2023-11-08
TL/UCP: poll for onesided alltoall completion

wfaderhold21:
## What
Change onesided alltoall to poll for completion rather than perform a single test.


samnordmann (approved):

Sergei-Lebedev (approved):

Sergei-Lebedev:
ok to test

PR #882, opened 2023-11-22
TEST: fix memcopy after kernel launch

Sergei-Lebedev:
## What
Fixes gtest hangs in Jenkins. Memory copy might hang if persistent kernel is running


Sergei-Lebedev:
thanks to @B-a-S for reporting the issue

samnordmann (approved):

bureddy (approved):

PR #887, opened 2023-12-05
TL/MLX5: fix context create hang

Sergei-Lebedev:
## What
Fix hanging in TL MLX5 context create.

## How ?
PD_OWNER_RANK doesn't start service bcast If no IB devices found, other ranks hang in sbcast waiting for PD_OWNER_RANK


samnordmann (approved):
Is it valid to do sbcast if ppn=1 ?

bureddy (approved):

PR #897, opened 2024-01-15
TL/MLX5: fix port query

Sergei-Lebedev:
## What
Fixes segfault when converting port string to number. Device port string is freed before atoi


samnordmann (approved):

manjugv (approved):

bureddy (approved):

MamziB (commented):
looks good to me.

PR #789, opened 2023-06-02
TEST: do local check in mpi tests

Sergei-Lebedev:
## What
Compute correct buf locally instead of rely on MPI.

PR #901, opened 2024-01-19
TEST: do local checks for gather

Sergei-Lebedev:
## What
Do correctness check of Gather and Gatherv locally.
Continues work from https://github.com/openucx/ucc/pull/884 and https://github.com/openucx/ucc/pull/789

### DIFF
--- a/src/utils/ucc_log.h
+++ b/src/utils/ucc_log.h
@@ -67,6 +67,26 @@
 #define ucc_coll_trace_debug(_fmt, ...)                                        \
     ucc_log_component_collective_trace(UCS_LOG_LEVEL_DEBUG, _fmt, ##__VA_ARGS__)
 
+/**
+ * Print a message regardless of current log level. Output can be
+ * enabled/disabled via environment variable/configuration settings.
+ *
+ * During debugging it can be useful to add a few prints to the code
+ * without changing a current log level. Also it is useful to be able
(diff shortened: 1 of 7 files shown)


src/utils/ucc_log.h
```diff
@@ -67,6 +67,26 @@
...
     ucc_log_component_collective_trace(UCS_LOG_LEVEL_DEBUG, _fmt, ##__VA_ARGS__)
 
+/**
+ * Print a message regardless of current log level. Output can be
+ * enabled/disabled via environment variable/configuration settings.
+ *
+ * During debugging it can be useful to add a few prints to the code
+ * without changing a current log level. Also it is useful to be able
+ * to see messages only from specific processes. For example, one may
+ * want to see prints only from rank 0 when debugging MPI.
```
samnordmann, line 77: great, this is gonna be useful. Can you please indicate what environment variable we need to set to print, and to only select rank 0 ?

samnordmann (approved):

src/utils/ucc_log.h
Sergei-Lebedev: You don't need to set any environment variables, it works just like printf but also includes additional information (hostname, pid, etc.) so that you can filter out ranks.

bureddy (approved):

PR #903, opened 2024-01-23
TL/UCP: enable linear alltoall

Sergei-Lebedev:
## What
Choose num_posts parameter for alltoall and alltoallv so that progress function posts all send and recv operations immediately for small communicators effectively enabling linear algorithm
Parameters are chosen based on previous experiments with HCOLL


src/components/tl/ucp/alltoall/alltoall_pairwise.c
```diff
@@ -11,6 +11,10 @@
 #include "utils/ucc_math.h"
 #include "tl_ucp_sendrecv.h"
 
+/* TODO: add as parameters */
```
samnordmann, line 14: copyright... :)

samnordmann (approved):

src/components/tl/ucp/alltoall/alltoall_pairwise.c
Sergei-Lebedev: done

manjugv (approved):

PR #904, opened 2024-01-24
TOPO: fix rank reordering for host ordered sbgp

Sergei-Lebedev:
## What
Add check if reordering is required before computing it.

## Why ?
Previous version of rank reordering algorithm is based on host hash, socket id and numa id. This algorithm can't detect if reordering is required and sometimes might generate permutation that mixes nodes but not ranks within node. Even though such permutation is equivalent to "no ordering" it still disables some algorithms (e.g. neighbor exchange allgather).


artemry-nv:
bot:retest

bureddy (approved):

PR #909, opened 2024-01-31
Add Support for Dynamic Symmetric Memory

wfaderhold21:
## What
Currently, UCC has support for symmetric memory segments that are added during context creation for use with one-sided collectives. This PR adds support for memory segments to be added after a context and team have been created to better support programming models such as MPI. In addition, this adds support for attaching a key with the memory region for use with DPUs and offloading of collectives. 

## Why ?
With the current approach, a user will need to perform a memory copy of source and destination buffers to a known and already mapped memory region before a collective can begin using one-sided or offloaded collectives. This removes the need for the memory copy at the expense of memory exchanges on the first collective. This is useful if multiple collectives occur with the same buffers.

### DIFF
--- a/src/components/tl/ucp/alltoall/alltoall_onesided.c
+++ b/src/components/tl/ucp/alltoall/alltoall_onesided.c
@@ -9,6 +9,7 @@
 #include "alltoall.h"
 #include "core/ucc_progress_queue.h"
 #include "utils/ucc_math.h"
+#include "tl_ucp_coll.h"
 #include "tl_ucp_sendrecv.h"
 
 void ucc_tl_ucp_alltoall_onesided_progress(ucc_coll_task_t *ctask);
@@ -23,24 +24,28 @@ ucc_status_t ucc_tl_ucp_alltoall_onesided_start(ucc_coll_task_t *ctask)
     ucc_rank_t         grank  = UCC_TL_TEAM_RANK(team);
     ucc_rank_t         gsize  = UCC_TL_TEAM_SIZE(team);
     ucc_rank_t         start  = (grank + 1) % gsize;
-    long *             pSync  = TASK_ARGS(task).global_work_buffer;
+    long              *pSync  = TASK_ARGS(task).global_work_buffer;
+    ucc_memory_type_t  mtype  = TASK_ARGS(task).src.info.mem_type;
     ucc_rank_t         peer;
 
     ucc_tl_ucp_task_reset(task, UCC_INPROGRESS);
+    ucc_tl_ucp_coll_dynamic_segments(&TASK_ARGS(task), task);
+
     /* TODO: change when support for library-based work buffers is complete */
     nelems = (nelems / gsize) * ucc_dt_size(TASK_ARGS(task).src.info.datatype);
     dest   = dest + grank * nelems;
     UCPCHECK_GOTO(ucc_tl_ucp_put_nb((void *)(src + start * nelems),
-                                    (void *)dest, nelems, start, team, task),
+                                    (void *)dest, nelems, start, mtype, team,
+                                    task),
                   task, out);
     UCPCHECK_GOTO(ucc_tl_ucp_atomic_inc(pSync, start, team), task, out);
 
     for (peer = (start + 1) % gsize; peer != start; peer = (peer + 1) % gsize) {
         UCPCHECK_GOTO(ucc_tl_ucp_put_nb((void *)(src + peer * nelems),
-                                        (void *)dest, nelems, peer, team, task),
+                                        (void *)dest, nelems, peer, mtype, team,
+                                        task),
                       task, out);
-        UCPCHECK_GOTO(ucc_tl_ucp_atomic_inc(pSync, peer, team), task,
-                      out);
+        UCPCHECK_GOTO(ucc_tl_ucp_atomic_inc(pSync, peer, team), task, out);
     }
 
     return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
--- a/src/components/tl/ucp/alltoall/alltoall_onesided.c
+++ b/src/components/tl/ucp/alltoall/alltoall_onesided.c
@@ -27,9 +27,14 @@ ucc_status_t ucc_tl_ucp_alltoall_onesided_start(ucc_coll_task_t *ctask)
     long              *pSync  = TASK_ARGS(task).global_work_buffer;
     ucc_memory_type_t  mtype  = TASK_ARGS(task).src.info.mem_type;
     ucc_rank_t         peer;
+    ucc_status_t       status;
 
     ucc_tl_ucp_task_reset(task, UCC_INPROGRESS);
-    ucc_tl_ucp_coll_dynamic_segments(&TASK_ARGS(task), task);
+    status = ucc_tl_ucp_coll_dynamic_segment_exchange(task);
+    if (UCC_OK != status) {
+        task->super.status = status;
+        goto out;
+    }
 
     /* TODO: change when support for library-based work buffers is complete */
     nelems = (nelems / gsize) * ucc_dt_size(TASK_ARGS(task).src.info.datatype);
@@ -66,4 +71,5 @@ void ucc_tl_ucp_alltoall_onesided_progress(ucc_coll_task_t *ctask)
 
     pSync[0]           = 0;
     task->super.status = UCC_OK;
+    ucc_tl_ucp_coll_dynamic_segment_finalize(task);
 }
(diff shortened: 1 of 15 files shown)


manjugv:
@wfaderhold21  Is this ready for review? @janjust Does this work for you?

janjust:
@manjugv In general this works, but I'm still a little concerned if keeping mapped data until tl destruction is the right approach, or if we should enable symmetric buffer unmapping.

wfaderhold21:
@manjugv yes. It’s ready for review.

janjust:
@wfaderhold21 Did we address the possibility of adding an imported key? What about the ability to important keys with an API? I think we did a follow up study and concluded that parallelization does not improve perf much, and requires a unique memory region to handle key imports otherwise we run into segfaults.

janjust:
OK - after WG meeting I think I realize that my point is moot. I think this is good as is, we just need to put a protection for when keys are imported by multiple threads.

wfaderhold21:
@Sergei-Lebedev @janjust was there any more feedback on this PR?

src/components/tl/ucp/tl_ucp_context.c
```diff
@@ -477,22 +492,42 @@ ucc_status_t ucc_tl_ucp_ctx_remote_populate(ucc_tl_ucp_context_t * ctx,
...
+
+            status = ucp_mem_map(ctx->worker.ucp_context, &mmap_params, &mh);
+            if (UCS_OK != status) {
+                tl_error(ctx->super.super.lib,
+                         "ucp_mem_map failed with error code: %d", status);
+                ucc_status = ucs_status_to_ucc_status(status);
+                goto fail_mem_map;
+            }
+            ctx->remote_info[i].packed_memh = NULL;
+        } else {
```
nsarka, line 509: Should this `else` have a check for the `use_xgvmi` field? If use_xgvmi was "n" then error?

janjust (approved):
LGTM!

nsarka (approved):

manjugv (approved):

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -155,6 +155,297 @@ ucc_status_t ucc_tl_ucp_coll_finalize(ucc_coll_task_t *coll_task)
...
+
+        maps[1].address = coll_args->dst.info.buffer;
+        maps[1].len = (coll_args->dst.info.count / UCC_TL_TEAM_SIZE(tl_team)) *
+                      ucc_dt_size(coll_args->dst.info.datatype);
+        maps[1].resource = NULL;
+
+        n_segments = 2;
+    }
+
+    ctx->dynamic_remote_info =
```
samnordmann, line 263: please check if null

wfaderhold21:
@samnordmann @nsarka I think I made the corrections you all asked, but I also updated some code to bugs I found. Could you all take another look to make sure everything is OK there?

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -234,45 +234,101 @@ ucc_status_t ucc_tl_ucp_memmap_segment(ucc_tl_ucp_task_t *task,
...
                                                   ucc_tl_ucp_task_t *task)
 {
-    ucc_tl_ucp_team_t    *tl_team = UCC_TL_UCP_TASK_TEAM(task);
-    ucc_tl_ucp_context_t *ctx     = UCC_TL_UCP_TEAM_CTX(tl_team);
-    int                   i       = 0;
-    ucc_status_t          status;
+    ucc_tl_ucp_team_t    *tl_team    = UCC_TL_UCP_TASK_TEAM(task);
+    ucc_tl_ucp_context_t *ctx        = UCC_TL_UCP_TEAM_CTX(tl_team);
+    int                   i          = 0;
+    uint64_t              need_map   = 0x7;
```
samnordmann, line 240: It would be preferable to use the generic interface for masks, for more readability. It requires defining a enum type instead of hardcoding `1`, `2`, `4`, e.g.,
```
typedef enum {
    UCC_TL_UCP_DYN_SEG_UPDATE_SRC    = UCC_BIT(0),
    UCC_TL_UCP_DYN_SEG_UPDATE_DST    = UCC_BIT(1),
    UCC_TL_UCP_DYN_SEG_UPDATE_GLOBAL = UCC_BIT(2),
} ucc_tl_ucp_dynamic_segment_update_mask_t;
```
, and to express `0x7` as `UCC_TL_UCP_DYN_SEG_UPDATE_SRC | UCC_TL_UCP_DYN_SEG_UPDATE_DST | UCC_TL_UCP_DYN_SEG_UPDATE_GLOBAL`
see, for example:
https://github.com/openucx/ucc/blob/985947a521d7c00e6866f032b4db4c18e5bf9e6a/src/ucc/api/ucc_def.h#L124

src/components/tl/ucp/alltoall/alltoall_onesided.c
```diff
@@ -61,4 +71,5 @@ void ucc_tl_ucp_alltoall_onesided_progress(ucc_coll_task_t *ctask)
 
     pSync[0]           = 0;
     task->super.status = UCC_OK;
+    ucc_tl_ucp_coll_dynamic_segment_finalize(task);
```
samnordmann, line 74: Now that this function returns an error code, should we handle it ?

src/components/tl/ucp/tl_ucp_coll.c
```diff
@@ -405,19 +462,24 @@ ucc_status_t ucc_tl_ucp_coll_dynamic_segment_exchange(ucc_tl_ucp_task_t *task)
...
     if (ctx->dyn_seg_buf) {
         /* unmap and release packed buffers */
         for (i = 0; i < ctx->n_dynrinfo_segs; i++) {
             if (ctx->dynamic_remote_info[i].mem_h) {
-                ucp_mem_unmap(ctx->worker.ucp_context,
+                status = ucp_mem_unmap(ctx->worker.ucp_context,
                               ctx->dynamic_remote_info[i].mem_h);
+                if (UCS_OK != status) {
+                    tl_error(UCC_TL_UCP_TEAM_LIB(tl_team), "Failed to unmap memory");
+                    return ucs_status_to_ucc_status(status);
```
samnordmann, line 481: nit: shouldn't we, even in the case of an error, still try to free and release the remaining resource?

samnordmann (approved):
LGTM!

src/components/tl/ucp/tl_ucp_coll.c
wfaderhold21: done. thanks!

src/components/tl/ucp/tl_ucp_coll.c
wfaderhold21: done

src/components/tl/ucp/alltoall/alltoall_onesided.c
wfaderhold21: i've changed this so the error code will be passed to the user as there would be something significantly wrong in order to fail to unmap.

janjust:
@Sergei-Lebedev @wfaderhold21 Hey guys, sorry to harp on this, but are we done with this PR? Is it ready to be merged, @Sergei-Lebedev any other input needed?

wfaderhold21:
@janjust AFIAK nothing to be added on my end.

wfaderhold21:
@Sergei-Lebedev @manjugv are there any outstanding comments? Can we push this forward?

manjugv:
> @Sergei-Lebedev @manjugv are there any outstanding comments? Can we push this forward?

+1 from my side.

wfaderhold21:
@Sergei-Lebedev @janjust @manjugv The requested changes were made to remove all API changes. I believe this PR now conforms to what we have discussed.

janjust:
@wfaderhold21 is this safe to "update with rebase"?

janjust:
Dependent on #1037

wfaderhold21:
> Dependent on #1037

This is different from #1037 and can be considered separately I believe

janjust:
> This is different from https://github.com/openucx/ucc/pull/1037 and can be considered separately I believe

@wfaderhold21 with the changes in #1037 is this true? I think we decided to let user do the allgather?

wfaderhold21:
Waiting on PR #1070 to reuse code from the symmetric mem map api as it will simplify this code path.

wfaderhold21:
closing until #1070 merged. will update code base and bring back afterwards.

janjust:
@wfaderhold21 now that #1070 is merged, can we reopen? Or will you reopen after you update the PR?

wfaderhold21:
@janjust I will reopen after I finish porting the PR. This should be soon.

PR #914, opened 2024-02-01
BUILD: fix GPU arch flags in amdclang build

nileshnegi:
## What
- Modified GPU architecture flags for compatibility with AMD GPUs.
- Add `-O3` for improving performance.

## Why ?
Incompatibility between explicitly specified GPU arch flags (like `--target-offload=gfx90a:xnack+`) and `--target-offload=native` led to build errors on some AMD GPUs.


Sergei-Lebedev (approved):

Sergei-Lebedev:
ok to test

PR #918, opened 2024-02-05
Nsarka/ucc int8 reduce bugfix

nsarka:
This PR fixes int8_t reductions not getting vectorized when compiled with GCC.

## Background

A sum reduction may look like this:

```
 size_t _i, count = ...
 int8_t * src_1 = ...
 int8_t * src_2 = ...
 int8_t * dst = ...

  for (_i = 0; _i < count; _i++) {
      d[_i] = src_1[_i] + src_2[_i]);
  }  
```

When `src_1` and `src_2` are arrays of type int8, the performance drops. The reason is that `int8` is a typedef for `char`, and `char` is the only type allowed to read and modify memory occupied by other types. (https://gist.github.com/alexei-zaripov/dcc14c78819c5f1354afe8b70932007c?permalink_comment_id=4677393). This means that the compiler does not know that `d[_i] = src_1[_i] + src_2[_i]);` isn't changing `src_1` or `src_2`, and thus can't vectorize the loop.

The `restrict` keyword--introduced in the C99 standard--allows the programmer to tell the compiler that the pointers do not overlap in any way and thus can be vectorized.

## Bug

Even though UCC already makes a new restricted pointer in ec_cpu_reduce.c:78 and uses that pointer in the reduction, what I assume is a bug in GCC makes it so the `restrict` keyword is not respected and loop auto-vectorization does not take place.

## Solution

Through some debugging, I found that GCC will respect the `restrict` keyword, but only if the pointer is passed via a function argument. See https://godbolt.org/z/feP766obr for a minimal reproducer of this issue.
To get the fix into ucc, I made the `ucc_ec_cpu_reduce` function take in a restrict pointer to the input vectors. Also, I made a copy of the _count variable because the loop bounds are indeterminate (task_args is not a const pointer, so task_args->reduce.count could change).

### DIFF
--- a/src/components/ec/cpu/ec_cpu.c
+++ b/src/components/ec/cpu/ec_cpu.c
@@ -114,7 +114,11 @@ ucc_status_t ucc_cpu_executor_task_post(ucc_ee_executor_t *executor,
     switch (task_args->task_type) {
     case UCC_EE_EXECUTOR_TASK_REDUCE:
         status = ucc_ec_cpu_reduce((ucc_eee_task_reduce_t *)&task_args->reduce,
-                                   task_args->flags);
+                                   (task_args->flags &
+                                        UCC_EEE_TASK_FLAG_REDUCE_SRCS_EXT) ?
+                                        task_args->reduce.srcs_ext :
+                                        task_args->reduce.srcs,
+                                    task_args->flags);
         if (ucc_unlikely(UCC_OK != status)) {
             goto free_task;
         }
@@ -147,7 +151,7 @@ ucc_status_t ucc_cpu_executor_task_post(ucc_ee_executor_t *executor,
         tr.dst    = trs->dst;
         tr.alpha  = trs->alpha;
 
-        status = ucc_ec_cpu_reduce(&tr, flags);
+        status = ucc_ec_cpu_reduce(&tr, srcs, flags);
         if (ucc_unlikely(UCC_OK != status)) {
             goto free_task;
         }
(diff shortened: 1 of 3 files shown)


janjust:
This fixes a performance issue found when running osu_allreduce, they defaulted to using char vectors.

Sergei-Lebedev:
Ok to test

Sergei-Lebedev:
@nsarka thanks for the fix! It looks good to me, but please fix commit titles to pass codestyle check

src/components/ec/cpu/ec_cpu.c
```diff
@@ -147,7 +151,12 @@ ucc_status_t ucc_cpu_executor_task_post(ucc_ee_executor_t *executor,
         tr.dst    = trs->dst;
         tr.alpha  = trs->alpha;
 
-        status = ucc_ec_cpu_reduce(&tr, flags);
+        status = ucc_ec_cpu_reduce((ucc_eee_task_reduce_t *)&task_args->reduce,
+                                   (task_args->flags &
+                                        UCC_EEE_TASK_FLAG_REDUCE_SRCS_EXT) ?
+                                        task_args->reduce.srcs_ext :
+                                        task_args->reduce.srcs,
+                                    task_args->flags);
```
Sergei-Lebedev, line 159: it should be tr instead of task_args

samnordmann (approved):
Looks great thanks!

Sergei-Lebedev (approved):

PR #919, opened 2024-02-06
CL/HIER: check number of TLs per SBGP

Sergei-Lebedev:
## What
Add check to CL/HIER to prevent SBGP TL list overflow.

## Why ?
For some systems CL/HIER might try to use more than 4 TLs per SBGP. 4 is compile time constant. Fixes bug https://redmine.mellanox.com/issues/3767158

## How ?
Ignore TLs if the list already full. Filter out TLs if we know in advance that SBGP size is not supported.

### DIFF
--- a/src/components/cl/hier/cl_hier_team.c
+++ b/src/components/cl/hier/cl_hier_team.c
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * Copyright (c) 2020-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See file LICENSE for terms.
  */
@@ -43,6 +43,11 @@ UCC_CLASS_INIT_FUNC(ucc_cl_hier_team_t, ucc_base_context_t *cl_context,
     ucc_config_names_array_t  *tls;
     ucc_subset_t               subset;
     struct ucc_team_team_desc *d;
+    ucc_tl_context_t          *tl_ctx;
+    ucc_tl_lib_t              *tl_lib;
+    ucc_base_lib_attr_t        attr;
+
+
     if (!params->team->topo) {
         cl_debug(cl_context->lib,
                 "can't create hier team without topology data");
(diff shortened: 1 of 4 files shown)


test/mpi/main.cc
```diff
@@ -574,15 +574,17 @@ void ProcessArgs(int argc, char** argv)
 
 int main(int argc, char *argv[])
 {
-    int failed                                                          = 0;
-    int total_done_skipped_failed[ucc_ilog2(UCC_COLL_TYPE_LAST) + 1][4] = {0};
+    int failed = 0;
+    int total_done_skipped_failed[ucc_ilog2(UCC_COLL_TYPE_LAST) + 1][4];
```
janjust, line 578: I guess unnecessary change for the bug, but does remove a compiler warning.

janjust (approved):

src/components/cl/hier/cl_hier_team.c
```diff
@@ -74,18 +79,51 @@ UCC_CLASS_INIT_FUNC(ucc_cl_hier_team_t, ucc_base_context_t *cl_context,
             hs->n_tls = 0;
             tls       = &lib->cfg.sbgp_tls[i].array;
```
bureddy, line 80: qq: is orders of TLS in this array always will be same? do you see any issues if there is a mismatched list of TLS selected for team on different ranks?

bureddy (approved):

test/mpi/main.cc
Sergei-Lebedev: you are right, it just fixes gcc 4.8.5 warning when build ucc_test_mpi

src/components/cl/hier/cl_hier_team.c
Sergei-Lebedev: it matters, there is separate fix for this https://github.com/openucx/ucc/pull/717, but it was rejected since the bigger change in UCC is needed to detect mismatches for every CL and core

src/components/cl/hier/cl_hier_team.c
```diff
@@ -74,18 +79,51 @@ UCC_CLASS_INIT_FUNC(ucc_cl_hier_team_t, ucc_base_context_t *cl_context,
...
             for (j = 0; j < tls->count; j++) {
+                if (hs->n_tls == CL_HIER_MAX_SBGP_TLS) {
+                    cl_debug(cl_context->lib,
+                             "skipping tl context %s for %s sbgp: "
+                             "max number of TLs per SBGP is reached",
+                             tls->names[j], ucc_sbgp_str(hs->sbgp_type));
+                    continue;
```
samnordmann, line 87: break ?

samnordmann (approved):

src/components/cl/hier/cl_hier_team.c
Sergei-Lebedev: we reached max and could break, but we want to print in debug msg what TLs were skipped

artemry-nv:
bot:retest

PR #924, opened 2024-02-12
CORE: add progress throttle

Sergei-Lebedev:
## What
Add progress throttle to ucc_context_progress()

## Why ?
Point to point benchmarks might be affected by ucc_context_progress() even though collectives are not used there. Progress throttle helps to save cycles and not calling ucp_worker_progress (in TL UCP) if there are no active collective requests.

### DIFF
--- a/src/components/tl/ucp/bcast/bcast_dbt.c
+++ b/src/components/tl/ucp/bcast/bcast_dbt.c
@@ -107,6 +107,8 @@ void ucc_tl_ucp_bcast_dbt_progress(ucc_coll_task_t *coll_task)
     task->bcast_dbt.state = SEND_T1;
 
 SEND_T1:
+/* test_recv is needed to progress ucp_worker */
+    ucc_tl_ucp_test_recv(task);
     if ((coll_root == rank) || (task->bcast_dbt.t1.recv > 0)) {
         for (i = 0; i < 2; i++) {
             if ((t1.children[i] != UCC_RANK_INVALID) &&
@@ -122,6 +124,8 @@ SEND_T1:
     task->bcast_dbt.state = SEND_T2;
 
 SEND_T2:
+/* test_recv is needed to progress ucp_worker */
+    ucc_tl_ucp_test_recv(task);
     if ((coll_root == rank) || (task->bcast_dbt.t2.recv > 0)) {
         for (i = 0; i < 2; i++) {
             if ((t2.children[i] != UCC_RANK_INVALID) &&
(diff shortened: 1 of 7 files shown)


janjust (approved):

src/core/ucc_context.h
```diff
@@ -77,6 +77,8 @@ typedef struct ucc_context {
     ucc_context_topo_t      *topo;
     uint64_t                 cl_flags;
     ucc_tl_team_t           *service_team;
+
```
samnordmann, line 80: extra blank line

src/core/ucc_progress_queue_mt.c
```diff
@@ -100,6 +101,20 @@ static int ucc_pq_mt_progress(ucc_progress_queue_t *pq)
     return n_progressed;
 }
 
+static int ucc_pq_locked_mt_is_empty(ucc_progress_queue_t *pq)
+{
+    ucc_pq_mt_locked_t *pq_mt = ucc_derived_of(pq, ucc_pq_mt_locked_t);
+
+/* this function should not be very accurate for the purpose of progress throttling */
```
samnordmann, line 108: I am not sure to understand the comment "should not be very accurate"

Also need to fix indentation. And maybe prefer inline comment style with "//" ?

src/core/ucc_progress_queue_mt.c
```diff
@@ -100,6 +101,20 @@ static int ucc_pq_mt_progress(ucc_progress_queue_t *pq)
...
+    return ucc_list_is_empty(&pq_mt->queue);
+}
+
+static int ucc_pq_mt_is_empty(ucc_progress_queue_t *pq) //NOLINT: pq is unused
+{
+/* lock free progress queue never use throttling */
```
samnordmann, line 114: Why not? I ask for my own understanding

src/components/tl/ucp/bcast/bcast_dbt.c
```diff
@@ -107,6 +107,8 @@ void ucc_tl_ucp_bcast_dbt_progress(ucc_coll_task_t *coll_task)
     task->bcast_dbt.state = SEND_T1;
 
 SEND_T1:
+/* test_recv is needed to progress ucp_worker */
+    ucc_tl_ucp_test_recv(task);
```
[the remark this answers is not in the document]
Sergei-Lebedev: This bug reproduces only with this PR. bcast_dbt and reduce_dbt suppose to progress ucp_worker in tl_ucp_reduce_dbt_progress and tl_ucp_bcast_dbt_progress.

src/core/ucc_progress_queue_mt.c
Sergei-Lebedev: There is a race condition here which might result in returning false result from is_empty(). To avoid race condition we need to use locks, however for performance it's better to call progress few extra times comparing to locking/unlocking  every single is_empty() call.

samnordmann: got it! Interesting

samnordmann (approved):

bureddy (approved):

PR #946, opened 2024-03-21
check MCAST/A2A ctx/team ready

MamziB:
check MCAST/A2A ctx/team ready before creating mlx5 teams and also check after mlx5 team creation to make sure every process created mcast/a2a team successfully


janjust:
@samnordmann @Sergei-Lebedev this is a critical PR for the April release

janjust (approved):
Overall looks good to me.

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -92,7 +95,9 @@ UCC_CLASS_CLEANUP_FUNC(ucc_tl_mlx5_team_t)
     tl_debug(self->super.super.context->lib, "finalizing tl team: %p", self);
 
     ucc_tl_mlx5_dm_cleanup(self);
-    ucc_tl_mlx5_alltoall_cleanup(self);
+    if (self->a2a_state != TL_MLX5_TEAM_STATE_ALLTOALL_NOT_AVAILABLE) {
```
samnordmann, line 98: just to make sure: was this necessary? I thought I made sure this was ok. Have you check there is no case of memory leak here if we just skip the cleanup?

src/components/tl/mlx5/tl_mlx5.h
```diff
@@ -140,7 +137,11 @@ typedef struct ucc_tl_mlx5_team {
     ucc_tl_mlx5_alltoall_t         *a2a;
     ucc_topo_t                     *topo;
     ucc_ep_map_t                    ctx_map;
+    int                             local_mcast_ctx_ready;
     ucc_tl_mlx5_mcast_team_t       *mcast;
+    ucc_status_t                   *local_status_array;
+    ucc_status_t                   *global_status_array;
+    int                             global_sync_in_progress;
```
samnordmann, line 144: in my opinion this bool flag is not necessary, the check can be done with `tl_team->scoll_req`

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,169 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+
+        return UCC_OK;
+    }
+
+    ucc_assert(!tl_team->global_sync_in_progress);
+    
+    if (tl_team->mcast_state == TL_MLX5_TEAM_STATE_INIT &&
+            tl_team->a2a_state == TL_MLX5_TEAM_STATE_INIT ) {
+        // check if ctx is ready for a2a and mcast
+        tl_team->local_status_array  = ucc_calloc(1, 2*sizeof(ucc_status_t));
```
samnordmann, line 244: those two arrays should be allocated statically. I would also suggest that indexing into it (occuring in many places in this file) should be made with macros or enum type instead of hardcoded, like
```
#define TL_MLX5_A2A_STATUS_INDEX 0
#define TL_MLX5_MCAST_STATUS_INDEX 1
```

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,169 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+        tl_team->global_status_array = ucc_calloc(1, 2*sizeof(ucc_status_t));
+
+        tl_team->local_status_array[0] = tl_team->a2a_status.local;
+        tl_team->local_status_array[1] = tl_team->local_mcast_ctx_ready?UCC_OK:UCC_ERR_NO_RESOURCE;
+
+        status = ucc_service_allreduce(
+            core_team, tl_team->local_status_array, tl_team->global_status_array,
+            UCC_DT_INT32, 2, UCC_OP_MIN, subset, &tl_team->scoll_req);
```
samnordmann, line 252: Also here the "2" is hardcoded for the service coll and for the allocation size in many places, but it should be the same literal value defined by a macro

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,169 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+        }
+
+        ucc_service_coll_finalize(tl_team->scoll_req);
+
+        tl_team->global_sync_in_progress = 0;
+
+        if (tl_team->mcast_state == TL_MLX5_TEAM_STATE_INIT &&
+            tl_team->a2a_state == TL_MLX5_TEAM_STATE_INIT ) {
+            tl_team->a2a_status.global = tl_team->global_status_array[0];
+            tl_team->a2a_state         = TL_MLX5_TEAM_STATE_ALLTOALL_INIT;
```
samnordmann, line 178: I am a bit confused between the different state enum, what is the high level idea? Like here why do we move from
`tl_team->a2a_state = TL_MLX5_TEAM_STATE_INIT` to `tl_team->a2a_state = TL_MLX5_TEAM_STATE_ALLTOALL_INIT` ?

src/components/tl/mlx5/tl_mlx5.h
```diff
@@ -108,19 +108,15 @@ typedef enum
...
+    TL_MLX5_TEAM_STATE_ALLTOALL_NOT_AVAILABLE,
     TL_MLX5_TEAM_STATE_MCAST_INIT,
     TL_MLX5_TEAM_STATE_MCAST_GRP_JOIN_POST,
     TL_MLX5_TEAM_STATE_MCAST_GRP_JOIN_READY,
     TL_MLX5_TEAM_STATE_MCAST_GRP_JOIN_FAILED,
     TL_MLX5_TEAM_STATE_MCAST_GRP_BCAST_POST,
     TL_MLX5_TEAM_STATE_MCAST_READY,
     TL_MLX5_TEAM_STATE_MCAST_NOT_AVAILABLE
-} ucc_tl_mlx5_team_mcast_state_t;
+} ucc_tl_mlx5_team_state_t;
```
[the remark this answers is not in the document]
samnordmann: if we already use two different field `mcast_state` and `a2a_state` of this enum type, then the enum shouldn't contain `alltoall` or `mcast`

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,169 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+    }
 
     a2a_status = ucc_tl_mlx5_a2a_team_test(team);
     if (a2a_status < 0) {
-        tl_error(team->context->lib, "ALLTOALL tl team: %p creation failed %d",
-                 team, a2a_status);
+        tl_warn(team->context->lib, "ALLTOALL tl team: %p creation failed %d",
+                team, a2a_status);
         tl_team->a2a_state = TL_MLX5_TEAM_STATE_ALLTOALL_NOT_AVAILABLE;
     }
```
samnordmann, line 268: I think it would be safer to return at this point depending on `a2a_status` to ensure a proper ordering between a2a and mcast init. I'm just worried that service collective would be called together and collide. What do you think?

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,169 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
             tl_team->mcast_state = TL_MLX5_TEAM_STATE_MCAST_NOT_AVAILABLE;
         }
     }
 
     if (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_NOT_AVAILABLE &&
         tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_NOT_AVAILABLE) {
-        tl_error(team->context->lib, "unable to initialize tl team: %p", team);
-        return UCC_ERR_NO_RESOURCE;
+        tl_warn(team->context->lib, "unable to initialize tl team: %p", team);
+        goto global_sync_post;
```
samnordmann, line 282: why do we need this goto?

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,169 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
         return UCC_INPROGRESS;
     }
 
-    tl_debug(team->context->lib, "initialized tl team: %p: MCAST component is %s ALLTOALL component is %s",
-            team, (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_READY)?"ENABLED":"DISABLED",
-            (tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_READY)?"ENABLED":"DISABLED");
-    return UCC_OK;
+
+global_sync_post:
+    tl_team->local_status_array[0] = (tl_team->a2a_state ==
```
samnordmann, line 291: this block from 
`tl_team->local_status_array[0]` to `tl_team->global_sync_in_progress = 1;` is duplicated from above and should be factored out

src/components/tl/mlx5/mcast/tl_mlx5_mcast_context.c
```diff
@@ -289,14 +289,14 @@ ucc_status_t ucc_tl_mlx5_mcast_clean_ctx(ucc_tl_mlx5_mcast_coll_context_t *ctx)
 
     if (ctx->pd) {
         if (ibv_dealloc_pd(ctx->pd)) {
-            tl_error(ctx->lib, "ibv_dealloc_pd failed errno %d", errno);
+            tl_warn(ctx->lib, "ibv_dealloc_pd failed errno %d", errno);
```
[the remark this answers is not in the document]
MamziB: Because it is not fatal and it is ok to continue.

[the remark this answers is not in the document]
samnordmann: but there is no fallback in this case, so it is an actual error that leads to memory leak

src/components/tl/mlx5/tl_mlx5.h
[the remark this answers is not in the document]
MamziB: _isn't there a problem here if we want, for example, tag both mcast and a2a as "not available"?_
I do not think there is problem. We already clean the team resources if they become unavailable.

_if we already use two different field mcast_state and a2a_state of this enum type, then the enum shouldn't contain alltoall or mcast_
I did not want to have two separate sets of enums for mcast and a2a. This way is easier to maintain. Furthermore, I think It is OK that some of the states in this mlx5 enum set are not applicable to a feature (a2a, mcast, or any future features).

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: No, I have not checked that. However, during team create, if the state becomes not-available, we already have called the cleanup function over there (given the fact that local a2a is READY but for some reasons, some of the process are not READY). So that is why we are avoiding the cleanup again. But if you think it is OK to do cleanup again, I can remove this check here.

src/components/tl/mlx5/tl_mlx5.h
MamziB: I agree, but this is just more readable and easier to understand. Especially that the scoll_req variable is used for many scenarios, not just for team create global status check.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: Sure I will use macro.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: Previously (before this patch) we had two more states where we do allreduce for allltoall. Before doing allreduce, the state was TL_MLX5_TEAM_STATE_INIT and after allreduce is done, the state becomes TL_MLX5_TEAM_STATE_ALLTOALL_INIT.  This is why here we are trying to maintain the same way.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: AFAIK, ucc_tl_mlx5_a2a_team_test() no longer has allreduce service call in it. All the sync service calls are just before and after a2a/mcast team create begin and end. Therefore, if we return here, since there is no service call in progress, we essentially will end up in the same line of code in this mlx5 team create test function.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: In this line, we know that both a2a and mcast are not available for this specific process. Therefore, there is nothing else that can be done for it. That is why in this line, we jump to the final stage where we tell all other processes our state.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: Sure, I will make the changes

MamziB:
@janjust @samnordmann Thank you guys for the reviews. I have resolved/answered them. Please let me know if you have more comments.

src/components/tl/mlx5/tl_mlx5.h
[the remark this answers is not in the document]
samnordmann: > This way is easier to maintain.

Why would it be easier to maintain? With what you propose, there are two field of the same type `a2a_state` and `mcast_state`, but we implicitely assume that each field of the enum is applicable to one and not the other.
In my opinion it is misleading and it adds complexity for no benefit.

> I did not want to have two separate sets of enums for mcast and a2a.

Why not? I would at least remove the duplicates in the enum and some of the explicit mention to MCAST and A2A in the enum. In the current version there is unnecessary complexity. For example, why having `TL_MLX5_TEAM_STATE_MCAST_INIT` `TL_MLX5_TEAM_STATE_MCAST_INIT` and `TL_MLX5_TEAM_STATE_INIT` while they are never used at the same time?

src/components/tl/mlx5/tl_mlx5_team.c
samnordmann: But this goto only jumps over this block:
```
if (UCC_OK != a2a_status || UCC_OK != mcast_status) {
    return UCC_INPROGRESS;
}
```
I'm confused: wouldn't we have the two status equal to UCC_OK? what happens if one component is available but not the other ?

src/components/tl/mlx5/tl_mlx5.h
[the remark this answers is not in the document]
MamziB: In this case, we should not have had ucc_tl_mlx5_team_state_t from the first place. Why did we have ucc_tl_mlx5_team_state_t, while only alltoall related states are stored in it? It was not like tl mlx5 was supposed to be all about alltoall. In this patch, I wanted to maintain the idea that was there from the beginning: having the same mlx5 state machine for all the features including a2a and mcast. 

Based on our discussion, let me remove ucc_tl_mlx5_team_state_t completely and then we will have two separate ucc_tl_mlx5_team_a2a_state_t and ucc_tl_mlx5_team_mcast_state_t instead.

src/components/tl/mlx5/tl_mlx5_team.c
samnordmann: but `ucc_tl_mlx5_a2a_team_test` does have service collectives called in it...

src/components/tl/mlx5/tl_mlx5_team.c
samnordmann: So the same was done before?
I don't know, but anyway there seems to be unnecessary complexity here, don't you agree ?

src/components/tl/mlx5/mcast/tl_mlx5_mcast_helper.c
```diff
@@ -401,14 +401,14 @@ ucc_status_t ucc_tl_mlx5_fini_mcast_group(ucc_tl_mlx5_mcast_coll_context_t *ctx,
 
     dst = inet_ntop(AF_INET6, &comm->mcast_addr, buf, 40);
     if (NULL == dst) {
-        tl_error(comm->lib, "inet_ntop failed");
+        tl_warn(comm->lib, "inet_ntop failed");
```
samnordmann, line 404: here also, there is no fallback, right? If yes it should be reported as an actual error.
Please revert this change where an error should be reported (i.e. when the system doesn't internally fall back on another mechanism)

samnordmann (commented):
Thanks!
Can you please ask for a new review through github by clicking "Re-request review"? This would help me. Also, please do not resolve my comments/questions that are not fixed, or that I will probably need to check again, so it will be easier for me to keep track

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: so before this new patch, we had the below code which was making sense, we needed to return UCC_ERR_NO_RESOURCE  immediately and avoid continuing and returning UCC_OK. 
With the new patch, we no longer need this here and this should be moved up where we know global status is not-avvailable for both components. In this case, mlx5 team create must return error.
```
    if (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_NOT_AVAILABLE &&
        tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_NOT_AVAILABLE) {
        tl_error(team->context->lib, "unable to initialize tl team: %p", team);
        return UCC_ERR_NO_RESOURCE;
    }

File: src/components/tl/mlx5/tl_mlx5_team.c (hunk 7 above)
Comment (MamziB) [reply src/components/tl/mlx5/tl_mlx5_team.c:253]: ah, ok. then let me introduce a second tl_team->scoll_req_global_sync just for our allreduce happening outside a2a and mcast.

File: src/components/tl/mlx5/tl_mlx5_team.c (hunk 5 above)
Comment (MamziB) [reply src/components/tl/mlx5/tl_mlx5_team.c:178]: it is indeed complex because we are trying to do team create in a nonblocking manner. If we were free to make it blocking, no state machine was needed. in this specific case, I made some changes. Please let me know how it looks in the new patch.

File: src/components/tl/mlx5/tl_mlx5.h (hunk 6 above)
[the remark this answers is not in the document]
Comment (samnordmann) [reply src/components/tl/mlx5/tl_mlx5.h]: Originally, we had one variable `ucc_tl_mlx5_team_state_t` storing the global state of team creation, tracking both both the common parts and the parts specific to alltoall. This state would be increased linearily and that would indicate the stage where we are at. In this case, it make sense to include in the enum all the possible states, also ones containing alltoall or mcast in it. But now there are two independent variables to track the team creation state....

File: src/components/tl/mlx5/tl_mlx5_team.c (hunk 5 above)
Comment (samnordmann) [reply src/components/tl/mlx5/tl_mlx5_team.c:178]: looks good, thanks

File: src/components/tl/mlx5/tl_mlx5_team.c (hunk 7 above)
Comment (samnordmann) [reply src/components/tl/mlx5/tl_mlx5_team.c:259]: Sorry if I'm confused but I am still seeing `scoll_req` used everywhere. Can you point me to the relevant change?
Also, even if we fix the global sync, I am still afraid there will be conflict between a2a and mcast service calls, am I confused here?
That's why I would still suggest to return at this point to avoid service call overlaps

File: src/components/tl/mlx5/mcast/tl_mlx5_mcast_context.c [hunk 12]
```diff
@@ -296,7 +296,7 @@ ucc_status_t ucc_tl_mlx5_mcast_clean_ctx(ucc_tl_mlx5_mcast_coll_context_t *ctx)
     }
 
     if (ctx->id && rdma_destroy_id(ctx->id)) {
-        tl_error(ctx->lib, "rdma_destroy_id failed errno %d", errno);
+        tl_warn(ctx->lib, "rdma_destroy_id failed errno %d", errno);
```
samnordmann, line 299: this should be error, right ?

MamziB: oh yeah, let me fix it

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: oh I missed to push those changes. Let me push them. 
_"even if we fix the global sync, I am still afraid there will be a conflict between a2a and mcast service calls, am I confused here?"_  During Mcast, we have a Bcast service call and it is using its own collective request (it is not scoll_req). So I think it should not collide with a2a.

src/components/tl/mlx5/tl_mlx5.h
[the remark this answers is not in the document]
MamziB: Yeah, you're right. Now we have two separate state machines, one for a2a and one for mcast. I think since these two components are independent, it makes sense to keep their state machines separate. what do you think?

MamziB:
@samnordmann Thanks for the new comments. I pushed the new changes. Please take a look.

samnordmann (approved):
Thanks!

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,153 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+                    ucc_free(tl_team->mcast);
+                }
+                tl_team->mcast_state = TL_MLX5_TEAM_STATE_MCAST_NOT_AVAILABLE;
+            } else {
+                tl_debug(UCC_TL_TEAM_LIB(tl_team),
+                         "all team members have mcast ctx ready");
+                tl_team->mcast_state = TL_MLX5_TEAM_STATE_MCAST_INIT;
+            }
+
+            return UCC_INPROGRESS;
```
Sergei-Lebedev, line 208: why return? don't you need to continue this function?

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,153 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+initial_sync_post:
+    status = ucc_service_allreduce(
+        core_team, tl_team->local_status_array, tl_team->global_status_array,
+        UCC_DT_INT32, UCC_TL_MLX5_FEATURES_COUNT, UCC_OP_MIN, subset, &tl_team->global_sync_req);
+    if (status < 0) {
+        tl_debug(UCC_TL_TEAM_LIB(tl_team),
+                 "failed to collect global status");
+        return status;
+    }
+    tl_team->global_sync_in_progress = 1;
```
Sergei-Lebedev, line 293: not needed, we usually check if global_sync_req is null

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,153 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+            }
+
+            tl_debug(team->context->lib, "attempted to initialize tl team: %p: MCAST component is %s ALLTOALL component is %s",
+                    team, (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_READY)?"ENABLED":"DISABLED",
+                    (tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_READY)?"ENABLED":"DISABLED");
+        }
+
+        if (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_NOT_AVAILABLE &&
+            tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_NOT_AVAILABLE) {
+            tl_warn(team->context->lib, "unable to initialize tl team as both a2a and mcast are not available: %p", team);
```
Sergei-Lebedev, line 235: please make log messages style identical, sometime it's ALLTOALL or alltoall or a2a

src/components/tl/mlx5/tl_mlx5.h
```diff
@@ -140,7 +146,11 @@ typedef struct ucc_tl_mlx5_team {
     ucc_tl_mlx5_alltoall_t         *a2a;
     ucc_topo_t                     *topo;
     ucc_ep_map_t                    ctx_map;
+    int                             local_mcast_ctx_ready;
     ucc_tl_mlx5_mcast_team_t       *mcast;
+    ucc_status_t                    local_status_array[UCC_TL_MLX5_FEATURES_COUNT];
+    ucc_status_t                    global_status_array[UCC_TL_MLX5_FEATURES_COUNT];
```
Sergei-Lebedev, line 152: imho having bitmask here is better, in that case you don't need to maintain UCC_TL_MLX5_FEATURES_COUNT. But I don't mind having it as array

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,153 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+        (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_READY) ? UCC_OK : UCC_ERR_NO_RESOURCE;
+
+    tl_debug(UCC_TL_TEAM_LIB(tl_team),
+             "posting global status, local status: a2a %d mcast %d",
+             (tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_READY),
+             (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_READY));
+
+initial_sync_post:
+    status = ucc_service_allreduce(
+        core_team, tl_team->local_status_array, tl_team->global_status_array,
```
Sergei-Lebedev, line 286: local_status_array is not actually needed, you can just make inplace allreaduce

src/components/tl/mlx5/tl_mlx5_team.c
```diff
@@ -169,40 +146,153 @@ static inline ucc_status_t ucc_tl_mlx5_a2a_team_test(ucc_base_team_t *team)
...
+        }
+
+        ucc_service_coll_finalize(tl_team->global_sync_req);
+
+        tl_team->global_sync_in_progress = 0;
+
+        if (tl_team->mcast_state == TL_MLX5_TEAM_STATE_MCAST_CTX_CHECK &&
+            tl_team->a2a_state == TL_MLX5_TEAM_STATE_ALLTOALL_CTX_CHECK ) {
+            tl_team->a2a_status.global = tl_team->global_status_array[UCC_TL_MLX5_A2A_STATUS_INDEX];
+            tl_team->a2a_state         = TL_MLX5_TEAM_STATE_ALLTOALL_INIT;
```
Sergei-Lebedev, line 178: why checking only mcast status and not alltoall?

src/components/tl/mlx5/tl_mlx5.h
MamziB: Thanks for the comment.  We will change it in future PRs

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: I agree that it is doable, however, it gives us a clear understanding to know our local status of different features in the same array (even after allreduce is done). This can be used for logging or clearing the local resources. Also, it is more readable and it does not consume too much memory.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: a2a is being handled separately somewhere else. regardless of the global status for a2a, we must set its state to TL_MLX5_TEAM_STATE_ALLTOALL_INIT and let it get handled under a2a calls.

src/components/tl/mlx5/tl_mlx5_team.c
MamziB: i did not want to add a goto to jump from the UCC_OK on line 238. So we return here and take care of the rest of progress in the next team test call

MamziB:
@Sergei-Lebedev Thanks for the comment. I have resolved them and I updated the commit.

Sergei-Lebedev (approved):

PR #954, opened 2024-04-05
EC/CUDA: check ctx valid before resource cleanup

Sergei-Lebedev:
## What
During EC CUDA resource cleanup original CUDA context can be already destroyed. In such case calls to cudaStreamDestroy, cudaEventDestroy etc. might result in segfault. This PR adds check that CUDA context is valid before doing cleanup.

### DIFF
--- a/src/components/ec/cuda/ec_cuda_resources.c
+++ b/src/components/ec/cuda/ec_cuda_resources.c
@@ -180,7 +180,16 @@ void ucc_ec_cuda_resources_cleanup(ucc_ec_cuda_resources_t *resources)
 {
     int i;
     CUcontext tmp_context;
+#if CUDA_VERSION >= 12000
+    CUresult status;
+    unsigned long long int cu_ctx_id;
 
+    status = cuCtxGetId(resources->cu_ctx, &cu_ctx_id);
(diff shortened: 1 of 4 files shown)


src/components/ec/cuda/ec_cuda_resources.c
```diff
@@ -180,7 +180,16 @@ void ucc_ec_cuda_resources_cleanup(ucc_ec_cuda_resources_t *resources)
 {
     int i;
     CUcontext tmp_context;
-
+#if CUDA_VERSION >= 12000
+    CUresult status;
+    unsigned long long int cu_ctx_id;
+
+    status = cuCtxGetId(resources->cu_ctx, &cu_ctx_id);
+    if (ucc_unlikely(status != CUDA_SUCCESS)) {
```
samnordmann, line 188: is `ucc_unlikely` relevant here ? If I understand correctly this is control path right ?

samnordmann (approved):

src/components/ec/cuda/ec_cuda_resources.c
Sergei-Lebedev: not really relevant, will remove it

bureddy (approved):

PR #902, opened 2024-01-23
Sliding Window Allreduce Stubs

nsarka:
This PR (1 out of 2 PRs) defines the structs and function stubs for sliding window allreduce. It also adds a gtest for the algorithm. The next PR will have the function implementations.

@janjust @manjugv @samnordmann @Sergei-Lebedev

PR #958, opened 2024-04-10
TL/UCP: Add Sliding Window allreduce implementation

nsarka:
This PR is the second in a series of two that will add Sliding Window allreduce to UCC. This one implements the function stubs left by the first PR (https://github.com/openucx/ucc/pull/902)


nsarka:
@Sergei-Lebedev It seems like the same ucc test is failing as the active_set. However, this one fails because of wrong cuda versions: `nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.1, please update your driver to a newer version, or use an earlier cuda container: unknown`

janjust:
@artemry-nv Seems like we have CI issues on this PR, unrelated to the PR

artemry-nv:
@B-a-S please take a look

B-a-S:
> @B-a-S please take a look

I've rerun the job. Take a look http://hpc-master.lab.mtl.com:8080/job/ucc/3282/

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+    ucc_assert(host_team_size > 0);
+
+    if (barrier_task != NULL) {
+        // mark sliding window task complete once barrier finishes
+        if (barrier_task->super.status == UCC_OK) {
+            ucc_tl_ucp_put_task(
+                ucc_derived_of(task->allreduce_sliding_window.barrier_task,
+                               ucc_tl_ucp_task_t));
+            task->allreduce_sliding_window.barrier_task = NULL;
+            task->super.status                          = UCC_OK;
```
samnordmann, line 328: IIUC the barrier only occurs at the end of the collective. If this is the case, this can be scheduled during initialization of allreduce (we have the ability to subscribe a task to the completion of another). See for example https://github.com/openucx/ucc/blob/b272589c0a462c3150cf7f6afe69d7d2212e4c30/src/components/tl/mlx5/alltoall/alltoall_coll.c#L876
Does it make sense ?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+
+    UCC_CHECK_GOTO(ucc_tl_ucp_allgather_knomial_init(&bargs,
+                    &team->super.super,
+                    &task->allreduce_sliding_window.allgather_task),
+        out, status);
+
+    task->allreduce_sliding_window.allgather_task->executor =
+                                        task->allreduce_sliding_window.executor;
+
+    UCC_CHECK_GOTO(ucc_tl_ucp_allgather_knomial_start(
```
samnordmann, line 94: same remark as for barrier, IIUC we have
allgather -> allreduce -> barrier
In this case this can be scheduled during init and managed automatically by the core.

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+        task, inplace);
+
+out:
+    ucc_tl_ucp_coll_finalize(allgather_task);
+    task->allreduce_sliding_window.allgather_task = NULL;
 }
 
-static inline void ucc_tl_ucp_allreduce_sliding_window_allgather_free_rkeys(//NOLINT
-    ucc_coll_task_t __attribute__((unused)) *coll_task)//NOLINT
+static inline void ucc_tl_ucp_allreduce_sliding_window_allgather_free_rkeys(
```
samnordmann, line 231: imo the name of the function shouldn't contain the word "allgather", wdyt?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+                 request, UCS_PTR_STATUS(request));
+        return ucs_status_to_ucc_status(UCS_PTR_STATUS(request));
+    } else {
+        return ucs_status_to_ucc_status(ucp_request_check_status(request));
+    }
 }
 
-static inline void ucc_tl_ucp_allreduce_sliding_window_allgather_info_test(//NOLINT
-    ucc_coll_task_t __attribute__((unused)) *coll_task)//NOLINT
+static inline void ucc_tl_ucp_allreduce_sliding_window_allgather_info_test(
```
samnordmann, line 201:
```suggestion
static inline void ucc_tl_ucp_allreduce_sliding_window_allgather_progress(
```

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+                         .mem_type = UCC_MEMORY_TYPE_HOST}
+        }
+    };
+
+    UCC_CHECK_GOTO(ucc_tl_ucp_allgather_knomial_init(&bargs,
+                    &team->super.super,
+                    &task->allreduce_sliding_window.allgather_task),
+        out, status);
+
+    task->allreduce_sliding_window.allgather_task->executor =
```
samnordmann, line 91: why not using `task->executor` directly?

samnordmann (commented):
The implementation looks nice and readable, thanks! I left some comments, most importantly about memory leaks

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
 
 static inline void
-ucc_tl_ucp_allreduce_sliding_window_test_reduction(ucc_tl_ucp_task_t __attribute__((unused)) *task)//NOLINT
+ucc_tl_ucp_allreduce_sliding_window_test_reduction(ucc_tl_ucp_task_t *task)
 {
+    ucc_status_t status;
+
+    #define SAVE_STATE(_reduce_in_progress) _reduce_in_progress = 1
+
+    EXEC_TASK_TEST(task->allreduce_sliding_window.reduce_in_progress,
```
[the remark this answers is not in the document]
nsarka: No, it can only happen with reduce_in_progress=1 because this func is called in two places: 1) right after it is set, and 2) within an if (reduce_in_progress) check.

Maybe we can just check whether etask is null, but what would the SAVE_STATE macro be in that case?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
nsarka: Sergey had the same comment--when I talked with him about it he agreed to accept how it is and we can make a follow-up PR later so that we can make the July release. Is that ok with you as well?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
nsarka: Thank you, same thing as the barrier--https://github.com/openucx/ucc/pull/958#discussion_r1602109352
Please let me know if that's ok.

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
 }
 
 ucc_status_t
-ucc_tl_ucp_allreduce_sliding_window_finalize(ucc_coll_task_t __attribute__((unused)) *coll_task) //NOLINT
+ucc_tl_ucp_allreduce_sliding_window_finalize(ucc_coll_task_t *coll_task)
 {
-    return UCC_OK;
+    ucc_tl_ucp_task_t *task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
+    ucc_status_t       st =
+        ucc_ee_executor_finalize(task->allreduce_sliding_window.executor);
```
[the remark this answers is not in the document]
nsarka: I had created the executor for the allgather task since I found that the allreduce task did not have one. Which function should free it? I don't see how this is freed unless I call `ucc_ee_executor_finalize` myself here.

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
samnordmann: ok, sounds good!

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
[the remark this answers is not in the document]
samnordmann: > Which function should free it? I don't see how this is freed unless I call `ucc_ee_executor_finalize` myself here.

I was thinking about this

https://github.com/openucx/ucc/blob/c65c4a3fb1732799b1498ce97b148f0acfe490bd/src/schedule/ucc_schedule.h#L212

> I had created the executor for the allgather task since I found that the allreduce task did not have one

For creation this should be handled here

https://github.com/openucx/ucc/blob/c65c4a3fb1732799b1498ce97b148f0acfe490bd/src/core/ucc_coll.c#L266

provided that `task->flags` has `UCC_COLL_TASK_FLAG_EXECUTOR`

Does it make sense?
@Sergei-Lebedev could you confirm that this is should be the workflow for the executor ?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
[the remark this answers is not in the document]
samnordmann: > Maybe we can just check whether etask is null, but what would the SAVE_STATE macro be in that case?

I fail to see how it is related, since, IIUC, `SAVE_STATE` has no effect (it reassigns `reduce_in_progress` to 1, but it already equaled this value). Checking through `etask==NULL` allows to save one variable, and it is usually how it is done in the codebase if I remember correctly

I am not sure to understand what `SAVE_STATE` is useful for. Could you explain?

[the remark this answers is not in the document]
nsarka: You're definitely right about `SAVE_STATE` and `etask`. I changed `etask` to `reduce_task` and used that instead of `reduce_in_progress`. Thank you, it's much better this way!

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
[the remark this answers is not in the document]
nsarka: After setting the flag, I confirmed it works by allocating in `ucc_collective_init` and freeing in `ucc_collective_finalize_internal`. Thank you Sam!!

nsarka:
The ucc gtest failed on
```
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/4
[2024-05-16T18:13:52.173Z] [swx-clx01:196  :0:196] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))
```

nsarka:
On this PR, I logged into the CI node and opened a shell inside the container that was running this gtest. Running it myself it passed:

```
swx-jenkins@swx-clx01:/opt/nvidia/src/ucc/build/test/gtest$ stdbuf -e0 -o0 /opt/nvidia/src/ucc/build/test/gtest/gtest --gtest_filter=*test_tl_mlx5_dm.MemcpyToDeviceMemory*
Note: Google Test filter = *test_tl_mlx5_dm.MemcpyToDeviceMemory*
[==========] Running 8 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 8 tests from test_tl_mlx5_dm
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/0
[  SKIPPED ] test_tl_mlx5_dm.MemcpyToDeviceMemory/0 (63 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/1
[       OK ] test_tl_mlx5_dm.MemcpyToDeviceMemory/1 (36 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/2
[  SKIPPED ] test_tl_mlx5_dm.MemcpyToDeviceMemory/2 (35 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/3
[       OK ] test_tl_mlx5_dm.MemcpyToDeviceMemory/3 (35 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/4
[       OK ] test_tl_mlx5_dm.MemcpyToDeviceMemory/4 (38 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/5
[  SKIPPED ] test_tl_mlx5_dm.MemcpyToDeviceMemory/5 (37 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/6
[       OK ] test_tl_mlx5_dm.MemcpyToDeviceMemory/6 (48 ms)
[ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/7
[       OK ] test_tl_mlx5_dm.MemcpyToDeviceMemory/7 (59 ms)
[----------] 8 tests from test_tl_mlx5_dm (351 ms total)

[----------] Global test environment tear-down
[==========] 8 tests from 1 test suite ran. (351 ms total)
[  PASSED  ] 5 tests.
[  SKIPPED ] 3 tests.
```

janjust (approved):
I went over this PR multiple times with Nick.
LGTM

nsarka:
I updated the PR. The get/reduce/put phase and the barrier part of the algorithm are now run via schedule. I left the allgather phase the way it was inside of the get/reduce/put phase because once Ferrol's PR goes in, I will be removing the allgather and using that instead for key exchange.

nsarka:
Please wait to review, there are some failures I should fix first.

artemry-nv:
bot:retest

nsarka:
@samnordmann The PR is ready for review, thank you

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -10,34 +10,277 @@
...
+    return 0;
+}
+
 ucc_status_t
-ucc_tl_ucp_allreduce_sliding_window_alloc_pipe(
-    ucc_base_coll_args_t __attribute__((unused)) *coll_args,//NOLINT
-    ucc_base_team_t __attribute__((unused)) *team,//NOLINT
-    ucc_tl_ucp_task_t __attribute__((unused)) *task)//NOLINT
+ucc_tl_ucp_allreduce_sliding_window_alloc_pipe(ucc_base_team_t      *team,
+                                               ucc_tl_ucp_task_t    *task)
```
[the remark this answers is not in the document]
samnordmann: wdyt about this?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+    }
+
+    ucc_tl_ucp_task_reset(task, UCC_INPROGRESS);
+
+    task->allreduce_sliding_window.reduce_in_progress = 0;
+    task->allreduce_sliding_window.barrier_task       = NULL;
+
+    return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+
+out:
```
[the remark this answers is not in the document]
samnordmann: wdyt about this ?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,505 @@
...
+
+    // Register the dst buf
+    status = ucc_tl_ucp_allreduce_sliding_window_register(
+        tl_ctx->worker.ucp_context, team,
+        rdma_task->allreduce_sliding_window.bufs->dst_ebuf,
+        gwbi_p->packed_dst_memh);
+    if (status != UCC_OK) {
+        tl_error(UCC_TASK_LIB(rdma_task), "failed to register dst memh: %s",
+                    ucc_status_string(status));
+        goto out;
```
samnordmann, line 103: don't we need to unregister the src buffer here?

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -11,33 +11,289 @@
...
+
+    task->allreduce_sliding_window.pipe = pipe;
+    task->allreduce_sliding_window.put_requests =
+        task->allreduce_sliding_window.pipe->put_requests;
+
     return UCC_OK;
+
+free_getbuf:
+    for (i = 0; i < num_get_bufs; i++) {
+        if (pipe->getbuf[i].buf == NULL)
```
samnordmann, line 79: codestyle: brackets

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -11,33 +11,289 @@
...
-    return UCC_OK;
+    ucs_status_t       ucs_status = UCS_OK;
+    ucc_base_team_t   *base_team  = sw_task->super.team;
+    ucc_tl_ucp_team_t *tl_team    = ucc_derived_of(base_team, ucc_tl_ucp_team_t);
+    ucc_rank_t         team_size  = base_team->params.size;
+    void              *recvbuf    = sw_task->allreduce_sliding_window.
+                                    allgather_task->bargs.args.dst.info.buffer;
+    ucc_tl_ucp_allreduce_sw_host_allgather_t *all_host_allgather = recvbuf;
+    ucc_status_t       status     = UCC_OK;
+    int                inplace    = UCC_IS_INPLACE(sw_task->super.bargs.args);
```
samnordmann, line 203:
```suggestion
    int                inplace    = UCC_IS_INPLACE(TASK_ARGS(sw_task));
```

samnordmann (approved):
LGTM! Thanks for addressing the comments. I still have left a couple of new minor comments

janjust (approved):

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
nsarka: If it fails here, then the schedule core will eventually call `ucc_tl_ucp_allreduce_sliding_window_rdma_task_finalize` which will unregister these

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+
+    pipe->my_count  = count_total / size;
+    pipe->my_offset = pipe->my_count * dt_size * rank;
+    if (rank == size - 1) {
+        pipe->my_count += count_total % size;
+    }
+
+    rdma_task->allreduce_sliding_window.reduce_task = NULL;
+
+    UCC_CHECK_GOTO(ucc_tl_ucp_allgather_ring_start(
```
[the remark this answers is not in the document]
nsarka: I left the allgather outside of the schedule because when Ferrol's PR is merged, I will remove it and use his code instead

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -11,33 +11,302 @@
...
+    void                 *dst_buf   = coll_args->args.dst.info.buffer;
+    ucc_tl_ucp_team_t    *tl_team   = ucc_derived_of(team, ucc_tl_ucp_team_t);
+    ucc_rank_t            team_size = UCC_TL_TEAM_SIZE(tl_team);
+    int                   inplace   = UCC_IS_INPLACE(coll_args->args);
+    ucc_tl_ucp_allreduce_sw_global_work_buf_info_t *gwbi_p = NULL;
+    size_t allgather_size = sizeof(ucc_tl_ucp_allreduce_sw_host_allgather_t);
+    ucc_tl_ucp_allreduce_sw_host_allgather_t *allgather_data;
+    ucc_rank_t            i;
+
+    ucc_assert(team_size > 0);
```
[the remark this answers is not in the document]
nsarka: the linter fails without this assertion even though it's clear that it will never be <= 0

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -11,33 +11,302 @@
...
+
+    task->allreduce_sliding_window.reduce_task = NULL;
+
+    task->allreduce_sliding_window.bufs->rbufs =
+        ucc_malloc(sizeof(void *) * team_size);
+    if (task->allreduce_sliding_window.bufs->rbufs == NULL) {
+        goto free_allgather_data;
+    }
+    task->allreduce_sliding_window.bufs->dst_rkeys =
+        ucc_malloc(sizeof(ucp_rkey_h) * team_size);
```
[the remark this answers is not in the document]
nsarka: i thought it would be simpler to have one malloc per buffer

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,507 @@
...
+                                                  ucc_tl_ucp_task_t);
+    ucc_tl_ucp_allreduce_sw_pipeline_t       *pipe;
+    ucc_tl_ucp_allreduce_sw_host_allgather_t *allgather_data;
+
+    pipe           = rdma_task->allreduce_sliding_window.pipe;
+    allgather_data = rdma_task->allreduce_sliding_window.allgather_data;
+
+    // Register the src buf
+    if (!inplace) {
+        status = ucc_tl_ucp_allreduce_sliding_window_register(
```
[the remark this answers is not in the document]
nsarka: for this i measured the difference between registering in init and registering in start. because init is called in serial by the main thread, the register would also get serialized. registering in parallel showed to be much faster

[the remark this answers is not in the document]
Sergei-Lebedev: Can you please explain how it works with multiple threads? I don't understand how you can do start in parallel and init sequential

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,515 @@
...
+
+        if (accbuf->state == RECVING) {
+            request = accbuf->ucp_req;
+            if (ucc_tl_ucp_allreduce_sliding_window_req_test(request, task) ==
+                UCC_OK) {
+                if (request)
+                    ucp_request_free(request);
+                accbuf->state   = REDUCING;
+                accbuf->ucp_req = NULL;
+            }
```
Sergei-Lebedev, line 405: i would recommend to use callbacks for this, you will be able to complete multiple requests in ucp_worker_progress and ucp_req object is actually not needed

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,515 @@
...
+                pipe->count_serviced += accbuf->count;
+
+                ucc_tl_ucp_allreduce_sliding_window_reset_buf(accbuf);
+                pipe->done_get = 0;
+                pipe->done_red = pipe->done_put = pipe->posted_put = 0;
+
+                for (i = 0; i < put_window_size;
+                     i++) {
+                    task->allreduce_sliding_window.put_requests[i] = NULL;
+                }
```
Sergei-Lebedev, line 515: we usually use memset for such things

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -11,33 +11,304 @@
...
+    int      num_get_bufs = cfg->allreduce_sliding_window_num_get_bufs;
+
+    ucc_tl_ucp_allreduce_sw_pipeline *pipe =
+        (ucc_tl_ucp_allreduce_sw_pipeline *)ucc_malloc(
+            sizeof(ucc_tl_ucp_allreduce_sw_pipeline));
+    if (pipe == NULL) {
+        goto err;
+    }
+
+    if (put_window_size <= 0) {
```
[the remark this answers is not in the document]
nsarka: These fields I noted in the config description that they can be set to 0 if their values should be set to the team size. The user can throttle the puts and gets by setting it to a value lower than the team size if they want to

src/components/tl/ucp/allreduce/allreduce.c
```diff
@@ -57,11 +63,121 @@ ucc_status_t ucc_tl_ucp_allreduce_knomial_init(ucc_base_coll_args_t *coll_args,
...
+    status = ucc_tl_ucp_allreduce_sliding_window_task_init(coll_args, team,
+                                                           rdma_task);
+    if (status != UCC_OK) {
+        ucc_error("failed to init task: %s", ucc_status_string(status));
+        goto out;
+    }
+
+    allgather_data = rdma_task->allreduce_sliding_window.allgather_data;
+    bargs.args.src.info.buffer = allgather_data;
+    bargs.args.dst.info.buffer = PTR_OFFSET(allgather_data, allgather_size);
```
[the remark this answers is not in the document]
nsarka: I am using ring because knomial fails, and I measured the difference to be ~10 us. At the target message size range (over 1 MB) this difference about the same as you would see in system noise. Is it ok if I leave it?

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
nsarka: this is a good optimization, but to meet the july deadline we agreed it would be ok to leave out for now

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,631 @@
...
+    ucc_base_coll_args_t *coll_args       = &coll_task->bargs;
+    ucc_schedule_t       *schedule        = ucc_derived_of(coll_task,
+                                                       ucc_schedule_t);
+    ucc_base_team_t      *base_team       = schedule->super.team;
+    ucc_tl_ucp_team_t    *team            = ucc_derived_of(base_team,
+                                                       ucc_tl_ucp_team_t);
+    ucc_tl_ucp_context_t *tl_ctx          = UCC_TL_UCP_TEAM_CTX(team);
+    ucc_rank_t            rank            = UCC_TL_TEAM_RANK(team);
+    uint32_t              count_total     = coll_task->bargs.args.dst.info.count;
+    ucc_rank_t            size            = coll_task->team->params.size;
```
Sergei-Lebedev, line 62: TL_TEAM_SIZE

src/components/tl/ucp/tl_ucp_dpu_offload.h
```diff
@@ -0,0 +1,55 @@
...
+
+struct ucc_tl_ucp_allreduce_sw_export_buf {
+    ucp_context_h ucp_context;
+    ucp_mem_h     memh;
+    void         *packed_memh;
+    size_t        packed_memh_len;
+    void         *packed_key;
+    size_t        packed_key_len;
+    uint64_t      memh_id;
+};
```
Sergei-Lebedev, line 32: are memh_id and packed_mem_len unused?

src/components/tl/ucp/allreduce/allreduce_sliding_window_setup.c
```diff
@@ -11,33 +11,280 @@
...
-   ucc_service_coll_req_t __attribute__((unused)) *scoll_req, //NOLINT
-   ucc_tl_ucp_task_t __attribute__((unused)) *sw_task)//NOLINT
+ucc_status_t ucc_tl_ucp_allreduce_sliding_window_allgather_info_finalize(
+                ucc_tl_ucp_task_t *sw_task)
 {
-    return UCC_OK;
+    ucs_status_t       ucs_status = UCS_OK;
+    ucc_base_team_t   *base_team  = sw_task->super.team;
+    ucc_tl_ucp_team_t *tl_team    = ucc_derived_of(base_team, ucc_tl_ucp_team_t);
+    ucc_rank_t         team_size  = base_team->params.size;
```
Sergei-Lebedev, line 190: TL_TEAM_SIZE

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,631 @@
...
+
+            pipe->done_get++;
+            if (pipe->done_get == host_team_size) {
+                pipe->count_received += count;
+            }
+        }
+
+        if (accbuf->state == RECVING) {
+            request = accbuf->ucp_req;
+            if (ucc_tl_ucp_allreduce_sliding_window_req_test(request, task) ==
```
Sergei-Lebedev, line 399: check for error

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
```diff
@@ -11,80 +11,631 @@
...
+                ucc_assert(pipe->done_red == host_team_size - 1);
+                ucc_assert(pipe->done_put == host_team_size);
+
+                pipe->count_serviced += accbuf->count;
+
+                ucc_tl_ucp_allreduce_sliding_window_reset_buf(accbuf);
+                pipe->done_get = 0;
+                pipe->done_red = pipe->done_put = pipe->posted_put = 0;
+
+                memset(task->allreduce_sliding_window.put_requests, 0,
```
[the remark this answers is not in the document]
nsarka: Actually here it is required since at the start of the algorithm they need to be initialized to NULL. The other memset in the algorithm was not required. I removed it.

Sergei-Lebedev (approved):
follow up changes:
1. Use single allocation instead of multiple mallocs
2. Get rid of explicit ucp request objects and use callbacks
3. Optimize ucp_worker calls

src/components/tl/ucp/allreduce/allreduce_sliding_window.c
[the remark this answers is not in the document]
nsarka: im not sure where my comment went, but the reason this is not needed is if we reach here then the status should be an error, then the schedule core will call finalize which will free and unregister allocations from task init

nsarka:
@Sergei-Lebedev @samnordmann The PR is ready to be merged

PR #966, opened 2024-04-23
TL/MLX5: Disable mlx5 team by default

janjust:
This PR increases the default min size supported for tl_mlx5 to uint32_max thus disabling tl_mlx5 team unless the users increase the min size via env. variable

We have noticed several corner-case bugs in tl_mlx5 algorithms, therefore until the features mature we disable tl_mlx5 when the team is built rather than not building the team at all.

### DIFF
--- a/src/components/tl/mlx5/tl_mlx5_lib.c
+++ b/src/components/tl/mlx5/tl_mlx5_lib.c
@@ -48,7 +48,7 @@ ucc_status_t ucc_tl_mlx5_get_lib_attr(const ucc_base_lib_t *lib, /* NOLINT */
 
 ucc_status_t ucc_tl_mlx5_get_lib_properties(ucc_base_lib_properties_t *prop)
 {
-    prop->default_team_size = 2;
+    prop->default_team_size = UCC_RANK_MAX;
     prop->min_team_size     = 2;
     prop->max_team_size     = UCC_RANK_MAX;
     return UCC_OK;


janjust:
Without setting `-x UCC_TL_MLX5_MIN_TEAM_SIZE= `
`[1713901146.159619] [cascade1:17008:0]          ucc_tl.c:293  TL_MLX5 DEBUG team size 16 is too small, min supported -2`

And when a user wants to enable tl_mlx5, he needs to increase it's min size as well.
eg: `-x UCC_TL_MLX5_MIN_TEAM_SIZE=2 `
`[1713901217.005463] [cascade1:17191:0]    tl_mlx5_team.c:99   TL_MLX5 DEBUG finalizing tl team: 0x329b700`

Sergei-Lebedev (approved):
@janjust  can you please fix log msg to print team size correctly
```diff
--- a/src/components/tl/ucc_tl.c
+++ b/src/components/tl/ucc_tl.c
@@ -289,13 +289,13 @@ UCC_CLASS_INIT_FUNC(ucc_tl_team_t, ucc_tl_context_t *tl_context,
     }

     if (attr.min_team_size > params->size) {
-        tl_debug(lib, "team size %d is too small, min supported %d",
+        tl_debug(lib, "team size %ul is too small, min supported %ul",
                  params->size, attr.min_team_size);
         return UCC_ERR_NOT_SUPPORTED;
     }

     if (attr.max_team_size < params->size) {
-        tl_debug(lib, "team size %d is too big, max supported %d",
+        tl_debug(lib, "team size %ul is too big, max supported %ul",
                  params->size, attr.max_team_size);
         return UCC_ERR_NOT_SUPPORTED;
     }
```

samnordmann:
Why not using a flag to deactivate the tl then? "Min team size" has another meaning and using it for that purpose sounds like a hack.
We should be able to deactivate the tl at the core level.

Since `MCAST_ENABLE` is disabled by default, should I understand that the problem comes from a2a team creation?

janjust:
Decided we'll do a different approach, closing this.

PR #991, opened 2024-06-24
UTIL: fix pipeline params check for auto

Sergei-Lebedev:
## What
Check for auto pipeline params by comparing fields of pipeline params struct

## Why ?
memset doesn't work due to possible uninitialized paddings between structure fields


samnordmann (approved):
fix copyright

bureddy (approved):

PR #1000, opened 2024-07-12
Asymmetric memory

nsarka:
In our UCC call, we defined two types of "asymmetric memory":

- Weak asymmetric memory: When the src and dst mem_type's mismatch, but the mismatch is the same across ranks
- Strong asymmetric memory: No rules for mem types at all, even across ranks. Src/dst mem_type's can mismatch on one rank but match on another

Since strong asymmetric memory would require an allreduce for every `ucc_collective_init`, this PR implements a solution for weak asymmetric memory. **Now the question is, what about for rooted collectives, where only the root is passed both a src and a dst buffer, and non-root ranks are only passed just a src OR a dst buffer?**

In this case, since we're dealing with weak asymmetric memory we can assume that the src buffer is the "true" mem_type, and make a scratch allocation for the dst buffer of size `dst.info.count * dt_size(dst.info.mem_type)` with mem_type `src.info.mem_type`. Then, run the collective. After the collective completes, copy out from the scratch allocation into the old dst buffer.


src/coll_score/ucc_coll_score_map.c
```diff
@@ -87,15 +87,16 @@ static ucc_status_t ucc_coll_score_map_lookup(ucc_score_map_t *map,
     ucc_list_link_t *list;
     ucc_msg_range_t *r;
 
-    if (mt == UCC_MEMORY_TYPE_ASYMMETRIC) {
```
nsarka, line 90: Should this still error for non-rooted collectives? Or should I update the PR to copy-out for those too?

samnordmann:
Is it ready for review? If so, please add the label

nsarka:
It is ready for review. I’m OOO, so I can’t add the label. Please let me know your opinion though on my question about supporting non-rooted collectivesOn Jul 18, 2024, at 7:27 AM, samnordmann ***@***.***> wrote:﻿
Is it ready for review? If so, please add the label

samnordmann (commented):
Thanks! It looks great overall! I just left a bunch of very minor comments. A couple of further question/notes:
1. The CI error looks relevant
2. Is there a concrete motivation for this patch? I mean, is it a user's request, or do we only think it is a nice feature to have?
3. Can we hope (in the future) to have a more efficient way to handle asymmetric memory, or does it need to be performed, as here, by emulating symmetric memory with a scratch buffer and copying in/out to the user's buffer?

test/gtest/asym_mem/test_asymmetric_memory.cc
```diff
@@ -0,0 +1,417 @@
...
+                ucc_free(rst);
+            }
+        }
+        
+        return ret;
+    }
+};
+
+
+class test_asymmetric_memory_v : public ucc::test,
```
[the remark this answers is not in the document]
nsarka: it's more than that--the buffer allocation, data init, data validate, and etc. all are different because of info_v and because it uses counts/displacements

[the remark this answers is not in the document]
nsarka: i think it can be merged into one, but i think it would be cleaner to have it separate just so that there isn't an extra layer of if/else for allocation/init/validation/free

nsarka:
> 1. The CI error looks relevant

Thanks, I will take a look

> 2. For my own curiosity: Is there a concrete motivation for this patch? I mean, is it a user's request, or do we only think it is a nice feature to have?

A hang was reported by the HPC SDK team (I think that's the name, I'm not sure, I heard it from Tommy) in the case of asymmetric memory in a rooted collective. Consider this case:
non-root ranks are UCC_MEMORY_TYPE_HOST, root rank is src=HOST dst=CUDA. Before this patch, the root will exit because the memory types are asymmetric. However, non-roots do not know that the root has exited. So there is a hang.

> 3. Can we hope (in the future) to have a more efficient way to handle asymmetric memory, or does it need to be performed, as here, by emulating symmetric memory with a scratch buffer and copying in/out to the user's buffer?

I believe ucp can handle mismatched memory without any problems which is why I was wondering about implementing a copy-out mechanism like this for non-rooted collectives. I will test it and report back. For rooted collectives, I'm not sure yet if there's a better way.

test/gtest/asym_mem/test_asymmetric_memory.cc
```diff
@@ -0,0 +1,417 @@
...
+                continue;
+            }
+
+            root = ctxs[i]->args->root;
+
+            if (i == root) {
+                dst_mem_type = ctxs[root]->args->dst.info.mem_type;
+
+                rst = (uint8_t*) ucc_malloc(ctxs[root]->rbuf_size, "validation buf");
+                EXPECT_NE(rst, nullptr);
```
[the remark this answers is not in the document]
nsarka: It seems like this change causes a linter error:

```
[2024-07-23T18:07:20.418Z] [91mIn file included from /opt/nvidia/src/ucc/test/gtest/common/test.h:17,
[2024-07-23T18:07:20.418Z]                  from /opt/nvidia/src/ucc/test/gtest/common/test_ucc.h:8,
[2024-07-23T18:07:20.418Z]                  from /opt/nvidia/src/ucc/test/gtest/asym_mem/test_asymmetric_memory.cc:6:
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/asym_mem/test_asymmetric_memory.cc: In member function ‘bool test_asymmetric_memory::data_validate()’:
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:7597:5: error: void value not ignored as it ought to be
[2024-07-23T18:07:20.418Z]  7596 |   ::testing::internal::AssertHelper(result_type, file, line, message) \
[2024-07-23T18:07:20.418Z]       |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z]  7597 |     = ::testing::Message()
[2024-07-23T18:07:20.418Z]       |     ^~~~~~~~~~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:7600:3: note: in expansion of macro ‘GTEST_MESSAGE_AT_’
[2024-07-23T18:07:20.418Z]  7600 |   GTEST_MESSAGE_AT_(__FILE__, __LINE__, message, result_type)
[2024-07-23T18:07:20.418Z]       |   ^~~~~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:7603:10: note: in expansion of macro ‘GTEST_MESSAGE_’
[2024-07-23T18:07:20.418Z]  7603 |   return GTEST_MESSAGE_(message, ::testing::TestPartResult::kFatalFailure)
[2024-07-23T18:07:20.418Z]       |          ^~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:12435:5: note: in expansion of macro ‘GTEST_FATAL_FAILURE_’
[2024-07-23T18:07:20.418Z] 12435 |     on_failure(gtest_ar.failure_message())
[2024-07-23T18:07:20.418Z]       |     ^~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:12504:3: note: in expansion of macro ‘GTEST_ASSERT_’
[2024-07-23T18:07:20.418Z] 12504 |   GTEST_ASSERT_(pred_format(#v1, #v2, v1, v2), \
[2024-07-23T18:07:20.418Z]       |   ^~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:12523:3: note: in expansion of macro ‘GTEST_PRED_FORMAT2_’
[2024-07-23T18:07:20.418Z] 12523 |   GTEST_PRED_FORMAT2_(pred_format, v1, v2, GTEST_FATAL_FAILURE_)
[2024-07-23T18:07:20.418Z]       |   ^~~~~~~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:14382:3: note: in expansion of macro ‘ASSERT_PRED_FORMAT2’
[2024-07-23T18:07:20.418Z] 14382 |   ASSERT_PRED_FORMAT2(::testing::internal::CmpHelperNE, val1, val2)
[2024-07-23T18:07:20.418Z]       |   ^~~~~~~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/common/gtest.h:14400:32: note: in expansion of macro ‘GTEST_ASSERT_NE’
[2024-07-23T18:07:20.418Z] 14400 | # define ASSERT_NE(val1, val2) GTEST_ASSERT_NE(val1, val2)
[2024-07-23T18:07:20.418Z]       |                                ^~~~~~~~~~~~~~~
[2024-07-23T18:07:20.418Z] /opt/nvidia/src/ucc/test/gtest/asym_mem/test_asymmetric_memory.cc:116:17: note: in expansion of macro ‘ASSERT_NE’
[2024-07-23T18:07:20.418Z]   116 |                 ASSERT_NE(rst, nullptr);
[2024-07-23T18:07:20.418Z]       |                 ^~~~~~~~~
[2024-07-23T18:07:20.986Z] [0m[91mmake[1]: *** [Makefile:1558: asym_mem/gtest-test_asymmetric_memory.o] Error 1
```

samnordmann (approved):

src/utils/ucc_coll_utils.c
```diff
@@ -94,6 +94,68 @@ ucc_coll_args_is_mem_symmetric(const ucc_coll_args_t *args,
...
+{
+    ucc_status_t status = UCC_OK;
+
+    if (UCC_IS_INPLACE(*args)) {
+        return UCC_ERR_INVALID_PARAM;
+    }
+    switch (args->coll_type) {
+    case UCC_COLL_TYPE_REDUCE:
+    case UCC_COLL_TYPE_GATHER:
+    case UCC_COLL_TYPE_SCATTER:
```
Sergei-Lebedev, line 114: why can we combine gather and scatter here? if memory type is asymmetric then
for scatter:
src buffer valid only for root, we should use dst buffer memory type and replace src buffer with scratch 
for gather:
dst buffer valid only for root, we should use src buffer memory type and replace dst buffer with scratch

src/utils/ucc_coll_utils.h
```diff
@@ -335,4 +335,16 @@ static inline size_t ucc_buffer_block_offset_aligned(size_t total_count,
                            operations. */
 int ucc_coll_args_is_predefined_dt(const ucc_coll_args_t *args, ucc_rank_t rank);
 
+int ucc_coll_args_is_mem_symmetric(const ucc_coll_args_t *args, ucc_rank_t rank);
+
+int ucc_coll_args_is_rooted(ucc_coll_type_t ct);
+
+typedef struct ucc_buffer_info_asymmetric_memtype ucc_buffer_info_asymmetric_memtype_t;
+typedef struct ucc_mc_buffer_header ucc_mc_buffer_header_t;
```
Sergei-Lebedev, line 343: pls include ucc_mc.h instead

src/utils/ucc_coll_utils.c
```diff
@@ -94,6 +94,68 @@ ucc_coll_args_is_mem_symmetric(const ucc_coll_args_t *args,
     return 0;
 }
 
+
+/* If this is the root and the src/dst buffers are asymmetric, the dst needs
```
Sergei-Lebedev, line 98: I'm not sure this what we agreed on, if I remember correctly new memory type should match memory type of non root rank buffer. On task completion copy back is not always needed, sometimes we need to copy src buffer but keep original dst buffer

src/utils/ucc_coll_utils.h
nsarka: If i include "components/mc/ucc_mc.h" then there is a circular header dependency inside ucc_parser.h that generates an error ` error: unknown type name ‘ucc_config_names_array_t’`, which is typedef'd inside ucc_parser.h after the dependent header. Is it worth fixing this in order to make this change?

nsarka:
Hi @Sergei-Lebedev , I have addressed all of your comments. For scatter and scatterv, a new src buf will be allocated and copied into. To support this as a persistent collective I moved the copy-in to ucc_collective_post. Then I added a persistent gtest that will:
- init with a buffer
- In a loop
     - update the src buffer with a different value
     - ucc_collective_post
     - data validate
- finalize

nsarka:
CI is failing with: `failed to register layer: write /opt/nvidia/hpc_sdk/Linux_x86_64/23.9/profilers/Nsight_Systems/target-linux-x64/libcupti.so.11.3: no space left on device
  Error: Docker pull failed with exit code 1`

nsarka:
bot:retest

src/utils/ucc_coll_utils.c
```diff
@@ -94,6 +94,172 @@ ucc_coll_args_is_mem_symmetric(const ucc_coll_args_t *args,
...
+        // copy in
+        status = ucc_mc_memcpy(save->scratch->addr,
+                        save->old_asymmetric_buffer.info.buffer,
+                        ucc_dt_size(save->old_asymmetric_buffer.info.datatype) *
+                            save->old_asymmetric_buffer.info.count,
+                        save->scratch->mt,
+                        save->old_asymmetric_buffer.info.mem_type);
+    } else if(task->bargs.args.coll_type == UCC_COLL_TYPE_GATHERV) {
+        // copy out
+        status = ucc_mc_memcpy(save->old_asymmetric_buffer.info_v.buffer,
```
Sergei-Lebedev, line 241: here we assume that destination buffer is contig, however I don't think it's always true. For example MPI spec doesn't say anything about displacements so I think it general case it won't work.
Example
```
#rank0
args.root = 0
args.src.info.count = 1
args.dst.info_v.counts= [1, 1]
args.dst.info_v.displacements = [1, 5]

#rank1:
args.root = 0
args.src.info.count = 1
```

nsarka: Hi Sergey, I updated this so that it can handle this case. My only remaining question is, is it ok that we copy the blank space in between `info_v.buf + displacements[0]` and `info_v.buf + displacements[1]`? The way I updated this PR will handle it by copying this memory.

I am assuming this works because the MPI standard defines these displacements as an offset from the buffer. The standard implies a negative stride is invalid, and also, every example allocates this empty space.

Sergei-Lebedev (approved):

Sergei-Lebedev:
@nsarka LGTM, please fix commit messages to pass codestyle check

janjust (approved):

nsarka:
It seems like the CI is failing on:
```
[2024-09-03T23:44:41.401Z] [ RUN      ] test_tl_mlx5_dm.MemcpyToDeviceMemory/0
[2024-09-03T23:44:41.401Z] [swx-clx01:452  :0:452] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))
```

PR #1001, opened 2024-07-16
TL/SHARP: fix service team oob

Sergei-Lebedev:
## What
In case of global service team we need to have an additional mapping from core team to core context to choose correct endpoint in service algorithms

## Why ?
Fixes hang in TL SHARP team create when service team is used for OOB


src/components/tl/sharp/tl_sharp_team.c
```diff
@@ -36,7 +37,10 @@ UCC_CLASS_INIT_FUNC(ucc_tl_sharp_team_t, ucc_base_context_t *tl_context,
     set.map    = UCC_TL_TEAM_MAP(self);
 
     if (UCC_TL_SHARP_TEAM_LIB(self)->cfg.use_internal_oob) {
-        self->oob_ctx.subset = set;
+        ucc_ep_map_create_nested(&UCC_TL_CORE_TEAM(self)->ctx_map,
```
samnordmann, line 40: The map needs to be cleaned in the error path of the current function

src/components/tl/sharp/tl_sharp_team.c
```diff
@@ -36,7 +37,10 @@ UCC_CLASS_INIT_FUNC(ucc_tl_sharp_team_t, ucc_base_context_t *tl_context,
...
+                                 &UCC_TL_TEAM_MAP(self),
+                                 &self->oob_ctx.subset.map);
+        self->oob_ctx.subset.myrank = UCC_TL_TEAM_RANK(self);
     } else {
         self->oob_ctx.oob = &UCC_TL_TEAM_OOB(self);
     }
```
samnordmann, line 46: only to make sure: is it normal to use below `set` and not `self->oob_ctx.subset` inside `topo_init`?

janjust (approved):
lgtm, though I'm echoing Sam's comment

src/components/tl/sharp/tl_sharp_team.c
Sergei-Lebedev: yes, we need tl team to core team map in topo_init

src/components/tl/sharp/tl_sharp_team.c
Sergei-Lebedev: fixed

bureddy (approved):

samnordmann (approved):

PR #1006, opened 2024-08-02
Replace bashisms with POSIX syntax

jeroen-mostert:
This replaces some bashisms (notably equality check with `==` and substring replacement) with POSIX equivalents, for distros where `sh` operates in strict POSIX mode. I will not claim that all bashisms are thereby purged from the configure files (they may lurk in conditionals not taken), but at least it allows a basic build to proceed.


manjugv:
@jeroen-mostert  What is your affiliation? Thanks.

jeroen-mostert:
This patch is a product of my spare time, neither commissioned by, nor approved of, nor sponsored by, nor produced during working hours, of the company I also happen to be employed for. I hope that answers your question.

PR #1021, opened 2024-09-19
TL/UCP: Allow self copy in allgather using network loopback

yaeliyac:
## What
Add option for self copy loopback in out of place start for allgather algorithms: knomial, ring, neighbor, sparbit 

## Why ?
When using UCC plugin for NCCL, using cuda_memcpy might cause deadlock

## How ?
Add UCC_TL_UCP_ALLGATHER_USE_LOOPBACK flag to control this
Tested on ISRAEL-1 with multiple test cases - showed good results and same performance with and without loopback
image

### DIFF
--- a/src/components/tl/ucp/allgather/allgather.c
+++ b/src/components/tl/ucp/allgather/allgather.c
@@ -58,3 +58,37 @@ char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
                       UCC_TL_UCP_ALLGATHER_DEFAULT_ALG_SELECT_STR, algo_num);
     return str;
 }
+
+ucc_status_t loopback_self_copy(void *rbuf, void *sbuf, size_t data_size,
+                                ucc_memory_type_t rmem, ucc_memory_type_t smem,
+                                ucc_rank_t rank, ucc_tl_ucp_team_t *team,
+                                ucc_tl_ucp_task_t *task)
+{
+    ucc_status_t status;
+    status = ucc_tl_ucp_send_nb(sbuf, data_size, smem, rank, team, task);
+    if (UCC_OK != status) {
+        task->super.status = status;
+        return task->super.status;
+    }
+    status = ucc_tl_ucp_recv_nb(rbuf, data_size, rmem, rank, team, task);
+    if (UCC_OK != status) {
+        task->super.status = status;
+        return task->super.status;
+    }
+    return UCC_OK;
+}
+ucc_status_t allgather_copy(void *rbuf, void *sbuf, size_t data_size,
+                            ucc_memory_type_t rmem, ucc_memory_type_t smem,
+                            ucc_rank_t rank, ucc_tl_ucp_team_t *team,
+                            ucc_tl_ucp_task_t *task)
+{
+    ucc_status_t status;
+    int use_loopback = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_loopback;
+    if (use_loopback) {
+        status = loopback_self_copy(rbuf, sbuf, data_size, rmem, smem, rank,
+                                    team, task);
+    } else {
+        status = ucc_mc_memcpy(rbuf, sbuf, data_size, rmem, smem);
+    }
+    return status;
+}
--- a/src/components/tl/ucp/allgather/allgather_knomial.c
+++ b/src/components/tl/ucp/allgather/allgather_knomial.c
@@ -13,6 +13,7 @@
 #include "coll_patterns/sra_knomial.h"
 #include "utils/ucc_math.h"
 #include "utils/ucc_coll_utils.h"
+#include "allgather.h"
 
 #define SAVE_STATE(_phase)                                                     \
     do {                                                                       \
@@ -54,8 +55,7 @@
 
 void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
 {
-    ucc_tl_ucp_task_t     *task      = ucc_derived_of(coll_task,
-                                                      ucc_tl_ucp_task_t);
+    ucc_tl_ucp_task_t *    task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
     ucc_coll_args_t       *args      = &TASK_ARGS(task);
     ucc_tl_ucp_team_t     *team      = TASK_TEAM(task);
     ucc_kn_radix_t         radix     = task->allgather_kn.p.radix;
@@ -66,10 +66,10 @@ void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
     size_t                 dt_size   = ucc_dt_size(GET_DT(args));
     ucc_rank_t             size      = task->subset.map.ep_num;
     size_t                 data_size = GET_TOTAL_COUNT(args, size);
-    ucc_rank_t             broot     = args->coll_type == UCC_COLL_TYPE_BCAST ?
-                                       args->root : 0;
-    ucc_rank_t             rank      = VRANK(task->subset.myrank, broot, size);
-    size_t                 local     = GET_LOCAL_COUNT(args, size, rank);
+    ucc_rank_t broot = args->coll_type == UCC_COLL_TYPE_BCAST ? args->root : 0;
+    ucc_rank_t rank  = VRANK(task->subset.myrank, broot, size);
+    size_t     local = GET_LOCAL_COUNT(args, size, rank);
+    int use_loopback = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_loopback;
     void                  *sbuf;
     ptrdiff_t              peer_seg_offset, local_seg_offset;
     ucc_rank_t             peer, peer_dist;
@@ -78,8 +78,14 @@ void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
     ucc_status_t           status;
     size_t                 extra_count;
 
-    EXEC_TASK_TEST(UCC_KN_PHASE_INIT, "failed during ee task test",
-                   task->allgather_kn.etask);
+    if (use_loopback) {
+        if (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
+            return;
+        }
+    } else {
+        EXEC_TASK_TEST(UCC_KN_PHASE_INIT, "failed during ee task test",
+                       task->allgather_kn.etask);
+    }
     task->allgather_kn.etask = NULL;
     UCC_KN_GOTO_PHASE(task->allgather_kn.phase);
     if (KN_NODE_EXTRA == node_type) {
@@ -209,6 +215,7 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
                                               ct == UCC_COLL_TYPE_BCAST ?
                                               args->root : 0, size);
     ucc_ee_executor_task_args_t eargs = {0};
+    int use_loopback = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_loopback;
     ucc_status_t       status;
     ptrdiff_t          offset;
     ucc_ee_executor_t *exec;
@@ -225,21 +232,34 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
                  ucc_dt_size(args->dst.info.datatype);
         rbuf   = args->dst.info.buffer;
         if (!UCC_IS_INPLACE(*args)) {
-            status = ucc_coll_task_get_executor(&task->super, &exec);
-            if (ucc_unlikely(status != UCC_OK)) {
-                task->super.status = status;
-                return status;
-            }
-            eargs.task_type = UCC_EE_EXECUTOR_TASK_COPY;
-            eargs.copy.dst  = PTR_OFFSET(args->dst.info.buffer, offset);
-            eargs.copy.src  = args->src.info.buffer;
-            eargs.copy.len  = args->src.info.count *
-                              ucc_dt_size(args->src.info.datatype);
-            status = ucc_ee_executor_task_post(exec, &eargs,
-                                               &task->allgather_kn.etask);
-            if (ucc_unlikely(status != UCC_OK)) {
-                task->super.status = status;
-                return status;
+            if (use_loopback) {
+                status = loopback_self_copy(
+                    PTR_OFFSET(args->dst.info.buffer, offset),
+                    args->src.info.buffer,
+                    args->src.info.count * ucc_dt_size(args->src.info.datatype),
+                    args->dst.info.mem_type, args->src.info.mem_type, rank,
+                    team, task);
+                if (ucc_unlikely(status != UCC_OK)) {
+                    return status;
+                }
+            } else {
+                /* Executer */
+                status = ucc_coll_task_get_executor(&task->super, &exec);
+                if (ucc_unlikely(status != UCC_OK)) {
+                    task->super.status = status;
+                    return status;
+                }
+                eargs.task_type = UCC_EE_EXECUTOR_TASK_COPY;
+                eargs.copy.dst  = PTR_OFFSET(args->dst.info.buffer, offset);
+                eargs.copy.src  = args->src.info.buffer;
+                eargs.copy.len =
+                    args->src.info.count * ucc_dt_size(args->src.info.datatype);
+                status = ucc_ee_executor_task_post(exec, &eargs,
+                                                   &task->allgather_kn.etask);
+                if (ucc_unlikely(status != UCC_OK)) {
+                    task->super.status = status;
+                    return status;
+                }
             }
         }
     } else if (ct == UCC_COLL_TYPE_ALLGATHERV) {
--- a/src/components/tl/ucp/allgather/allgather_neighbor.c
+++ b/src/components/tl/ucp/allgather/allgather_neighbor.c
@@ -81,9 +81,11 @@ void ucc_tl_ucp_allgather_neighbor_progress(ucc_coll_task_t *coll_task)
     ucc_datatype_t     dt        = TASK_ARGS(task).dst.info.datatype;
     size_t             count     = TASK_ARGS(task).dst.info.count;
     size_t             data_size = (count / tsize) * ucc_dt_size(dt);
+    int use_loopback = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_loopback;
     ucc_rank_t         neighbors[2], i;
     int                i_parity, even_rank;
     void              *tmprecv, *tmpsend;
+    int                counter;
 
     if (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
         return;
@@ -98,8 +100,13 @@ void ucc_tl_ucp_allgather_neighbor_progress(ucc_coll_task_t *coll_task)
         neighbors[1] = (trank + 1) % tsize;
     }
 
-    while (task->tagged.send_posted < (tsize / 2)) {
-        i        = task->tagged.send_posted;
+    if ((!UCC_IS_INPLACE(TASK_ARGS(task))) && use_loopback) {
+        counter = task->tagged.send_posted - 1;
+    } else {
+        counter = task->tagged.send_posted;
+    }
+    while (counter < (tsize / 2)) {
+        i        = counter;
         i_parity = i % 2;
 
         tmprecv =
@@ -117,6 +124,11 @@ void ucc_tl_ucp_allgather_neighbor_progress(ucc_coll_task_t *coll_task)
 
         if (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
             return;
+        }
+        if ((!UCC_IS_INPLACE(TASK_ARGS(task))) && use_loopback) {
+            counter = task->tagged.send_posted - 1;
+        } else {
+            counter = task->tagged.send_posted;
         }
     }
 
@@ -150,13 +162,15 @@ ucc_status_t ucc_tl_ucp_allgather_neighbor_start(ucc_coll_task_t *coll_task)
     ucc_tl_ucp_task_reset(task, UCC_INPROGRESS);
 
     if (!UCC_IS_INPLACE(TASK_ARGS(task))) {
-        status = ucc_mc_memcpy(PTR_OFFSET(rbuf, data_size * trank), sbuf,
-                               data_size, rmem, smem);
+        status = allgather_copy(PTR_OFFSET(rbuf, data_size * trank), sbuf,
+                                data_size, rmem, smem, trank, team, task);
         if (ucc_unlikely(UCC_OK != status)) {
             return status;
         }
     }
 
+    while ((!UCC_IS_INPLACE(TASK_ARGS(task))) && (UCC_INPROGRESS == ucc_tl_ucp_test(task))) {
+    }
     if (trank % 2) {
         neighbor = (trank - 1 + tsize) % tsize;
     } else {
--- a/src/components/tl/ucp/allgather/allgather_ring.c
+++ b/src/components/tl/ucp/allgather/allgather_ring.c
@@ -31,15 +31,16 @@ static ucc_rank_t ucc_tl_ucp_allgather_ring_get_recv_block(ucc_subset_t *subset,
 
 void ucc_tl_ucp_allgather_ring_progress(ucc_coll_task_t *coll_task)
 {
-    ucc_tl_ucp_task_t *task       = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
-    ucc_tl_ucp_team_t *team       = TASK_TEAM(task);
-    ucc_rank_t         trank      = task->subset.myrank;
-    ucc_rank_t         tsize      = (ucc_rank_t)task->subset.map.ep_num;
-    void              *rbuf       = TASK_ARGS(task).dst.info.buffer;
-    ucc_memory_type_t  rmem       = TASK_ARGS(task).dst.info.mem_type;
-    size_t             count      = TASK_ARGS(task).dst.info.count;
-    ucc_datatype_t     dt         = TASK_ARGS(task).dst.info.datatype;
-    size_t             data_size  = (count / tsize) * ucc_dt_size(dt);
+    ucc_tl_ucp_task_t *task      = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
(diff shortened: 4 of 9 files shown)


Sergei-Lebedev:
ok to test

src/components/cl/basic/cl_basic_context.c
```diff
@@ -47,7 +47,7 @@ UCC_CLASS_INIT_FUNC(ucc_cl_basic_context_t,
         self->super.tl_ctxs = NULL;
         return UCC_ERR_NOT_FOUND;
     }
-    cl_debug(cl_config->cl_lib, "initialized cl context: %p", self);
+    cl_debug(cl_config->cl_lib, "initialized :) cl context: %p", self);
```
nsarka, line 50: Is this left over from a debugging? Is this change needed?

src/components/ec/cuda/ec_cuda_executor.c
```diff
@@ -45,7 +45,7 @@ ucc_status_t ucc_cuda_executor_init(const ucc_ee_executor_params_t *params,
         eee->requested_ops = 1;
     }
 
-    ec_debug(&ucc_ec_cuda.super, "executor init, eee: %p", eee);
+    //ec_debug(&ucc_ec_cuda.super, "executor init, eee: %p", eee);
```
nsarka, line 48: Remove unused code

src/components/ec/cuda/ec_cuda_executor.c
```diff
@@ -76,7 +76,7 @@ ucc_status_t ucc_cuda_executor_finalize(ucc_ee_executor_t *executor)
     ucc_ec_cuda_executor_t *eee = ucc_derived_of(executor,
                                                  ucc_ec_cuda_executor_t);
 
-    ec_debug(&ucc_ec_cuda.super, "executor free, eee: %p", eee);
+    //ec_debug(&ucc_ec_cuda.super, "executor free, eee: %p", eee);
```
nsarka, line 79: Remove unused code

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -78,9 +82,17 @@ void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
     ucc_status_t           status;
     size_t                 extra_count;
 
-    EXEC_TASK_TEST(UCC_KN_PHASE_INIT, "failed during ee task test",
-                   task->allgather_kn.etask);
+    uint32_t USE_CUDA = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_cuda;
+
+    if(!USE_CUDA){
+        if (UCC_INPROGRESS == ucc_tl_ucp_test(task)){
```
nsarka, line 88: Would it be more concise to have`if(!USE_CUDA && UCC_INPROGRESS == ucc_tl_ucp_test(task))`?

Also, if `USE_CUDA` is a local variable it should be snake case by convention, `use_cuda`

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -225,21 +240,29 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
...
+                                ucc_dt_size(args->src.info.datatype);
+                status = ucc_ee_executor_task_post(exec, &eargs,
+                                                &task->allgather_kn.etask);
+                if (ucc_unlikely(status != UCC_OK)) {
+                    task->super.status = status;
+                    return status;
+                }
+            } else {
+                /*Loopback*/
+                UCPCHECK_GOTO(ucc_tl_ucp_send_nb(args->src.info.buffer, args->src.info.count * ucc_dt_size(args->src.info.datatype),
```
nsarka, line 262: Why not UCC_CHECK_GOTO?

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -293,8 +318,9 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
     ucc_tl_ucp_team_t *tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
     ucc_tl_ucp_task_t *task;
     ucc_sbgp_t        *sbgp;
-
     task = ucc_tl_ucp_init_task(coll_args, team);
+    //ucc_coll_args_t   *args = &coll_args->args;
```
nsarka, line 322: Remove unused code

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -306,6 +332,13 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
     task->super.post           = ucc_tl_ucp_allgather_knomial_start;
     task->super.progress       = ucc_tl_ucp_allgather_knomial_progress;
     *task_h                    = &task->super;
+    /*if(!UCC_IS_INPLACE(*args)){
```
nsarka, line 335: Remove unused code

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -175,4 +183,6 @@ ucc_status_t ucc_tl_ucp_allgather_neighbor_start(ucc_coll_task_t *coll_task)
         task, out);
 out:
     return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+out2:
```
nsarka, line 186: Rename to `err` instead of out2?

src/components/tl/ucp/tl_ucp.c
```diff
@@ -140,6 +140,10 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT},
 
+     {"ALLGATHER_USE_CUDA", "1", "Flag weather to use mc cuda copy",
```
nsarka, line 143: Should it be "Flag whether to use network loopback for cuda copy"? I think "mc cuda" is just saying it's GPU memory

nsarka (approved):

src/components/tl/ucp/tl_ucp.c
nsarka: If so, should the flag be "ALLGATHER_USE_LOOPBACK"?

janjust:
@Sergei-Lebedev why do we have to keep approving runs when force pushing?? Seems an outlier here

src/components/tl/ucp/allgather/allgather_ring.c
```diff
@@ -89,21 +92,30 @@ ucc_status_t ucc_tl_ucp_allgather_ring_start(ucc_coll_task_t *coll_task)
...
+            status = ucc_mc_memcpy(PTR_OFFSET(rbuf, data_size * block),
                                sbuf, data_size, rmem, smem);
-        if (ucc_unlikely(UCC_OK != status)) {
-            return status;
+            if (ucc_unlikely(UCC_OK != status)) {
+                return status;
+            }
+        } else {
+            /* Loopback */
+            ucc_rank_t rank = ucc_ep_map_eval(task->subset.map, trank);
```
Sergei-Lebedev, line 111: declare variables in the beginning of function

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -144,16 +144,23 @@ ucc_status_t ucc_tl_ucp_allgather_neighbor_start(ucc_coll_task_t *coll_task)
...
                                data_size, rmem, smem);
-        if (ucc_unlikely(UCC_OK != status)) {
-            return status;
+            if (ucc_unlikely(UCC_OK != status)) {
+                return status;
+            }
+        } else {
+            /* Loopback */
+            UCPCHECK_GOTO(ucc_tl_ucp_send_nb(sbuf, data_size, smem, trank, team, task),task, err);
+            UCPCHECK_GOTO(ucc_tl_ucp_recv_nb(PTR_OFFSET(rbuf, data_size * trank), data_size, rmem, trank, team, task),task, err);
```
Sergei-Lebedev, line 163: maybe make a loopback copy function, this pattern repeats in every algorithm

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -77,9 +77,15 @@ void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
...
 
-    EXEC_TASK_TEST(UCC_KN_PHASE_INIT, "failed during ee task test",
-                   task->allgather_kn.etask);
+    if (!use_cuda) {
+        if (UCC_INPROGRESS == ucc_tl_ucp_test(task)){
+            return;
+        }
+    }
+    
+    if (use_cuda) EXEC_TASK_TEST(UCC_KN_PHASE_INIT, "failed during ee task test", task->allgather_kn.etask);
```
Sergei-Lebedev, line 88: just else for prev if

src/components/tl/ucp/allgather/allgather_ring.c
```diff
@@ -43,15 +43,18 @@ void ucc_tl_ucp_allgather_ring_progress(ucc_coll_task_t *coll_task)
...
         return;
     }
     sendto   = ucc_ep_map_eval(task->subset.map, (trank + 1) % tsize);
     recvfrom = ucc_ep_map_eval(task->subset.map, (trank - 1 + tsize) % tsize);
 
-    while (task->tagged.send_posted < tsize - 1) {
-        step = task->tagged.send_posted;
+    int iter = use_cuda ? tsize - 1 : tsize; //when using loopback tagged.send_posted has 1 more which will cause non-complete ring algorithm
+    while (task->tagged.send_posted < iter) {
+        step = use_cuda ? task->tagged.send_posted : task->tagged.send_posted - 1; //when using loopback tagged.send_posted has 1 more which will cause wrong calculation of send/recv
```
Sergei-Lebedev, line 56: since loopback affects algorithm flow I would rather consider checking loopback copy result somehow differently, maybe add custom callback for ucp recv completion

src/components/tl/ucp/allgather/allgather_sparbit.c
```diff
@@ -123,20 +123,29 @@ ucc_status_t ucc_tl_ucp_allgather_sparbit_start(ucc_coll_task_t *coll_task)
...
                                data_size, rmem, smem);
-        if (ucc_unlikely(UCC_OK != status)) {
-            return status;
+            if (ucc_unlikely(UCC_OK != status)) {
+                return status;
+            }
+        } else {
+            /* Loopback */
+            UCPCHECK_GOTO(ucc_tl_ucp_send_nb(sbuf, data_size, smem, trank, team, task),task, out);
+            UCPCHECK_GOTO(ucc_tl_ucp_recv_nb(PTR_OFFSET(rbuf, data_size * trank), data_size, rmem, trank, team, task),task, out);
```
samnordmann, line 144: this whole code block is duplicated in neighbor, ring and sparbit. It should be refactored into a helper function and called there.

Ideally, this could also be refactored with the call from knomial, with an additional control about whether to use `ucc_mc_memcpy` or the executor.

src/components/tl/ucp/allgather/allgather_sparbit.c
```diff
@@ -123,20 +123,29 @@ ucc_status_t ucc_tl_ucp_allgather_sparbit_start(ucc_coll_task_t *coll_task)
...
         }
     }
 
     return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
+out:
```
samnordmann, line 149: a goto redirection is not needed here. We could simply check status when it is returned, and return it if it is in error state.

Apply in other places of this PR where this  pattern is used

src/components/tl/ucp/allgather/allgather_ring.c
```diff
@@ -43,15 +43,18 @@ void ucc_tl_ucp_allgather_ring_progress(ucc_coll_task_t *coll_task)
...
 
     if (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
         return;
     }
     sendto   = ucc_ep_map_eval(task->subset.map, (trank + 1) % tsize);
     recvfrom = ucc_ep_map_eval(task->subset.map, (trank - 1 + tsize) % tsize);
 
-    while (task->tagged.send_posted < tsize - 1) {
-        step = task->tagged.send_posted;
+    int iter = use_loopback ? tsize : tsize - 1; //when using loopback tagged.send_posted has 1 more which will cause non-complete ring algorithm
```
samnordmann, line 54: declaration of variable goes to the beginning of the block

src/components/tl/ucp/allgather/allgather_ring.c
```diff
@@ -43,15 +43,18 @@ void ucc_tl_ucp_allgather_ring_progress(ucc_coll_task_t *coll_task)
     ucc_rank_t         sendto, recvfrom, sblock, rblock;
     int                step;
     void              *buf;
+    int                use_loopback = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_loopback;
```
samnordmann, line 46: initialized variables go first

src/components/tl/ucp/allgather/allgather_ring.c
samnordmann: I am not sure to understand the comment. What do you mean by "will cause non-complete ring algorithm"? 
Can you use a more explicit name than `iter` for that variable?

samnordmann (commented):
Looks good, left a bunch of comments mainly on codestyle.

However, please add tests for that feature.

Also please note from the CI:
```
Commit title is too long: 59
Bad commit title: 'TL/UCP: Allow self copy in allgather using network loopback'
```

src/components/tl/ucp/tl_ucp.c
```diff
@@ -140,6 +140,10 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT},
 
+     {"ALLGATHER_USE_CUDA", "1", "If set to 1 uses mc cuda copy, otherwise performs loopback for self copy",
```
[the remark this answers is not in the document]
yaeliyac: this was just changed

samnordmann:
@yaeliyac btw, for knomial inplace, we don't expect any change, right? Do we have an explanation for the degradation in the graph?

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -225,21 +234,31 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
...
-            eargs.copy.dst  = PTR_OFFSET(args->dst.info.buffer, offset);
-            eargs.copy.src  = args->src.info.buffer;
-            eargs.copy.len  = args->src.info.count *
-                              ucc_dt_size(args->src.info.datatype);
-            status = ucc_ee_executor_task_post(exec, &eargs,
-                                               &task->allgather_kn.etask);
-            if (ucc_unlikely(status != UCC_OK)) {
-                task->super.status = status;
-                return status;
+            if (!use_loopback) {
```
ikryukov, line 237: minor: suggestion for better readability is to use direct logic in cases:
```
if (use_loopback) {
...
} else {
...
}
```

src/components/tl/ucp/tl_ucp.c
```diff
@@ -140,6 +140,10 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT},
 
+     {"ALLGATHER_USE_LOOPBACK", "0", "If set to 1 uses mc cuda copy, otherwise performs loopback for self copy",
```
ikryukov, line 143: minor: alignment of {

src/components/tl/ucp/allgather/allgather.c
```diff
@@ -58,3 +58,22 @@ char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
...
+    }
+    status = ucc_tl_ucp_recv_nb(rbuf, data_size, rmem, rank, team, task);
+    if (UCC_OK != status) {
+        task->super.status = status;
+        return task->super.status;
+    }
+    return UCC_OK;
+}
+
+
```
ikryukov, line 79: minor: extra line

src/components/tl/ucp/allgather/allgather_sparbit.c
```diff
@@ -111,28 +111,35 @@ void ucc_tl_ucp_allgather_sparbit_progress(ucc_coll_task_t *coll_task)
...
+    ucc_memory_type_t  rmem         = TASK_ARGS(task).dst.info.mem_type;
+    ucc_datatype_t     dt           = TASK_ARGS(task).dst.info.datatype;
+    ucc_rank_t         trank        = UCC_TL_TEAM_RANK(team);
+    ucc_rank_t         tsize        = UCC_TL_TEAM_SIZE(team);
+    size_t             data_size    = (count / tsize) * ucc_dt_size(dt);
+    int                use_loopback = UCC_TL_UCP_TEAM_LIB(team)->cfg.allgather_use_loopback;
     ucc_status_t       status;
-
+    
+
```
ikryukov, line 128: extra line

src/components/tl/ucp/allgather/allgather_bruck.c
```diff
@@ -254,4 +254,4 @@ ucc_status_t ucc_tl_ucp_allgather_bruck_start(ucc_coll_task_t *coll_task)
     }
 
     return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
-}
+}
```
ikryukov, line 257: new line

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -54,22 +55,21 @@
 
 void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
 {
-    ucc_tl_ucp_task_t     *task      = ucc_derived_of(coll_task,
-                                                      ucc_tl_ucp_task_t);
-    ucc_coll_args_t       *args      = &TASK_ARGS(task);
-    ucc_tl_ucp_team_t     *team      = TASK_TEAM(task);
-    ucc_kn_radix_t         radix     = task->allgather_kn.p.radix;
+    ucc_tl_ucp_task_t *    task  = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
```
ikryukov, line 58: suggestion is not to reformat existing code (* symbol), because clang format formats different depending on version, so it is better to leave as before. @samnordmann @Sergei-Lebedev what do you think?

samnordmann: indeed, clang format often does this undesirable reformatting... We should have a long-term fix -- in the meantime I suggest to manually place the "*" where they should

src/components/tl/ucp/allgather/allgather_neighbor.c
```diff
@@ -150,13 +162,15 @@ ucc_status_t ucc_tl_ucp_allgather_neighbor_start(ucc_coll_task_t *coll_task)
...
-        status = ucc_mc_memcpy(PTR_OFFSET(rbuf, data_size * trank), sbuf,
-                               data_size, rmem, smem);
+        status = allgather_copy(PTR_OFFSET(rbuf, data_size * trank), sbuf,
+                                data_size, rmem, smem, trank, team, task);
         if (ucc_unlikely(UCC_OK != status)) {
             return status;
         }
     }
 
+    while (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
```
samnordmann, line 172: just checking: is this call safe in the "inplace" case ?

src/components/tl/ucp/tl_ucp.c
```diff
@@ -140,6 +140,12 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT},
 
+    {"ALLGATHER_USE_LOOPBACK", "0",
+     "If set to 1 performs network loopback for self copy, otherwise uses mc "
+     "cuda copy",
```
samnordmann, line 145:
```suggestion
     "If set to 1 performs network loopback for self copy, otherwise uses mc ",
```
it is not necessarily cuda

samnordmann (approved):
LGTM!

src/components/tl/ucp/allgather/allgather.c
```diff
@@ -58,3 +58,37 @@ char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
...
     return str;
 }
+
+ucc_status_t loopback_self_copy(void *rbuf, void *sbuf, size_t data_size,
+                                ucc_memory_type_t rmem, ucc_memory_type_t smem,
+                                ucc_rank_t rank, ucc_tl_ucp_team_t *team,
+                                ucc_tl_ucp_task_t *task)
+{
+    ucc_status_t status;
+    status = ucc_tl_ucp_send_nb(sbuf, data_size, smem, rank, team, task);
```
nsarka, line 68: Instead of `status = ... ; if (UCC_OK != status) ...` you can use UCC_CHECK_GOTO(..., err)

src/components/tl/ucp/allgather/allgather.c
```diff
@@ -58,3 +58,37 @@ char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
...
+    if (UCC_OK != status) {
+        task->super.status = status;
+        return task->super.status;
+    }
+    status = ucc_tl_ucp_recv_nb(rbuf, data_size, rmem, rank, team, task);
+    if (UCC_OK != status) {
+        task->super.status = status;
+        return task->super.status;
+    }
+    return UCC_OK;
```
nsarka, line 78:
```
err:
    return status;
```

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -54,8 +55,7 @@
 
 void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
 {
-    ucc_tl_ucp_task_t     *task      = ucc_derived_of(coll_task,
-                                                      ucc_tl_ucp_task_t);
+    ucc_tl_ucp_task_t *    task = ucc_derived_of(coll_task, ucc_tl_ucp_task_t);
```
nsarka, line 58: *task

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -66,10 +66,10 @@ void ucc_tl_ucp_allgather_knomial_progress(ucc_coll_task_t *coll_task)
     size_t                 dt_size   = ucc_dt_size(GET_DT(args));
     ucc_rank_t             size      = task->subset.map.ep_num;
     size_t                 data_size = GET_TOTAL_COUNT(args, size);
-    ucc_rank_t             broot     = args->coll_type == UCC_COLL_TYPE_BCAST ?
-                                       args->root : 0;
-    ucc_rank_t             rank      = VRANK(task->subset.myrank, broot, size);
-    size_t                 local     = GET_LOCAL_COUNT(args, size, rank);
+    ucc_rank_t broot = args->coll_type == UCC_COLL_TYPE_BCAST ? args->root : 0;
```
nsarka, line 69: align =

src/components/tl/ucp/allgather/allgather_ring.c
```diff
@@ -67,6 +69,8 @@ void ucc_tl_ucp_allgather_ring_progress(ucc_coll_task_t *coll_task)
         if (UCC_INPROGRESS == ucc_tl_ucp_test(task)) {
             return;
         }
+        step = use_loopback ? task->tagged.send_posted - 1
```
nsarka, line 72: Should this line check for inplace as well? If use_loopback is true and inplace is false, will step be incorrect?

nsarka: Similar for neighbor exchange

janjust:
@Sergei-Lebedev @samnordmann @yaeliyac @lappazos 

The more I look at this implementation the less I think this is the right approach, and I'm surprised we would allow this in. We're introducing an if-else statement all throughout critical paths of the algorithms to avoid this issues specific to the nccl_plugin and only during inplace.
What if we need to add additional algorithms, then we need to be aware of this limitation and add the same if-else, or if we need to expand this for A2A, A2AV, or AGV? Why is AG the only collective that suffers this deadlock?
Thinking about this in more detail I don't think this is a good approach, we should really consider this carefully, regardless of the deadlock fix. As an immediate alternative, I think at the very least we should provide a loop-back copy or the mc_copy from a higher level.

But on a broader note, what exactly is the deadlock here? It seems to be an issue specifically with nccl_plugin, and why does loopback fix it?

Can we have you guys jump on another devcall and review this with us please.

src/components/tl/ucp/allgather/allgather.c
```diff
@@ -58,3 +58,37 @@ char *ucc_tl_ucp_allgather_score_str_get(ucc_tl_ucp_team_t *team)
...
+}
+ucc_status_t allgather_copy(void *rbuf, void *sbuf, size_t data_size,
```
nsarka, line 80: `allgather_copy` should probably be renamed to `copy_sbuf_inplace`

samnordmann:
> @Sergei-Lebedev @samnordmann @yaeliyac @lappazos
> 
> The more I look at this implementation the less I think this is the right approach, and I'm surprised we would allow this in. We're introducing an if-else statement all throughout critical paths of the algorithms to avoid this issues specific to the nccl_plugin and only during inplace. What if we need to add additional algorithms, then we need to be aware of this limitation and add the same if-else, or if we need to expand this for A2A, A2AV, or AGV? Why is AG the only collective that suffers this deadlock? Thinking about this in more detail I don't think this is a good approach, we should really consider this carefully, regardless of the deadlock fix. As an immediate alternative, I think at the very least we should provide a loop-back copy or the mc_copy from a higher level.
> 
> But on a broader note, what exactly is the deadlock here? It seems to be an issue specifically with nccl_plugin, and why does loopback fix it?
> 
> Can we have you guys jump on another devcall and review this with us please.

I agree that the right approach is to add an EC component for copying data using the NIC on a loopback connection. We discussed this during the code review, but IIRC it was postponed to later because of a certain sense of urgency regarding merging the present PR.

PR #1022, opened 2024-09-19
TL/MLX5: fix memtype in bcast reliability

MamziB:
TL/MLX5: fix memtype in bcast reliability

### DIFF
--- a/src/components/tl/mlx5/mcast/tl_mlx5_mcast_coll.c
+++ b/src/components/tl/mlx5/mcast/tl_mlx5_mcast_coll.c
@@ -182,7 +182,9 @@ static inline ucc_status_t ucc_tl_mlx5_mcast_prepare_bcast(void* buf, size_t siz
     req->am_root = (root == comm->rank);
     req->mr      = comm->pp_mr;
     req->rreg    = NULL;
-    req->proto   = (req->length < comm->max_eager) ? MCAST_PROTO_EAGER : MCAST_PROTO_ZCOPY;
+    /* cost of copy is too high in cuda buffers */
+    req->proto   = (req->length < comm->max_eager && !comm->cuda_mem_enabled) ?
+                            MCAST_PROTO_EAGER : MCAST_PROTO_ZCOPY;
 
     status = ucc_tl_mlx5_mcast_prepare_reliable(comm, req, req->root);
     if (ucc_unlikely(UCC_OK != status)) {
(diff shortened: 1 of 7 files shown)


samnordmann (commented):
I understand it is a fix, but I am surprised we didn't catch that bug before. It seems that we are not testing mcast bcast cuda memory at all, is it correct? If so, can you please add a test?

MamziB:
@samnordmann you are correct, it was a shortcoming from my side. Reliablity protocol kicks when a packet is dropped and it does not happen in normal scenarios.
Previously, I tried forcing the dropped packet in the HOST version for testing purposes, but not for GPU. Since we saw a dropped packet on the ROCK system, where MTT runs, we saw this error, and here is a fix for that.

samnordmann:
> @samnordmann you are correct, it was a shortcoming from my side. Reliablity protocol kicks when a packet is dropped and it does not happen in normal scenarios. Previously, I tried forcing the dropped packet in the HOST version for testing purposes, but not for GPU. Since we saw a dropped packet on the ROCK system, where MTT runs, we saw this error, and here is a fix for that.

Can you please add a test, covering both Host and CUDA memtype, checking that the reliability is not boggy?

MamziB:
@samnordmann OK I will open a separate PR soon

janjust:
@MamziB no need for a separate PR, just push the commit here

MamziB:
@janjust I have added the gtest before but it fails the github ucc tests for some unknown reasons. I rather open a new PR so that it does not affect this PR as I would like this PR to go in ASAP. I will open a PR today for that gtest changes.

janjust:
I don't think this PR will go in until the test is in, open the other PR, but then this PR will be predicated on the test going in.

MamziB:
I opened https://github.com/openucx/ucc/pull/1023 for testing mcast

janjust:
@samnordmann @Sergei-Lebedev 
https://github.com/openucx/ucc/pull/1023

Here is the test

samnordmann:
> @samnordmann @Sergei-Lebedev #1023
> 
> Here is the test

Thank you, but shouldn't we also test the reliability protocol in case of dropped packets? The present PR fixes a bug that we cannot catch nor reproduce, so how can we ensure that it is effectively fixed?

I don't want to block the PR if there are some time considerations though, so let me know what you guys think

MamziB:
@samnordmann Sure, I will update the other PR to force the bcast to have dropped packets

samnordmann (approved):
Thanks!

MamziB:
@janjust can you please let me know if you have comments are this?

janjust:
@MamziB not specifically on this PR, but I'm still curious why the other PR, the test is not failing, I haven't investigated it yet.

MamziB:
@janjust thanks Tommy. Since I am preparing a PR which is gonna be based on this PR, can we merge this PR so that we can proceed?

janjust (approved):

src/components/tl/mlx5/mcast/tl_mlx5_mcast_coll.c
```diff
@@ -300,6 +302,33 @@ ucc_status_t ucc_tl_mlx5_mcast_check_memory_type_cap(ucc_base_coll_args_t *coll_
...
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    if (UCC_OK != ucc_tl_mlx5_mcast_check_memory_type_cap(coll_args, team)) {
+        tl_trace(team->context->lib, "mcast bcast not compatible with this memory type");
+        return UCC_ERR_NOT_SUPPORTED;
+    }
+
+    if (args->src.info.mem_type == UCC_MEMORY_TYPE_CUDA &&
+            buf_size > 4000) {
```
Sergei-Lebedev, line 322: pls make a macro for 4000

Sergei-Lebedev (approved):

PR #1025, opened 2024-09-24
MC/CUDA: reduce verbosity for cu ctx doesn't exist

Sergei-Lebedev:
## What
Don't print error if cuCtx doesn't exist

## Why ?
UCC can continue and fallback to default ctx


samnordmann (approved):
update copyright

janjust (approved):

manjugv (approved):

PR #1026, opened 2024-09-27
CORE: Fix asym mem for triggered tasks

nsarka:
This PR initializes the scratch buffer to NULL in `ucc_coll_task_init`. The value not being NULL was causing the asymmetric memory copy-out to run for triggered tasks, making the DLRM jenkins test fail.

### DIFF
--- a/src/schedule/ucc_schedule.c
+++ b/src/schedule/ucc_schedule.c
@@ -119,6 +119,10 @@ ucc_status_t ucc_coll_task_init(ucc_coll_task_t *task,
     task->post                 = ucc_dummy_post;
     task->finalize             = ucc_dummy_finalize;
     task->progress             = ucc_dummy_progress;
+
+    // Prevent asymmetric memory copy-out of garbage address at task complete
+    task->bargs.asymmetric_save_info.scratch = NULL;
+
     if (bargs) {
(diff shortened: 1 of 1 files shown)


src/schedule/ucc_schedule.c
```diff
@@ -119,6 +119,10 @@ ucc_status_t ucc_coll_task_init(ucc_coll_task_t *task,
     task->post                 = ucc_dummy_post;
     task->finalize             = ucc_dummy_finalize;
     task->progress             = ucc_dummy_progress;
+
+    // Prevent asymmetric memory copy-out of garbage address at task complete
+    task->bargs.asymmetric_save_info.scratch = NULL;
```
nsarka, line 124: Should this be aligned with the previous assignments, even though I added a comment and some spacing?

samnordmann (approved):

src/schedule/ucc_schedule.c
janjust: I think it looks cleaner if the majority are aligned, and the one off, whose nested elements would skew everything, is best left as the outlier, just my 2c

janjust (approved):

Sergei-Lebedev (approved):

PR #1027, opened 2024-10-01
TL/UCP: Grace tuning

nsarka:
This PR updates UCC to allow a socket range in ucc.conf in addition to vendor/arch/nodes/ppn and uses it to tune TL/UCP for Grace systems.

### DIFF
--- a/src/utils/ucc_parser.c
+++ b/src/utils/ucc_parser.c
@@ -594,6 +594,7 @@ ucc_status_t ucc_add_team_sections(void                *team_cfg,
     ucc_rank_t             sock_max  = ucc_topo_max_socket_size(team_topo);
     ucc_rank_t             nnodes    = ucc_topo_nnodes(team_topo);
     ucc_rank_t             team_size = team_topo->set.map.ep_num;
+    int                    found     = 0;
     khash_t(ucc_sec)      *sec_h;
     khiter_t               i, j;
     const char            *sec_name;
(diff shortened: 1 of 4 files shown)


nsarka:
Once my data collection finishes I will update the conf to have the proper Grace tuning

samnordmann (approved):

src/utils/ucc_parser.c
```diff
@@ -589,18 +606,18 @@ ucc_status_t ucc_add_team_sections(void                *team_cfg,
...
+                              ppn_min, ppn_max, sock_min, sock_max, nnodes)) {
             sec_h = &sec->vals_h;
             j = kh_get(ucc_sec, sec_h, tune_key);
             if (j != kh_end(sec_h)) {
                 *tuning_str = kh_val(sec_h, j);
             }
             status = ucc_apply_file_cfg(team_cfg, tl_fields, env_prefix,
                                         component_prefix, sec_name);
-            return status;
+            found = 1;
```
nsarka, line 617: This is to allow the conf file to match multiple sections and take whatever section is matched last. The idea is that in the conf, the sections can be sorted from generic to specific going from top of the file to bottom. This allows a generic Grace config in case the more specific configs (say 2 socket 72 procs/socket) don't match

janjust (approved):

Sergei-Lebedev (approved):

Issue #1031, opened 2024-10-08
ucc compilation is getting failed on master branch


shashank-parsi:
Hello All,
I see there is an compilation issue on ucc with master branch.

steps followed:
1.  git clone https://github.com/openucx/ucc.git
2. cd ucc
3. ./autogen.sh
4. ./configure --prefix=<UCC installed dir>  --with-ucx=<UCX installed Dir> --with-rocm=<ROCm installed dir> --enable-gtest
5. make -j 

issue seen:
make[3]: Entering directory '/home/master/rastra/rocm_tests/hipmpi/ucc/src/components/tl/ucp'
  CC       libucc_tl_ucp_la-tl_ucp_dpu_offload.lo
tl_ucp_dpu_offload.c: In function â€˜ucc_tl_ucp_allreduce_sliding_window_registerâ€™:
tl_ucp_dpu_offload.c:18:35: error: â€˜UCP_MEM_MAP_PARAM_FIELD_EXPORTED_MEMH_BUFFERâ€™ undeclared (first use in this function)
   18 |     params.field_mask           = UCP_MEM_MAP_PARAM_FIELD_EXPORTED_MEMH_BUFFER;
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tl_ucp_dpu_offload.c:18:35: note: each undeclared identifier is reported only once for each function it appears in
tl_ucp_dpu_offload.c:19:11: error: â€˜ucp_mem_map_params_tâ€™ {aka â€˜struct ucp_mem_map_paramsâ€™} has no member named â€˜exported_memh_bufferâ€™
   19 |     params.exported_memh_buffer = packed_memh;
      |           ^
make[3]: *** [Makefile:1242: libucc_tl_ucp_la-tl_ucp_dpu_offload.lo] Error 1
make[3]: Leaving directory '/home/master/rastra/rocm_tests/hipmpi/ucc/src/components/tl/ucp'
make[2]: *** [Makefile:1592: install-recursive] Error 1
make[2]: Leaving directory '/home/master/rastra/rocm_tests/hipmpi/ucc/src/components/tl/ucp'
make[1]: *** [Makefile:1409: install-recursive] Error 1
make[1]: Leaving directory '/home/master/rastra/rocm_tests/hipmpi/ucc/src'
make: *** [Makefile:576: install-recursive] Error 1

NOTE: issue is not seen with branch v1.3.x

Test enviromnent:
[...]

PR #1032, opened 2024-10-08
TL/UCP: Check for ucp_memh_pack, if it's not present then disable sliding window and the doca plugin

nsarka:
Patch for https://github.com/openucx/ucc/issues/1031

### DIFF
--- a/config/m4/ucx.m4
+++ b/config/m4/ucx.m4
@@ -8,6 +8,7 @@ UCX_MIN_REQUIRED_MAJOR=1
 UCX_MIN_REQUIRED_MINOR=11
 AS_IF([test "x$ucx_checked" != "xyes"],[
     ucx_happy="no"
+    ucp_memh_happy="no"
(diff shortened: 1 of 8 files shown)


Sergei-Lebedev (commented):
CI error is relevant

janjust (approved):

janjust:
@nsarka Something is not quite right

"configure: error: conditional "HAVE_UCP_MEMH_PACK" was never defined.
Usually this means the macro was only invoked conditionally.
Error: Process completed with exit code 1."

janjust (changes requested):
let's fix the configure error, otherwise, looks good to me

samnordmann (approved):

janjust (approved):

nsarka:
Closing--it was decided in a UCC meeting that instead of merging this PR we will bump the minimum UCX version

PR #1040, opened 2024-10-24
TL/MLX5: add nonblocking cudaMemcpy support

MamziB:
adding support for nonblocking cuda memcpy in mcast

### DIFF
--- a/src/components/tl/mlx5/mcast/tl_mlx5_mcast_progress.c
+++ b/src/components/tl/mlx5/mcast/tl_mlx5_mcast_progress.c
@@ -391,9 +391,10 @@ ucc_status_t ucc_tl_mlx5_mcast_process_packet(ucc_tl_mlx5_mcast_coll_comm_t *com
                                               ucc_tl_mlx5_mcast_coll_req_t *req,
                                               struct pp_packet* pp)
 {
-    ucc_status_t      status = UCC_OK;
-    void             *dest;
-    ucc_memory_type_t mem_type;
+    ucc_status_t                status = UCC_OK;
+    void                       *dest;
+    ucc_ee_executor_task_args_t eargs;
+    ucc_ee_executor_t          *exec;
     ucc_assert(pp->psn >= req->start_psn &&
            pp->psn < req->start_psn + req->num_packets);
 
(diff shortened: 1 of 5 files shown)


janjust (approved):

janjust:
Please fix linter

janjust:
@MamziB rebase, and fix CI, then good to go

janjust (approved):

MamziB:
@Sergei-Lebedev can you please take a look? all the tests now are passing

src/components/tl/mlx5/mcast/tl_mlx5_mcast_progress.c
```diff
@@ -402,18 +403,29 @@ ucc_status_t ucc_tl_mlx5_mcast_process_packet(ucc_tl_mlx5_mcast_coll_comm_t *com
...
+            eargs.copy.src  = (void*) pp->buf;
+            eargs.copy.dst  = dest;
+            eargs.copy.len  = pp->length;
+
+            assert(req->exec_task == NULL);
+            status = ucc_ee_executor_task_post(exec, &eargs, &req->exec_task);
+            if (ucc_unlikely(status != UCC_OK)) {
+                return status;
+            }
+        } else {
```
Sergei-Lebedev, line 427: condition is not needed, executor can copy host buffers as well

MamziB: That is a good question. However, it seems like i had to use ```schedule``` if i wanted to use executors for a task. Please look at `ucc_coll_task_get_executor()` function. As you can see, we need to have ```task->schedule``` ready in order to obtain a ```task->executor```.

src/components/tl/mlx5/tl_mlx5_coll.c
```diff
@@ -31,25 +35,46 @@ ucc_status_t ucc_tl_mlx5_coll_mcast_init(ucc_base_coll_args_t *coll_args,
 
     switch (coll_args->args.coll_type) {
     case UCC_COLL_TYPE_BCAST:
-        status = ucc_tl_mlx5_mcast_bcast_init(task);
+        status = ucc_tl_mlx5_get_schedule(tl_team, coll_args,
```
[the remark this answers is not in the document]
MamziB: That is a good question. However, it seems like i had to use ```schedule``` if i wanted to use executors for a task. Please look at `ucc_coll_task_get_executor()` function. As you can see, we need to have ```task->schedule``` ready in order to obtain a ```task->executor```.

MamziB:
@Sergei-Lebedev thanks for the new comments. please take a look at my responses and let me know your thoughts.

src/components/tl/mlx5/tl_mlx5_coll.c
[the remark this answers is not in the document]
Sergei-Lebedev: In ucc_coll_tag_get_executor we check schedule only if executor is not present in the task. It doesn't mean you should always use schedule when you need executor. Regular task will also work

src/components/tl/mlx5/mcast/tl_mlx5_mcast_progress.c
Sergei-Lebedev: It will not be nonblocking in case of host memory. Do you see perf difference when you use executor vs regular memcpy?

MamziB: oh this is good to know that for host copy it is already blocking, I see not much performance diff as well, I will make this changes accordingly.

MamziB:
@Sergei-Lebedev i made the two requested changes. please take a look

janjust:
bot: retest

Sergei-Lebedev (approved):

MamziB:
@janjust All the checks have passed now.

PR #1046, opened 2024-11-08
CORE: fix coll trace for service team

Sergei-Lebedev:
## What
Fixes segfault when service team finalizes collective and UCC_COLL_TRACE=debug is set

## How ?
TL SHARP and TL MLX5 bypass service_coll interface and use function pointers directly, in this case some fields in request object are not valid and coll trace doesn't work. In this patch appropriate finalize function is used instead


bureddy (approved):

samnordmann (approved):
Thank you for this fix!

manjugv (approved):

PR #1051, opened 2024-12-03
CORE: fix n deps overflow in pipelined schedule

Sergei-Lebedev:
## What
Fixes overflow which happens if number of restarts in pipelined schedule is bigger than 256

### DIFF
--- a/src/schedule/ucc_schedule.h
+++ b/src/schedule/ucc_schedule.h
@@ -107,9 +107,9 @@ typedef struct ucc_coll_task {
         /* used for lf mt progress queue */
         ucc_lf_queue_elem_t            lf_elem;
     };
-    uint8_t                            n_deps;
-    uint8_t                            n_deps_satisfied;
-    uint8_t                            n_deps_base;
+    uint32_t                           n_deps;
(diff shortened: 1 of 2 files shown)


src/schedule/ucc_schedule_pipelined.c
```diff
@@ -256,7 +256,8 @@ ucc_status_t ucc_dependency_handler(ucc_coll_task_t *parent,
     ucc_status_t status;
     uint8_t      n_deps_satisfied;
```
samnordmann, line 257: shouldn't this be `uint32_t` as well?

src/schedule/ucc_schedule.h
```diff
@@ -107,9 +107,9 @@ typedef struct ucc_coll_task {
         /* used for lf mt progress queue */
         ucc_lf_queue_elem_t            lf_elem;
     };
-    uint8_t                            n_deps;
```
samnordmann, line 110: copyright

janjust (approved):

samnordmann (approved):

PR #1067, opened 2025-01-02
TL/MLX5: a2a various optimizations

samnordmann:
# What

This PR contains various optimizations for TL/MLX5/a2a, leading to significant performance gain
before_after

## Support rectangular blocks

this is a critical optimization that brings immediate performance benefits since it gives more flexibility in choosing the block dimension to better saturate the transpose unit. To complete this feature, we expose two (independent) options for determining the block dimensions h and w:
   - `ALLTOALL_BLOCK_SHAPE=long`, imposing `h >= w`
   - `ALLTOALL_BLOCK_SHAPE=wide`, imposing `h <= w`
   - `ALLTOALL_BLOCK_SHAPE=wide`, imposing `h = w`
rect_blocks

## Reuse device memory chunks for several blocks
as long as two blocks need to be sent to the same remote peer, the WQEs dealing with those blocks can 1) be enqueued on the same QPE and 2) use the same device memory chunks. This allows to use one dm chunk to post (and offload to the NIC) the processing of a batch of blocks. This makes the algorithm wait less on free device memory chunks. This option is controlled by the option `NBR_SERIALIZED_BATCHES`.

output5

## batch the inter-node RDMA sends
we allow successive results of the transpose WQE to be batched before being sent to a remote peer. This allows to better saturate the network by aggregating the message. The batch size is controlled by `SEND_BATCH_SIZE`

batch_size

## Iterate across nodes before blocks when posting the WQEs
allows to better saturate the NW. This option is controlled by `NBR_BATCHES_PER_PASSAGE` which sets the number of batches to send to a remote peer before moving to the next one. The old behavior corresponds to large values of this parameter, i.e.,`NBR_BATCHES_PER_PASSAGE >> 1`

batch_per_passage2

## option to force regular case
through the TL/MLX5's env `FORCE_REGULAR`, forcing the chosen block dimensions to divide ppn. This option is useful 1) for debug purposes, and also 2) since the regular case is expected to perform better than the irregular case.

All those optimizations are independent, but we introduce them in a single PR to avoid resolving many conflicts.

### DIFF
--- a/src/components/tl/mlx5/alltoall/alltoall_coll.c
+++ b/src/components/tl/mlx5/alltoall/alltoall_coll.c
@@ -82,7 +82,10 @@ static ucc_status_t ucc_tl_mlx5_poll_cq(struct ibv_cq *cq, ucc_base_lib_t *lib)
         } else {
             ucc_tl_mlx5_dm_chunk_t *dm = (ucc_tl_mlx5_dm_chunk_t *)wcs[i].wr_id;
             dm->task->alltoall.blocks_completed++;
-            ucc_mpool_put(dm);
+            dm->completed_sends++;
+            if (dm->posted_all && dm->completed_sends == dm->posted_sends) {
+                ucc_mpool_put(dm);
+            }
         }
     }
     return UCC_OK;
@@ -379,33 +382,57 @@ static void ucc_tl_mlx5_asr_barrier_progress(ucc_coll_task_t *coll_task)
     }
 }
 
+ucc_tl_mlx5_dm_chunk_t *
+ucc_tl_mlx5_a2a_wait_for_dm_chunk(ucc_tl_mlx5_schedule_t *task)
+{
+    ucc_base_lib_t         *lib  = UCC_TASK_LIB(task);
+    ucc_tl_mlx5_team_t     *team = TASK_TEAM(&task->super);
+    ucc_tl_mlx5_dm_chunk_t *dm   = NULL;
+
+    dm = ucc_mpool_get(&team->dm_pool);
+    while (!dm) {
+        if (UCC_OK != ucc_tl_mlx5_poll_cq(team->a2a->net.cq, lib)) {
+            return NULL;
+        }
+        dm = ucc_mpool_get(&team->dm_pool);
+    }
+    dm->task = task;
+
+    return dm;
+}
+
 // add polling mechanism for blocks in order to maintain const qp tx rx
 static ucc_status_t ucc_tl_mlx5_send_blocks_start(ucc_coll_task_t *coll_task)
 {
-    ucc_tl_mlx5_schedule_t *task           = TASK_SCHEDULE(coll_task);
-    ucc_tl_mlx5_team_t     *team           = SCHEDULE_TEAM(task);
-    ucc_tl_mlx5_alltoall_t *a2a            = team->a2a;
-    int                     node_size      = a2a->node.sbgp->group_size;
-    int                     net_size       = a2a->net.sbgp->group_size;
-    int                     op_msgsize     = node_size * a2a->max_msg_size
-                                                     * UCC_TL_TEAM_SIZE(team)
-                                                     * a2a->max_num_of_columns;
-    int                     node_msgsize   = SQUARED(node_size)
-                                                     * task->alltoall.msg_size;
-    int                     block_size     = task->alltoall.block_size;
-    int                     col_msgsize    = task->alltoall.msg_size
-                                                      * block_size * node_size;
-    int                     block_msgsize  = SQUARED(block_size)
-                                                     * task->alltoall.msg_size;
-    int                     dm_host        = UCC_TL_MLX5_TEAM_LIB(team)
-                                                                 ->cfg.dm_host;
-    ucc_status_t            status         = UCC_OK;
-    ucc_base_lib_t         *lib            = UCC_TASK_LIB(task);
-    int                     seq_index      = task->alltoall.seq_index;
-    int                     i, j, k, rank, dest_rank, cyc_rank;
-    uint64_t                src_addr, remote_addr;
-    ucc_tl_mlx5_dm_chunk_t *dm;
-    uintptr_t               dm_addr;
+    ucc_tl_mlx5_schedule_t *task      = TASK_SCHEDULE(coll_task);
+    ucc_base_lib_t *        lib       = UCC_TASK_LIB(task);
+    ucc_tl_mlx5_team_t *    team      = TASK_TEAM(&task->super);
+    ucc_tl_mlx5_alltoall_t *a2a       = team->a2a;
+    ucc_rank_t              node_size = a2a->node.sbgp->group_size;
+    ucc_rank_t              net_size  = a2a->net.sbgp->group_size;
+    size_t op_msgsize = node_size * a2a->max_msg_size * UCC_TL_TEAM_SIZE(team) *
+                        a2a->max_num_of_columns;
+    size_t       node_msgsize  = SQUARED(node_size) * task->alltoall.msg_size;
+    int          block_h       = task->alltoall.block_height;
+    int          block_w       = task->alltoall.block_width;
+    size_t       col_msgsize   = task->alltoall.msg_size * block_w * node_size;
+    size_t       line_msgsize  = task->alltoall.msg_size * block_h * node_size;
+    size_t       block_msgsize = block_h * block_w * task->alltoall.msg_size;
+    ucc_status_t status        = UCC_OK;
+    int          node_grid_w   = node_size / block_w;
+    int      node_nbr_blocks   = (node_size * node_size) / (block_h * block_w);
+    int      seq_index         = task->alltoall.seq_index;
+    int      block_row         = 0;
+    int      block_col         = 0;
+    uint64_t remote_addr       = 0;
+    ucc_tl_mlx5_dm_chunk_t *dm = NULL;
+    int batch_size = UCC_TL_MLX5_TEAM_LIB(team)->cfg.block_batch_size;
+    int num_serialized_batches =
+        UCC_TL_MLX5_TEAM_LIB(team)->cfg.num_serialized_batches;
+    int num_batches_per_passage =
+        UCC_TL_MLX5_TEAM_LIB(team)->cfg.num_batches_per_passage;
+    int i, j, k, send_to_self, block_idx, rank, dest_rank, cyc_rank, node_idx;
+    uint64_t src_addr;
 
     coll_task->status       = UCC_INPROGRESS;
     coll_task->super.status = UCC_INPROGRESS;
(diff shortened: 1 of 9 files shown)


src/components/tl/mlx5/alltoall/alltoall_coll.c
```diff
@@ -752,26 +783,44 @@ static inline int power2(int value)
...
 
-static inline int block_size_fits(size_t msgsize, int block_size)
+static inline int block_size_fits(size_t msgsize, int height, int width)
 {
-    return block_size * ucc_max(power2(block_size) * msgsize, MAX_MSG_SIZE) <=
-           MAX_TRANSPOSE_SIZE;
+    size_t t1    = power2(ucc_max(msgsize, 8));
+    size_t tsize = height * ucc_max(power2(width) * t1, MAX_MSG_SIZE);
+
+    return tsize <= MAX_TRANSPOSE_SIZE && msgsize <= 128 && height <= 64 &&
```
samnordmann, line 791: define literals and not hardcode

src/components/tl/mlx5/tl_mlx5.c
```diff
@@ -28,6 +28,19 @@ static ucc_config_field_t ucc_tl_mlx5_lib_config_table[] = {
      ucc_offsetof(ucc_tl_mlx5_lib_config_t, dm_buf_num),
      UCC_CONFIG_TYPE_ULUNITS},
 
+    {"FORCE_REGULAR", "y",
+     "Force the regular case where the block dimensions "
```
janjust, line 32: I would word this a bit differently, up to you.

"Enforce the regular case where the block dimensions evenly divide ppn. This option requires BLOCK_SIZE = 0."

src/components/tl/mlx5/tl_mlx5.c
```diff
@@ -104,6 +117,24 @@ static ucc_config_field_t ucc_tl_mlx5_lib_config_table[] = {
      ucc_offsetof(ucc_tl_mlx5_lib_config_t, mcast_conf.one_sided_reliability_enable),
      UCC_CONFIG_TYPE_BOOL},
 
+    {"SEND_BATCH_SIZE", "2",
+     "number of blocks that are transposed "
```
janjust, line 121: We should probably capitalize the start of every option description

src/components/tl/mlx5/tl_mlx5.c
```diff
@@ -104,6 +117,24 @@ static ucc_config_field_t ucc_tl_mlx5_lib_config_table[] = {
...
+     "on the NIC before being sent as a batch to a remote peer",
+     ucc_offsetof(ucc_tl_mlx5_lib_config_t, block_batch_size),
+     UCC_CONFIG_TYPE_UINT},
+
+    {"NBR_SERIALIZED_BATCHES", "4",
+     "number of block batches "
```
janjust, line 127: Same here. We should probably capitalize the start of every option description

src/components/tl/mlx5/alltoall/alltoall_coll.c
janjust: Are you talking about the 128 and 64? If so, yes agreed.

src/components/tl/mlx5/alltoall/alltoall_coll.c
```diff
@@ -786,12 +835,15 @@ UCC_TL_MLX5_PROFILE_FUNC(ucc_status_t, ucc_tl_mlx5_alltoall_init,
                                                         == a2a->node.asr_rank);
     int                     n_tasks   = is_asr ? 5 : 3;
     int                     curr_task = 0;
+    int                       ppn       = tl_team->a2a->node.sbgp->group_size;
```
janjust, line 838: Maybe align assignments and variables?

janjust (approved):
I left minor comments, but overall looks very solid - thanks!

src/components/tl/mlx5/alltoall/alltoall_coll.c
```diff
@@ -752,26 +783,44 @@ static inline int power2(int value)
...
-           MAX_TRANSPOSE_SIZE;
+    size_t t1    = power2(ucc_max(msgsize, 8));
+    size_t tsize = height * ucc_max(power2(width) * t1, MAX_MSG_SIZE);
+
+    return tsize <= MAX_TRANSPOSE_SIZE && msgsize <= 128 && height <= 64 &&
+           width <= 64;
 }
 
-static inline int get_block_size(ucc_tl_mlx5_schedule_t *task)
+static inline void get_block_dimensions(int ppn, int msgsize, int force_regular,
```
Sergei-Lebedev, line 795: maybe do something like this. It will need less number of iteration to converge

```c
static inline void get_block_dimensions(int ppn, int msgsize, int force_regular,
                                        int force_longer, int force_wider,
                                        int *block_height, int *block_width)
{
    size_t max_area = MAX_TRANSPOSE_SIZE / msgsize;
    int h = (int)sqrt(max_area); 
    int w = max_area / h;

    h = (h > 64) ? 64 : (h < 1) ? 1 : h;
    w = (w > 64) ? 64 : (w < 1) ? 1 : w;

    if (force_regular) {
        while (h > 1 && ppn % h != 0) h--;
        while (w > 1 && ppn % w != 0) w--;
    }
    if (force_longer && w > h) {
        int temp = h; h = w; w = temp;
    }
    if (force_wider && h > w) {
        int temp = h; h = w; w = temp;
    }

    while (h > 1 && !block_size_fits(msgsize, h, w)) h--;
    while (w > 1 && !block_size_fits(msgsize, h, w)) w--;

    *block_height = h;
    *block_width  = w;
}
```

src/components/tl/mlx5/alltoall/alltoall_coll.c
samnordmann: yes

src/components/tl/mlx5/alltoall/alltoall_coll.c
samnordmann: clang format did that. It is aligned with the line below. Is it ok or do you suggest something else?

src/components/tl/mlx5/tl_mlx5.c
```diff
@@ -28,6 +28,19 @@ static ucc_config_field_t ucc_tl_mlx5_lib_config_table[] = {
...
+     ucc_offsetof(ucc_tl_mlx5_lib_config_t, force_regular),
+     UCC_CONFIG_TYPE_BOOL},
+
+    {"FORCE_LONGER", "y", "Force the blocks to have more height than width",
+     ucc_offsetof(ucc_tl_mlx5_lib_config_t, force_longer),
+     UCC_CONFIG_TYPE_BOOL},
+
+    {"FORCE_WIDER", "n", "Force the blocks to have more width than height",
+     ucc_offsetof(ucc_tl_mlx5_lib_config_t, force_wider), UCC_CONFIG_TYPE_BOOL},
+
```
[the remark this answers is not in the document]
samnordmann: all the three options are actually orthogonal.

regular/not regular is about dividing ppn,
and wider/longer is understood with closed inequalities, so for example both wider and longer needs to be activated to ensure square blocks.

Does it make sense ? Maybe there is something cleaner but I couldn't figure it out.

src/components/tl/mlx5/alltoall/alltoall_coll.c
```diff
@@ -752,26 +783,44 @@ static inline int power2(int value)
     return p;
 }
 
-static inline int block_size_fits(size_t msgsize, int block_size)
+static inline int block_size_fits(size_t msgsize, int height, int width)
 {
-    return block_size * ucc_max(power2(block_size) * msgsize, MAX_MSG_SIZE) <=
-           MAX_TRANSPOSE_SIZE;
+    size_t t1    = power2(ucc_max(msgsize, 8));
```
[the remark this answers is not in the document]
samnordmann: I don't think so. The goal of this function is to round to the closest power of 2 which is equal or greater than x.
But `ucc_align_up_pow2` takes the power of two as argument, doesn't compute the closest greater power of 2.

But please correct me if I'm wrong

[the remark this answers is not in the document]
samnordmann: btw, there is a typo in the current implementation: we need to start from p=1, not p=2

src/components/tl/mlx5/alltoall/alltoall_coll.c
samnordmann: I'm afraid your proposition is not able to compute the optimal `(h,w)`.

I agree your function is cheaper as it is O(n) while mine is O(n**2), but since n=64 is fixed here, I wouldn't worry too much about the latency (but I might be wrong).

I admit there might be a more clever way to achieve what I want, but unfortunately your proposition doesn't seem to work optimally, as you can see in the following csv comparing the results of the two versions (the current one is called "v1", yours "v2")
[block_dimensions.csv](https://github.com/user-attachments/files/19009743/block_dimensions.csv)

I generated it from the following python script:
```python
import math
import pandas as pd

MAX_MSIZE = 8192
MAX_NCOLS = 64
MAX_NROWS = 64
MAX_MSGSIZE = 128

def round_to_next_power_of_2(n: int) -> int:
    return 2 ** math.ceil(math.log2(n))

def block_size_for_HW_limit(msgsize: int, nrows: int, ncols: int) -> int:
    return nrows * max(round_to_next_power_of_2(ncols) * round_to_next_power_of_2(max(msgsize, 8)), 128)

def block_size_fits(msgsize: int, nrows: int, ncols: int) -> bool:
    return msgsize <= MAX_MSGSIZE and nrows <= MAX_NROWS and ncols <= MAX_NCOLS and block_size_for_HW_limit(msgsize, nrows, ncols) <= MAX_MSIZE

def get_block_dimensions(ppn: int, msgsize: int, force_regular: bool, force_longer: bool, force_wider: bool) -> tuple[int, int]:
    h_best: int = 1
    w_best: int = 1
    for h in range(1, MAX_NCOLS + 1):
        if force_regular and ppn % h != 0:
            continue
        for w in range(1, MAX_NCOLS + 1):
            if (force_regular and ppn % w != 0) or (force_wider and w < h) or (force_longer and w > h):
                continue
            if block_size_fits(msgsize, h, w) and h * w >= h_best * w_best:
                if h * w > h_best * w_best or abs(h / w - 1) < abs(h_best / w_best - 1):
                    h_best = h
                    w_best = w
    return (h_best, w_best)

def get_block_dimensions_v2(ppn: int, msgsize: int, force_regular: bool, force_longer: bool, force_wider: bool) -> tuple[int, int]:
    max_area = MAX_MSIZE / MAX_MSGSIZE
    h = int(math.sqrt(max_area))
    w = int(max_area // h)

    h = max(1, min(h, MAX_NROWS))
    w = max(1, min(w, MAX_NCOLS))

    if force_regular:
        while h > 1 and ppn % h != 0:
            h -= 1
        while w > 1 and ppn % w != 0:
            w -= 1
    if force_longer and w > h:
        h, w = w, h
    if force_wider and h > w:
        h, w = w, h

    while h > 1 and not block_size_fits(msgsize, h, w):
        h -= 1
    while w > 1 and not block_size_fits(msgsize, h, w):
        w -= 1

    return (h, w)

def compute_all_block_dimensions():
    PPN = [2 ** i for i in range(1, 7)]
    MSGSIZE = [2 ** i for i in range(8)]
    FORCE_REGULAR=[False, True]
    FORCE_LONGER=[False, True]
    FORCE_WIDER=[False, True]
    VERSIONS = [("v1", get_block_dimensions), ("v2", get_block_dimensions_v2)]
    results = []
    for ppn in PPN:
        for msgsize in MSGSIZE:
            for force_regular in FORCE_REGULAR:
                for force_longer in FORCE_LONGER:
                    for force_wider in FORCE_WIDER:
                        for v_name, f in VERSIONS:
                            h, w = f(ppn, msgsize, force_regular, force_longer, force_wider)
                            results.append({'version': v_name,
                                            'ppn': ppn,
                                            'msgsize': msgsize,
                                            'force_regular': force_regular,
                                            'force_longer': force_longer,
                                            'force_wider': force_wider,
                                            'h': h,
                                            'w': w,
                                            'h * w': h * w})
    return pd.DataFrame(results)

if __name__ == "__main__":
    results = compute_all_block_dimensions()
    print(results)
    results.to_csv('block_dimensions.csv', index=False)
```

Sergei-Lebedev: ok, but it's possible to cut number of iterations depending on block shape. For instance
1. if regular then max block size is ppn
2. if force longer than max w is h and vice versa
3. also min w should be equal or greater  than (h_best * w_best) / h

src/components/tl/mlx5/alltoall/alltoall_coll.c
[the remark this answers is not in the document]
Sergei-Lebedev: right, pls move it to ucc_math.h

src/components/tl/mlx5/tl_mlx5.c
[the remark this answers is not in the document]
samnordmann: ok got it, I made the change as you suggest

src/components/tl/mlx5/alltoall/alltoall_coll.c
samnordmann: right, fixed

Sergei-Lebedev (approved):

PR #1073, opened 2025-02-06
UTIL: fix names array memory leak

MamziB:
we see memory leaks from CI tests in https://github.com/openucx/ucc/pull/1023

This PR fixes these memory leaks. One instance of leaks can happen when a function such ```ucc_cl_basic_get_lib_attr()``` gets called multiple times, which can potentially lead to calling ```ucc_config_names_array_dup``` on the same cl_lib object. Each time this ```duplicate``` function is called, we were allocating new memory for names array, without making sure that that array has already been allocated or not.

### DIFF
--- a/src/utils/ucc_parser.c
+++ b/src/utils/ucc_parser.c
@@ -238,6 +238,9 @@ ucc_status_t ucc_config_names_array_dup(ucc_config_names_array_t *dst,
 {
     int i;
 
+    if (dst->count != 0) {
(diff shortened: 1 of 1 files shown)


janjust (approved):

janjust:
@Sergei-Lebedev everything is passing, can we get this in?

Sergei-Lebedev (approved):

PR #1084, opened 2025-02-28
TL/SHARP: enable reduce-scatter only when SAT is enabled.

bureddy:
Enable SHARP reduce-scatter only when SAT is enabled


Sergei-Lebedev (approved):
please fix commit title

janjust (approved):
Looks good!

PR #1092, opened 2025-03-11
BUILD: Add support for optional filter argument in config parser print function

Sergei-Lebedev:
## What
This commit adds compatibility for a newer version of UCX's config parser print function by conditionally adding an optional NULL filter argument. The changes include:

- Modifying config/m4/ucx.m4 to check for the new function signature
- Updating src/utils/ucc_parser.c and src/utils/ucc_parser.h to conditionally call the function with an additional NULL argument
- Using preprocessor macros to maintain backward compatibility

### DIFF
--- a/src/utils/ucc_parser.h
+++ b/src/utils/ucc_parser.h
@@ -51,6 +51,14 @@ typedef struct ucc_file_config ucc_file_config_t;
     };
 #endif
 
+#ifdef UCS_HAVE_PARSER_PRINT_FILTER_ARG
+#define UCS_CONFIG_PARSER_PRINT_OPTS(_stream, _title, _opts, _fields, _tprefix, _prefix, _flags) \
+    ucs_config_parser_print_opts((_stream), (_title), (_opts), (_fields), (_tprefix), (_prefix), (_flags), NULL)
+#else
+#define UCS_CONFIG_PARSER_PRINT_OPTS(_stream, _title, _opts, _fields, _tprefix, _prefix, _flags) \
+    ucs_config_parser_print_opts((_stream), (_title), (_opts), (_fields), (_tprefix), (_prefix), (_flags))
+#endif
+
 #define UCC_CONFIG_GET_TABLE(_table)    &_table##_config_entry
 #define UCC_CONFIG_TYPE_LOG_COMP        UCS_CONFIG_TYPE_LOG_COMP
 #define UCC_CONFIG_REGISTER_TABLE       UCS_CONFIG_REGISTER_TABLE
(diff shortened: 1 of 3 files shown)


janjust (approved):

janjust:
@MamziB please review asap

MamziB:
Why not update the code to the latest UCX API instead of using all these if/else statements? I would rather this and users who want to use the latest UCC version must also update their UCX version.

If we don’t want to force users to update their UCX version, I’d rather reduce the use of #ifdef and #else in the code.

Instead, we can create a simple MACRO like this and use it instead of calling the actual function:

```
#ifdef UCS_HAVE_PARSER_PRINT_FILTER_ARG
#define UCC_UCS_CONFIG_PARSER_PRINT_OPTS(stream, title, opts, fields, tprefix, prefix, flags) \
    ucs_config_parser_print_opts((stream), (title), (opts), (fields), (tprefix), (prefix), (flags), NULL)
#else
#define UCC_UCS_CONFIG_PARSER_PRINT_OPTS(stream, title, opts, fields, tprefix, prefix, flags) \
    ucs_config_parser_print_opts((stream), (title), (opts), (fields), (tprefix), (prefix), (flags))
#endif

```

src/utils/ucc_parser.h
```diff
@@ -211,8 +211,13 @@ static inline void ucc_config_parser_print_opts(FILE *stream, const char *title,
     ucs_config_print_flags_t ucs_flags;
 
     ucs_flags = ucc_print_flags_to_ucs_print_flags(flags);
+#ifdef UCS_HAVE_PARSER_PRINT_FILTER_ARG
```
MamziB, line 214: Why not update the code to the latest UCX API instead of using all these if/else statements? I would rather this and users who want to use the latest UCC version must also update their UCX version.

If we don’t want to force users to update their UCX version, I’d rather reduce the use of #ifdef and #else in the code.

Instead, we can create a simple MACRO like this and use it instead of calling the actual function:

```
#ifdef UCS_HAVE_PARSER_PRINT_FILTER_ARG
#define UCC_UCS_CONFIG_PARSER_PRINT_OPTS(stream, title, opts, fields, tprefix, prefix, flags) \
    ucs_config_parser_print_opts((stream), (title), (opts), (fields), (tprefix), (prefix), (flags), NULL)
#else
#define UCC_UCS_CONFIG_PARSER_PRINT_OPTS(stream, title, opts, fields, tprefix, prefix, flags) \
    ucs_config_parser_print_opts((stream), (title), (opts), (fields), (tprefix), (prefix), (flags))
#endif

```

samnordmann (approved):

src/utils/ucc_parser.h
Sergei-Lebedev: I added macro as you suggested, however dropping support for older UCX versions is out of scope of the discussion here. We might do this but it should be agreed on WG meeting

PR #1050, opened 2024-12-02
CL/HIER: Add allgatherv

nsarka:
This PR adds CL/HIER allgatherv and updates the allgatherv gtest to test non-contiguous dst buffers. It is meant to be used in conjunction with my TL/SHM gatherv implementation.

The algorithm is:
- Node-level gatherv, all sendbufs on the node are put into a scratch buffer
- Leader-level allgatherv on the scratch buffer, the result is put into the dst buffer
- Node-level in-place bcast on the dst buffer
- If the buffer is non-contig, unpack by memcpying the packed buffer into the right displacements

## Comparison Data

2 nodes, 32 PPN on lego-grace cg (1 socket, 72 cores/socket)

Size | UCC after PR (cl_hier + tl_shm + tl_ucp) | UCC before PR (cl_basic + tl_ucp) | HCOLL | %-Gain, After vs. HCOLL | %-Gain, Old vs. HCOLL | %-Gain, Before vs. After
-- | -- | -- | -- | -- | -- | --
1 | 7.37 | 19.88 | 7.41 | 0.54% | -168.29% | 62.93%
2 | 7.81 | 19.75 | 7.47 | -4.55% | -164.39% | 60.46%
4 | 7.62 | 19.79 | 7.65 | 0.39% | -158.69% | 61.50%
8 | 8.35 | 19.78 | 8.13 | -2.71% | -143.30% | 57.79%
16 | 9.49 | 19.8 | 9.44 | -0.53% | -109.75% | 52.07%
32 | 9.34 | 22.91 | 9.67 | 3.41% | -136.92% | 59.23%
64 | 9.17 | 22.96 | 10.03 | 8.57% | -128.91% | 60.06%
128 | 10.47 | 27.55 | 10.88 | 3.77% | -153.22% | 62.00%
256 | 12.68 | 33.37 | 12.46 | -1.77% | -167.82% | 62.00%
512 | 19.57 | 180 | 25.31 | 22.68% | -611.18% | 89.13%
1024 | 25.38 | 189.21 | 308.33 | 91.77% | 38.63% | 86.59%
2048 | 38.86 | 198.27 | 311.55 | 87.53% | 36.36% | 80.40%

PR #1098, opened 2025-03-19
CORE: Get node sbgps and leaders

nsarka:
This is a followup PR to https://github.com/openucx/ucc/pull/1050.

It adds two fields to `ucc_topo_t`, `node_leaders` and `all_nodes` and corresponding functions to calculate them on demand. `all_nodes` is the node-sbgp equivalent of `all_sockets`, i.e. it's an array of every node subgroup in the comm.

Knowing the node subgroup mapping on other nodes than the one I'm on is useful in the allgatherv unpack stage from my PR. For example, say you ran with `mpirun --map-by node ...`. Then the node/rank mapping might look like:

```
node0: {rank0, rank2}
node1: {rank1, rank3}
```

Then the destination buffer after the allgatherv is done on each node will look like `{0,2,1,3}`, when it really should be in order of the ranks in the team, like `{0,1,2,3}`. To fix this, I will unpack the elements according to their node-level rank mapping from `all_nodes` once this PR goes in.

My justification for adding `topo->node_leaders` is that the topo is a much better place to calculate and store what each rank's node leader is. In my allgatherv PR, I was doing it in `ucc_cl_hier_allgatherv_init`. I will remove that and use this once it goes in.

### DIFF
--- a/src/components/topo/ucc_topo.c
+++ b/src/components/topo/ucc_topo.c
@@ -374,7 +374,7 @@ ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo, ucc_rank_t **node_leade
     ucc_subset_t *set = &topo->set;
     ucc_rank_t size = ucc_subset_size(set);
     ucc_rank_t nnodes = topo->topo->nnodes;
-    ucc_rank_t i;
+    ucc_rank_t i, j;
     ucc_rank_t *ranks_seen_per_node;
     ucc_rank_t *per_node_leaders;
     ucc_rank_t *node_leaders;
@@ -430,8 +430,8 @@ ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo, ucc_rank_t **node_leade
         ucc_rank_t ctx_rank = ucc_ep_map_eval(set->map, i);
         ucc_host_id_t current_host = topo->topo->procs[ctx_rank].host_id;
 
-        for (ucc_rank_t j = 0; j < nnodes; j++) {
-            ucc_rank_t ctx_rank_j = ucc_ep_map_eval(set->map, j);
+        for (j = 0; j < nnodes; j++) {
+            ucc_rank_t ctx_rank_j = ucc_ep_map_eval(set->map, per_node_leaders[j]);
             ucc_host_id_t current_host_j = topo->topo->procs[ctx_rank_j].host_id;
 
             if (current_host_j == current_host) {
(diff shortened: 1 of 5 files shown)


src/components/topo/ucc_topo.c
```diff
@@ -270,3 +283,167 @@ ucc_status_t ucc_topo_get_all_numas(ucc_topo_t *topo, ucc_sbgp_t **sbgps,
...
+            per_node_leaders[current_host] = i;
+        }
+    }
+
+    /* Second pass: propagate node leaders to all ranks */
+    for (i = 0; i < size; i++) {
+        ucc_rank_t ctx_rank = ucc_ep_map_eval(set->map, i);
+        ucc_host_id_t current_host = topo->topo->procs[ctx_rank].host_id;
+
+        for (j = 0; j < nnodes; j++) {
```
nsarka, line 433: Instead of this being a for loop, can it be a hash table with the key being host_id and the value being the leader number? Does ucx have something for this?

src/components/topo/ucc_topo.c
```diff
@@ -270,3 +284,168 @@ ucc_status_t ucc_topo_get_all_numas(ucc_topo_t *topo, ucc_sbgp_t **sbgps,
...
+    for (i = 0; i < nnodes; i++) {
+        ucc_rank_t ldr_team_rank = ucc_ep_map_eval(leader_sbgp->map, i);
+        topo->set.myrank = ldr_team_rank;
+        if (ucc_rank_on_local_node(myrank, topo)) {
+            /* Skip creating sbgp, we have this node's sbgp already */
+            sbgps[i] = *(ucc_topo_get_sbgp(topo, UCC_SBGP_NODE));
+            leader_rank = topo->node_leader_rank;
+        } else {
+            status = ucc_sbgp_create_node(topo, &sbgps[i]);
+            if (status == UCC_ERR_NO_MESSAGE) {
```
Sergei-Lebedev, line 328: i would change status to UCC_ERR_NOT_FOUND maybe

src/components/topo/ucc_topo.c
```diff
@@ -270,3 +284,168 @@ ucc_status_t ucc_topo_get_all_numas(ucc_topo_t *topo, ucc_sbgp_t **sbgps,
...
+                /* ucc_sbgp_create_node returned because 0 == node_size */
+                sbgps[i].status = UCC_SBGP_NOT_EXISTS;
+                continue;
+            } else if (status != UCC_OK) {
+                ucc_error("failed to create all_node subgroup %d", i);
```
Sergei-Lebedev, line 333: need to reset myrank back to what it was always when exit function

Sergei-Lebedev (approved):

janjust (approved):

PR #1100, opened 2025-03-21
TEST: add buffers for onesided tests

wfaderhold21:
## What
Adds extra src/dst/work buffers for onesided algorithms in the MPI test.

## Why ?
These are necessary to ensure RDMA writes are not performed on the next test before all processes have completed the initial test.

### DIFF
--- a/test/mpi/test_mpi.cc
+++ b/test/mpi/test_mpi.cc
@@ -72,11 +72,11 @@ UccTestMpi::UccTestMpi(int argc, char *argv[], ucc_thread_mode_t _tm,
         if (with_onesided) {
             onesided_ctx_params = ctx_params;
             for (auto i = 0; i < UCC_TEST_N_MEM_SEGMENTS; i++) {
-                onesided_buffers[i] = ucc_calloc(UCC_TEST_MEM_SEGMENT_SIZE,
+                onesided_buffers[i] = ucc_calloc(UCC_TEST_MEM_SEGMENT_SIZE * 2,
                                                  size, "onesided buffers");
                 UCC_MALLOC_CHECK(onesided_buffers[i]);
                 segments[i].address = onesided_buffers[i];
-                segments[i].len     = UCC_TEST_MEM_SEGMENT_SIZE * size;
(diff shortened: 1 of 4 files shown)


janjust (approved):

samnordmann (commented):
Thanks @wfaderhold21 
We briefly discussed this PR during our WG. Can you please explain what precisely this patch aims to fix and why we need both a barrier and doubling the buffers?

test/mpi/test_mpi.cc
```diff
@@ -523,6 +523,9 @@ std::vector<ucc_test_mpi_result_t> UccTestMpi::exec_tests(
             }
         } while (num_done != tcs.size());
         for (auto tc: tcs) {
+            if (tc->args.flags & UCC_COLL_ARGS_FLAG_MEM_MAPPED_BUFFERS) {
+                MPI_Barrier(MPI_COMM_WORLD);
```
wfaderhold21, line 527: confirm if this is needed or if just need an extra barrier before the test and check

wfaderhold21: can confirm that only need the barrier. not the extra buffers. However, the barrier should be prior to collective start.

wfaderhold21:
> Thanks @wfaderhold21 We briefly discussed this PR during our WG. Can you please explain what precisely this patch aims to fix and why we need both a barrier and doubling the buffers?

Sure. 

The aim of this patch is to fix a potential issue when running the MPI test when multiple transports (i.e., RC + XPMEM) are being used. This could allow for one process (e.g., rank 0) to finish the collective and move to the next collective before another process (e.g., rank 1) has finished the check on its destination buffer. 

The use of multiple buffers was to allow for processes to alternate between the buffers and prevent overwriting of a destination buffer because onesided operations are being used. However, this could be simply reduced to a barrier before collective start, which will prevent a process from entering the next collective while another process is checking the result of the previous collective.

wfaderhold21:
@janjust are we including this for 1.5?

janjust:
yes - but we have several critical PRs to go through before, we will do one rebase at a time until we can merge them all.

Sergei-Lebedev (approved):

Sergei-Lebedev:
ok to test

PR #1101, opened 2025-03-23
TL/UCP: fix inplace checking in service allgather

ikryukov:
## What
Fix checking if allgather should use the in-place version or not, by checking the **logical ranks** of the team. (No need to remap to physical ranks.)
## Why ?
In situations involving team odd-even, this results in a problem where, for instance, the tl/cuda cannot communicate.
This issue starts to become visible after fixing another issue in #1089.


ikryukov:
ok to test

Sergei-Lebedev (approved):

Sergei-Lebedev:
ok to test

janjust (approved):

PR #1103, opened 2025-03-26
CL/HIER: Allgatherv node sbgp unpack

nsarka:
This PR is a followup to https://github.com/openucx/ucc/pull/1098 and https://github.com/openucx/ucc/pull/1050.

In my previous allgatherv PR, there was an issue in the case that the ranks in each node subgroup are not contiguous in the team rank order. For example, say you ran with mpirun --map-by node .... Then the node/rank mapping might look like:

```
node0: {rank0, rank2}
node1: {rank1, rank3}
```

Then the destination buffer after the allgatherv is done on each node will look like {0,2,1,3}, when it really should be in order of the ranks in the team, like {0,1,2,3}.

To fix this, in this PR, I unpack the elements according to their node-level rank mapping from all_nodes.

### DIFF
--- a/src/components/cl/hier/allgatherv/allgatherv.c
+++ b/src/components/cl/hier/allgatherv/allgatherv.c
@@ -52,7 +52,7 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
     ucc_status_t        status;
 
     ucc_assert(team_rank >= 0 && team_rank < UCC_CL_TEAM_SIZE(cl_team));
-    ucc_assert(SBGP_ENABLED(cl_team, NODE_LEADERS));
+    ucc_assert(SBGP_EXISTS(cl_team, NODE_LEADERS));
 
     status = ucc_topo_get_node_leaders(core_team->topo, &node_leaders, NULL);
     if (UCC_OK != status) {
@@ -151,10 +151,10 @@ UCC_CL_HIER_PROFILE_FUNC(ucc_status_t, ucc_cl_hier_allgatherv_init,
     }
     node_gathered_data = NULL;
 
-    /* If I'm a node leader, calculate leader_counts, leader_disps, and set the
+    /* If node ldr sbgp, calculate leader_counts, leader_disps, and set the
        dst buffer of the gatherv to the right displacements for the in-place
-       node-leader allgatherv */
-    if(SBGP_ENABLED(cl_team, NODE) && SBGP_ENABLED(cl_team, NODE_LEADERS)) {
+       node-leader allgatherv, even on non-node-leader ranks */
+    if(SBGP_ENABLED(cl_team, NODE) && SBGP_EXISTS(cl_team, NODE_LEADERS)) {
         /* Sum up the counts on each node to get the count for each node leader */
         for (i = 0; i < team_size; i++) {
             UCC_CHECK_GOTO(
--- a/src/components/cl/hier/allgatherv/allgatherv.c
+++ b/src/components/cl/hier/allgatherv/allgatherv.c
@@ -46,42 +46,21 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
                                             ucc_rank_t       team_rank,
                                             ucc_rank_t      *rank_out)
 {
-    ucc_cl_hier_team_t *cl_team       = ucc_derived_of(team, ucc_cl_hier_team_t);
-    ucc_team_t         *core_team     = team->params.team;
-    ucc_rank_t          team_size     = UCC_CL_TEAM_SIZE(cl_team);
-    ucc_rank_t          ldr_sbgp_size = SBGP_SIZE(cl_team, NODE_LEADERS);
-    ucc_rank_t          i, j;
+    ucc_cl_hier_team_t *cl_team   = ucc_derived_of(team, ucc_cl_hier_team_t);
+    ucc_team_t         *core_team = team->params.team;
+    ucc_rank_t         *node_leaders = NULL;
+    ucc_status_t        status;
 
-    ucc_assert(team_rank >= 0 && team_rank < team_size);
+    ucc_assert(team_rank >= 0 && team_rank < UCC_CL_TEAM_SIZE(cl_team));
     ucc_assert(SBGP_ENABLED(cl_team, NODE_LEADERS));
 
-    /* Allocate and populate node_leaders and leader_list */
-    if (ucc_unlikely(cl_team->node_leaders == NULL)) {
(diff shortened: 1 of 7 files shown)


src/components/cl/hier/allgatherv/unpack.c
```diff
@@ -68,48 +68,89 @@ ucc_status_t ucc_cl_hier_allgatherv_unpack_start(ucc_coll_task_t *task)
...
+        
+        // Find the position of this leader in the node_leaders_sbgp
+        node_leader_idx = ucc_ep_map_local_rank(node_leaders_sbgp->map, leader_team_rank);
+        
+        // Get source displacement from leader_displacements (passed in src.info_v)
+        src_rank_disp = ucc_coll_args_get_displacement(
+            args, args->src.info_v.displacements, node_leader_idx);
+            
+        // Find the correct node for this rank
+        for (node_id = 0; node_id < n_nodes; node_id++) {
```
nsarka, line 125: I think this loop can be removed so long as the ranks in the node leader subgroup mapping are in the same order as they are in the team. I've seen this isn't the case though. @Sergei-Lebedev do you think I should try to hunt down the cause and fix it?

src/components/topo/ucc_topo.h
```diff
@@ -253,9 +254,10 @@ static inline ucc_rank_t ucc_topo_nnodes(ucc_topo_t *topo)
...
 }
 
-/* Returns an array mapping each rank to its node leader.
-   The array is cached in topo->node_leaders. */
-ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo,
-                                       ucc_rank_t **node_leaders_out);
+/* Returns node leaders array - array that maps each rank to the TEAM RANK that 
+   is the leader of that rank's node. Also returns per-node leaders array - array
+   mapping node_id to the TEAM RANK of that node's leader */
+ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo, ucc_rank_t **node_leaders_out,
```
nsarka, line 260: Here I updated the previous PR that defined this function to also take in an out-argument for the node leaders. The rest of the functionality is the same.

test/gtest/core/test_topo.cc
```diff
@@ -619,6 +619,7 @@ UCC_TEST_F(test_topo, node_leaders)
     ucc_subset_t     set;
     ucc_rank_t       i;
     ucc_rank_t      *node_leaders;
+    ucc_rank_t      *per_node_leaders;
```
nsarka, line 622: In this file I update the test to take into account per_node_leaders.

src/components/cl/hier/allgatherv/allgatherv.c
```diff
@@ -46,42 +46,21 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
...
-        }
+    ucc_cl_hier_team_t *cl_team   = ucc_derived_of(team, ucc_cl_hier_team_t);
+    ucc_team_t         *core_team = team->params.team;
+    ucc_rank_t         *node_leaders = NULL;
+    ucc_status_t        status;
+
+    ucc_assert(team_rank >= 0 && team_rank < UCC_CL_TEAM_SIZE(cl_team));
+    ucc_assert(SBGP_EXISTS(cl_team, NODE_LEADERS));
+
+    status = ucc_topo_get_node_leaders(core_team->topo, &node_leaders, NULL);
```
nsarka, line 57: Here I use the func I introduced in https://github.com/openucx/ucc/pull/1098, and get rid of what that func replaces.

src/components/cl/hier/allgatherv/unpack.c
Sergei-Lebedev: can you pls elaborate in what case this loop can be removed? you still need to find correct src_rank_disp

nsarka: I removed the loop and added an assert. The loop was just doing what ep_map_local_rank does in the updated version

src/components/topo/ucc_topo.c
```diff
@@ -445,8 +461,13 @@ ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo, ucc_rank_t **node_leade
     }
 
     topo->node_leaders = node_leaders;
+    topo->per_node_leaders = per_node_leaders;
+    //NOLINTNEXTLINE
     *node_leaders_out = node_leaders;
+    if (per_node_leaders_out) {
+        //NOLINTNEXTLINE
+        *per_node_leaders_out = per_node_leaders;
```
samnordmann, line 469: why is it necessary if the caller already gets it through `topo->per_node_leaders` ?

src/components/topo/ucc_topo.c
```diff
@@ -380,7 +384,8 @@ ucc_status_t ucc_topo_get_all_nodes(ucc_topo_t *topo, ucc_sbgp_t **sbgps,
     return status;
 }
 
-ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo, ucc_rank_t **node_leaders_out)
+ucc_status_t ucc_topo_get_node_leaders(ucc_topo_t *topo, ucc_rank_t **node_leaders_out,
+                                      ucc_rank_t **per_node_leaders_out)
```
samnordmann, line 388: if topo gets freed then `per_node_leaders_out` becomes a dangling pointer, right?

samnordmann (approved):
LGTM!

artemry-nv:
bot:retest

artemry-nv:
bot:retest

Sergei-Lebedev (approved):

src/components/cl/hier/allgatherv/allgatherv.c
```diff
@@ -46,42 +46,63 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
...
+    }
+
+    *rank_out = node_leaders[team_rank];
+    return UCC_OK;
+}
+
+/* Check if the ranks are block ordered. If they aren't, we'll need to 
+   unpack the data after the allgatherv into the right position, even if the
+   dst buffer is contiguous */
+static inline ucc_status_t is_block_ordered(ucc_cl_hier_team_t *cl_team, int *ordered)
```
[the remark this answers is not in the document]
nsarka: It turns out this does not work--it should always return true if ranks are block ordered among the nodes. It does not matter what the distribution of ranks is within the node. But for the `UCC_SBGP_FULL_HOST_ORDERED == identity` check, it will return false in the case that the ranks within the node are round-robin ordered over all of the sockets.

PR #1105, opened 2025-04-01
UTIL: Optimized ucc_round_up_power2 function

ikryukov:
## What
Optimized version of `ucc_lowest_greater_power2()` using bitwise operations

## Why ?

- constant time
- mul free only bitwise ops
- prevent overflow

References: 
https://stackoverflow.com/questions/364985/algorithm-for-finding-the-smallest-power-of-two-thats-greater-or-equal-to-a-giv 
https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2

### DIFF
--- a/src/utils/ucc_math.h
+++ b/src/utils/ucc_math.h
@@ -146,7 +146,7 @@ static inline int ucc_lowest_greater_power2(int value)
     if (value <= 1) {
         return 1;
     }
-    return 1 << (32 - ucc_count_leading_zero_bits(value));
+    return 1 << (32 - ucc_count_leading_zero_bits(value - 1));
 }
 
 #endif
--- a/src/utils/ucc_math.h
+++ b/src/utils/ucc_math.h
@@ -141,7 +141,7 @@ static inline uint32_t ucc_ilog2_ceil(uint32_t n)
 }
 
 /* return the lowest greater or equal power of 2 */
-static inline int ucc_lowest_greater_power2(int value)
+static inline int ucc_round_up_power2(int value)
 {
     if (value <= 1) {
         return 1;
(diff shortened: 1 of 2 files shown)


artemry-nv:
bot:retest

src/utils/ucc_math.h
```diff
@@ -143,11 +143,22 @@ static inline uint32_t ucc_ilog2_ceil(uint32_t n)
...
+        return 1;
     }
-    return p;
+
+    // Handle integer overflow case
+    if (value > (INT32_MAX / 2)) {
+        return INT32_MAX;  
+    }
+
+    --value;  // pow2 case
```
samnordmann, line 155: I am not sure to understand the comment

samnordmann (approved):
Looks good, thanks!
nit: I wonder if there could be an even more optimized implementation with `lzcnt`

ikryukov:
> Looks good, thanks! nit: I wonder if there could be an even more optimized implementation with `lzcnt`
like this: 
```
int shift = 31 - __builtin_clz(value - 1);
return 1 << (shift + 1);
```

src/utils/ucc_math.h
ikryukov: corner case if value is already power of two?

ikryukov: I see, so for we need calculate value **strictly** greater than input, fo 4 it should be 8

samnordmann: Do as you think is best -- personally I would leave without any comment or would write "decreasing by 1 is necessary to handle the case where value is a power of 2"

samnordmann (approved):

src/utils/ucc_math.h
samnordmann: nvm that's outdated

janjust (approved):
Thanks!

PR #1096, opened 2025-03-17
TL/UCP: transition to barrier for sync for onesided a2a

wfaderhold21:
## What
Switch from using pSync array with atomic increment to TL/UCP barrier for synchronization

## Why ?
There are multiple reason to switch to this: (1) knomial barrier scales better and has better performance than atomic increment and (2) there can be instances where processes leave the alltoall collective before remote writes have been completed.

PR #1107, opened 2025-04-02
TL/UCP: add cong. avoidant onesided alltoall

wfaderhold21:
## What
The current onesided alltoall algorithm in TL/UCP can suffer from endpoint congestion in particular situations (e.g., high PPN counts in jobs). This algorithm adds a congestion avoidant algorithm allowing users to supply bandwidth in MB/s and a percentage of bandwidth to use for a particular alltoall collective.

### DIFF
--- a/src/components/tl/ucp/Makefile.am
+++ b/src/components/tl/ucp/Makefile.am
@@ -24,12 +24,13 @@ allgatherv =                        \
 	allgatherv/allgatherv_ring.c    \
 	allgatherv/allgatherv_knomial.c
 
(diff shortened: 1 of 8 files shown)


wfaderhold21:
from WG comments:

- Merge code with PR #1096: add into original onesided alltoall algorithm instead of a new algorithm
- Remove ONESIDED_CA_RATE environment variable, obtain through UCP
- Remove ONESIDED_CA_PPN environment variable, obtain through topology

wfaderhold21:
combined with #1096

PR #1108, opened 2025-04-03
TEST: check device version in transpose test

samnordmann:
## What
Skip transpose test if the NIC is not CX7

## Why ?
https://redmine.mellanox.com/issues/4280707


janjust (approved):
LGTM!

PR #1110, opened 2025-04-03
UTIL: ucc_assume macro

ikryukov:
## What
Introduce `ucc_assume` macro to fix clang tidy warnings
## Why ?
clang tidy reports errors division by zero, compiler unable to ensure that some variables couldn't be 0


janjust (approved):

Sergei-Lebedev (approved):
I would change UCC_ASSUME to ucc_assume to make it same style as ucc_likely and ucc_unlikely

PR #1113, opened 2025-04-07
TL/MLX5: ctx global status check

Sergei-Lebedev:
## What
Added checks for successful creation of the rcache and topology initialization.

## Why ?
All the resource initialization must happen in context create epilog. Otherwise if context create fails on one rank the other ranks will hang in context create epilog broadcast

### DIFF
--- a/src/components/tl/mlx5/tl_mlx5_context.c
+++ b/src/components/tl/mlx5/tl_mlx5_context.c
@@ -80,15 +80,8 @@ UCC_CLASS_INIT_FUNC(ucc_tl_mlx5_context_t,
         return status;
     }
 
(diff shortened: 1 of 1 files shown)


samnordmann (approved):
Thank you for the patch!
Not a request, but for readability, I would suggest to move topo and rcache init out of 
`ucc_tl_mlx5_context_ib_ctx_pd_setup` and place it directly in `ucc_tl_mlx5_context_create_epilog`
But if this causes too much refactoring, we probably shouldn't bother

janjust (approved):

B-a-S:
bot:retest

PR #1123, opened 2025-04-16
CL/HIER: Add flag for nonroot info

nsarka:
Adds flag for cl/hier to tell tl/shm gatherv nonroots they can check coll_args->dst.info_v to determine max msg size

### DIFF
--- a/src/components/base/ucc_base_iface.h
+++ b/src/components/base/ucc_base_iface.h
@@ -163,7 +163,9 @@ typedef struct ucc_base_team_iface {
 } ucc_base_team_iface_t;
 
 enum {
-    UCC_BASE_CARGS_MAX_FRAG_COUNT = UCC_BIT(0)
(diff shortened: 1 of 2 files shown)


janjust (approved):

Sergei-Lebedev:
why do we need to introduce another flag and not use bargs.max_frag_count?

nsarka:
> why do we need to introduce another flag and not use bargs.max_frag_count?

It would work, but the name implies this is specifically for frag count. In the allgatherv I'm not using any frag pipelining

nsarka:
Closing, I will just reuse it even though in my opinion the name might be confusing.

nsarka:
@Sergei-Lebedev I updated it so that it has a new CARGS mask bit flag. I think this is the proper solution.

Sergei-Lebedev (approved):

PR #1129, opened 2025-05-05
TOPO: Node ldr ordered by team

nsarka:
## What

This PR sorts the node leader subgroup rank mapping by team rank to avoid unnecessary unpacking in allgatherv. It also removes the `is_host_ordered` function in the allgatherv code, because it was previously checking whether or not the leaders were sorted by team rank. Now it always will be.

## Why ?

Previously the node leader subgroup was sorted by hostid, meaning that if a hostname happened to be lower than another hostname by the comparator that UCC uses, the node leader allgatherv step in my cl/hier allgatherv might actually be incorrect and need the extra unpack step to rearrange the buffer to have the right order.

Previous PR for reference here: https://github.com/openucx/ucc/pull/1103

## How ?
Sorts the rank mapping and updates group_rank. In addition to removing `is_host_ordered`, I also updated `is_block_ordered`, which previously checked if the FULL_ORDERED subgroup was the same as the current team. This  was a tighter restriction than it needed to be, since FULL_ORDERED also sorts by the socket, and the allgatherv only cares whether ranks are block ordered on each node.

## Data

https://docs.google.com/spreadsheets/d/1icnHUZFCYvThZ_bsZ3PS3fCL5HLG58fGrpViF25dHDY/edit?usp=sharing

### DIFF
--- a/src/components/cl/hier/allgatherv/allgatherv.c
+++ b/src/components/cl/hier/allgatherv/allgatherv.c
@@ -69,53 +69,48 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
    dst buffer is contiguous */
 static inline ucc_status_t is_block_ordered(ucc_cl_hier_team_t *cl_team, int *ordered)
 {
-    ucc_topo_t *topo = cl_team->super.super.params.team->topo;
-    ucc_sbgp_t *sbgp;
-    int         is_block_ordered;
+    ucc_topo_t  *topo             = cl_team->super.super.params.team->topo;
+    ucc_sbgp_t  *all_nodes        = NULL;
+    int          is_block_ordered = 1;
+    int          n_nodes;
+    ucc_status_t status;
+    ucc_rank_t   prev_rank, curr_rank;
+    ucc_rank_t   i, j;
 
     if (cl_team->is_block_ordered != -1) {
         is_block_ordered = cl_team->is_block_ordered;
     } else {
-        sbgp = ucc_topo_get_sbgp(topo, UCC_SBGP_FULL_HOST_ORDERED);
-        is_block_ordered = ucc_ep_map_is_identity(&sbgp->map) ? 1 : 0;
+        status = ucc_topo_get_all_nodes(topo, &all_nodes, &n_nodes);
+        if (status != UCC_OK) {
+            return status;
+        }
+
+        // For each node, check if its ranks are consecutive in team order
+        for (i = 0; i < n_nodes; i++) {
+            if (all_nodes[i].status == UCC_SBGP_ENABLED
+                && all_nodes[i].group_size > 1) {
+                // Get first rank in this node
+                prev_rank = ucc_ep_map_eval(all_nodes[i].map, 0);
+
+                // Check if all other ranks are consecutive
+                for (j = 1; j < all_nodes[i].group_size; j++) {
+                    curr_rank = ucc_ep_map_eval(all_nodes[i].map, j);
+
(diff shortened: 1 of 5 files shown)


src/components/cl/hier/allgatherv/allgatherv.h
```diff
@@ -17,7 +17,7 @@ enum
 extern ucc_base_coll_alg_info_t
     ucc_cl_hier_allgatherv_algs[UCC_CL_HIER_ALLGATHERV_ALG_LAST + 1];
 
-#define UCC_CL_HIER_ALLGATHERV_DEFAULT_ALG_SELECT_STR "allgatherv:0-2k:host:@gab"
+#define UCC_CL_HIER_ALLGATHERV_DEFAULT_ALG_SELECT_STR "allgatherv:0-7864320:host:@gab"
```
nsarka, line 20: This might not be a sane default, especially for smaller teams.

janjust (approved):
generally ok, but maybe change that 7m number to something more generic like 8mb

src/components/cl/hier/allgatherv/allgatherv.h
janjust: echoing my comment here, instead of 7864320, just maybe put 8mb to be more intuitive

janjust:
bot:retest

samnordmann (approved):

src/components/cl/hier/allgatherv/allgatherv.c
```diff
@@ -69,53 +69,48 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
...
-    *ordered = is_block_ordered;
-
-    return UCC_OK;
-}
-
-/* Node leader subgroup is always ordered by ascending host_id. If the team's ranks are
-   not in the same order, then node leader subgroup allgatherv won't be enough, we'll
-   have to use a staging buffer and unpack to reorder each leader's contribution.
-   So, this func will check that the team ranks in the ldr sbgp are ascending */
-static inline ucc_status_t is_host_ordered(ucc_cl_hier_team_t *cl_team, int *ordered)
```
nsarka, line 93: Here I remove is_host_ordered. It checks that the node leader subgroup is in order with the team's ranks (the comment actually has it backwards). This function isn't needed anymore though because I sort the leader ranks to be in order of the team during sbgp creation.

src/components/cl/hier/allgatherv/allgatherv.c
```diff
@@ -69,53 +69,48 @@ static inline ucc_status_t find_leader_rank(ucc_base_team_t *team,
...
-    } else {
-        if (SBGP_EXISTS(cl_team, NODE_LEADERS)) {
-            is_host_ordered = 1;
-            max_rank = ucc_ep_map_eval(SBGP_MAP(cl_team, NODE_LEADERS), 0);
-            for (i = 1; i < SBGP_SIZE(cl_team, NODE_LEADERS); i++) {
-                team_rank = ucc_ep_map_eval(SBGP_MAP(cl_team, NODE_LEADERS), i);
-                if (team_rank < max_rank) {
-                    is_host_ordered = 0;
-                    break;
+        // For each node, check if its ranks are consecutive in team order
```
nsarka, line 88: Here I revert the implementation of is_block_ordered back to the original way I did it--for each node subgroup, check that the team's ranks are ascending and consecutive. The reason the old implementation doesn't work is that the full ordered subgroup also orders by socket and numa, and for the purpose of allgatherv the socket/numa placement doesn't matter.

PR #1130, opened 2025-05-06
CORE: Add UCC_DEBUGGER_WAIT env

nsarka:
Allows user to pass UCC_DEBUGGER_WAIT=<ctx rank> to make that rank wait for the debugger. Only works if UCC was configured with `--enable-debug`


nsarka:
@Sergei-Lebedev Please let me know about this one

PR #1133, opened 2025-05-11
Add NVTX events in triggered task start/finish

x41lakazam:


ikryukov (commented):
I believe it should be hidden behind an ifdef statement and guarded by a define or profile build. This approach will prevent performance regressions and simplify the build process.

janjust:
stale PR

PR #1140, opened 2025-06-13
TL/MLX5: recover from ipoib issue in mcast init

MamziB:
TL/MLX5: Recover from ipoib failure in mcast init and it avoids a segfault. It removes the unnecessary dereg function as registration has not happened before we fail here.


janjust:
@MamziB can you explain this PR in the description, why are we removing dereg_mr when init fails?

janjust (approved):

Sergei-Lebedev (approved):

PR #1142, opened 2025-06-18
TL/MLX5: progress service coll in mcast init

MamziB:
TL/MLX5: progress service collectives during mcast init. This is to fix a hang in ucc_perftest during subcomm creatation.


janjust:
Closing in favor of real fix in ucc_perf test. Need to call progress in ucc perf test

PR #1144, opened 2025-06-25
TL/CUDA: NVLink SHARP (NVLS) Reduce Scatter

ikryukov:
## What
Initial prototype of NVLS-based reduce-scatter implementation for UCC, leveraging NVIDIA's NVLink SHARP to accelerate collective operations with minimal SM usage.

## Why ?
Supports the latest NVLS technologies to enable UCC to take full advantage of cutting-edge hardware (H100+). NVLS provides significant performance benefits over existing GPU-based implementations by offloading collective operations to hardware, reducing SM usage and enabling more efficient resource utilization for compute workloads.

bandwidth_comparison

## Requirements

1. **CUDA 12.0 or later**: NVLS support requires CUDA 12.0+
2. **NVLS-capable hardware**: Requires hardware that supports NVLS (typically H100 and newer GPUs)

### Default `configure` Behavior
- If `--with-nvls` is not specified, NVLS will be disabled by default
- To enable NVLS, you must explicitly use `--with-nvls`
- If `--with-nvls` is specified but CUDA version is less than 12.0, configure will fail with an error
 
## How ?
- Implements CUDA Multicast memory allocation and management infrastructure
- Provides CUDA kernel-based reduce-scatter using NVLS multicast operations
- Currently supports SUM operation with float32 datatype on single node NVLINK systems (up to 8 GPUs)
- Uses staged execution with CPU barriers for synchronization between ranks
- Integrates with UCC's existing TL/CUDA framework and build system


config/m4/nvls.m4
```diff
@@ -0,0 +1,63 @@
+#
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# See file LICENSE for terms.
+#
+
+AC_DEFUN([CHECK_NVLS],[
```
Sergei-Lebedev, line 6: why is it separate file? imho it should be part of cuda.m4

src/components/tl/cuda/kernels/reduce_scatter_kernel.cu
```diff
@@ -0,0 +1,79 @@
+/**
+ * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_TL_CUDA_NVLS
```
Sergei-Lebedev: is it really needed? we include/exclude this file at make level

Sergei-Lebedev (approved):

config/m4/nvls.m4
ikryukov: decided to use a separate configuration because it is less invasive (we don't want to break anything)

janjust (approved):

B-a-S:
bot:retest

B-a-S:
test bot:retest plz

PR #1145, opened 2025-06-25
TL/UCP: fix allreduce knomial data consistency

Sergei-Lebedev:
## What
Allreduce knomial relies on  associativity property of reduce operations, however it does different reduction order on different ranks which is not allowed according to MPI standard. This PR fixes this issues

## How ?
Sort reduce buffers by their knomial group index at every iteration before doing reduce operation.

### DIFF
--- a/src/components/tl/ucp/allreduce/allreduce_knomial.c
+++ b/src/components/tl/ucp/allreduce/allreduce_knomial.c
@@ -41,6 +41,7 @@ void ucc_tl_ucp_allreduce_knomial_progress(ucc_coll_task_t *coll_task)
     ucc_status_t           status;
     ucc_kn_radix_t         loop_step;
     int                    is_avg;
+    ucc_kn_radix_t         index;
 
     if (UCC_IS_INPLACE(*args)) {
         sbuf = rbuf;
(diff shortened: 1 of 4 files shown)


src/components/tl/ucp/allreduce/allreduce_knomial.c
```diff
@@ -200,7 +204,10 @@ ucc_status_t ucc_tl_ucp_allreduce_knomial_start(ucc_coll_task_t *coll_task)
                (TASK_ARGS(task).src.info.mem_type == mem_type));
     cfg_radix = ucc_tl_ucp_get_radix_from_range(team, data_size, mem_type, p,
                                                 UCC_UUNITS_AUTO_RADIX);
-    ucc_knomial_pattern_init(size, rank, ucc_min(cfg_radix, size),
+    radix = ucc_min(cfg_radix, size);
+    /* max radix is limited by the number of buffers in the executor */
+    radix = ucc_min(radix, UCC_EE_EXECUTOR_NUM_BUFS);
```
Sergei-Lebedev, line 209: UCC_EE_EXECUTOR_NUM_BUFS-1

nsarka (approved):
LGTM, just don't forget to adjust allreduce_knomial.c:209

PR #1146, opened 2025-06-30
EC/ROCM: Include stdbool.h for new versions of ROCM

akolliasAMD:
## What
This PR includes stdbool.h in a rocm file.

## Why ?
In ROCM 7.0 Hip is removing some library inclusions. This will allow usage of `bool`

### DIFF
--- a/src/components/ec/rocm/ec_rocm_executor_interruptible.c
+++ b/src/components/ec/rocm/ec_rocm_executor_interruptible.c
@@ -5,6 +5,7 @@
  * See file LICENSE for terms.
  */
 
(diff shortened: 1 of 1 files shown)


Sergei-Lebedev:
Ok to test

akolliasAMD:
Hello, Do i need to do anything more? (Rebase etc?)

janjust:
@akolliasAMD no, we have 2 approvals and it's ready to merge but we're having issues with our CI which is why the tests are taking longer. We're working with our devops teams to bring them up and running so that we can get everything passing and this will be merged as is.

PR #1161, opened 2025-07-22
TL/MLX5: do reliability-init after mcast setup

MamziB:
mcast: postpone reliability-init until all ranks are ready

Fixes/Improvements:
- Skip the reliability‐init allgather during team-create when some ranks have not yet opened the mcast transport.
- Correct PD deallocation in ucc_finalize.
- Minor clean-ups and style fixes.

@janjust Shanny has already confirmed that this PR fixes all her tests.


src/components/tl/mlx5/mcast/tl_mlx5_mcast_team.c
```diff
@@ -327,10 +323,23 @@ ucc_status_t ucc_tl_mlx5_mcast_coll_setup_comm_resources(ucc_tl_mlx5_mcast_coll_
...
+        }
+    }
+
+    return UCC_OK;
 
 error:
-    ucc_tl_mlx5_clean_mcast_comm(comm);
+    /* Skip cleanup here: resources may be only partially initialized and
+     * ucc_tl_mlx5_clean_mcast_comm expects fully initialized structures.
+     * The higher-level fallback logic will dispose of @comm later if needed. */
```
nsarka, line 347: Where is the higher level logic that will free it? For example, where does `comm->pp_buf_header` get freed if the min allreduce is 0?

nsarka (approved):
LGTM, I just had a question about where the partial comm allocations are freed

src/components/tl/mlx5/mcast/tl_mlx5_mcast_team.c
MamziB: When the MIN-allreduce shows that at least one rank failed (```transport_ready_global == 0``` in the ```MCAST_RELIAB_SYNC``` state), we immediately call ```ucc_tl_mlx5_clean_mcast_comm(comm)```. That routine frees every resource that may have been partially initialised—including ```pp_buf_header```  and finally releases comm itself, so nothing is leaked even when the transport setup fails early.

janjust (approved):

artemry-nv:
Please rebase to latest master to make CI passed.

PR #1166, opened 2025-07-28
EC/CUDA: Set default threads to 256

nsarka:
Workaround for https://redmine.mellanox.com/issues/4545852

In UCC we use a kernel do handle copies and reductions for cuda buffers. The kernel is failing in the repro because of “requested resources too large”.
The workaround is to set UCC_EC_CUDA_EXEC_NUM_THREADS to 256, down from the default of 512.

The kernel that's failing to launch requires Registers Per Thread: 156 according to my nsys profile. By default we launch with 512 threads, so 512 * 156 = 79872

On the B300 node im running on, I printed the device caps:

Device #0: NVIDIA Graphics Device
  Compute Capability: 10.3
  Max threads per block: 1024
  Shared memory per block: 49152 bytes
  Registers per block: 65536
  Warp size: 32
  Max threads per multiprocessor: 2048
  Number of SMs: 152

79872 is greater than 65536, so that explains why it fails at 512 threads and passes at 256 threads, which only takes 156 * 256 = 39936

However, I thought that nvcc should spill any extra registers to LMEM. @Sergei-Lebedev do you have any ideas why it's failing to launch instead of spilling?


bureddy (approved):

janjust (approved):

Sergei-Lebedev:
@nsarka  I'm not sure compiler can do register spilling in runtime, but can we use cudaOccupancyMaxActiveBlocksPerMultiprocessor instead of hardcoded value?

nsarka:
> @nsarka I'm not sure compiler can do register spilling in runtime, but can we use cudaOccupancyMaxActiveBlocksPerMultiprocessor instead of hardcoded value?

The issue is the register count per SM, not the number of blocks on the SM. Or am I misunderstanding this API somehow?

Sergei-Lebedev:
> The issue is the register count per SM, not the number of blocks on the SM. Or am I misunderstanding this API somehow?

The idea is if this function return 0 blocks for the given thread count we decrease number of threads. Not sure if it will work, but something like this
```
for (num_threads = 1024; num_threads > 0; num_threads /= 2)
{
   int nblocks=0;
   cudaOccupancyMaxActiveBlocksPerMultiprocessor(&nblocks, cuda_executor, num_threads, 0);
   if (nblocks > 0) {
     break
   }
}
if (num_threads == 0) {
  // GPU devices is not supported
  return UCC_ERR_NOT_SUPPORTED;
}
```

nsarka:
> > The issue is the register count per SM, not the number of blocks on the SM. Or am I misunderstanding this API somehow?
> 
> The idea is if this function return 0 blocks for the given thread count we decrease number of threads. Not sure if it will work, but something like this
> 
> ```
> for (num_threads = 1024; num_threads > 0; num_threads /= 2)
> {
>    int nblocks=0;
>    cudaOccupancyMaxActiveBlocksPerMultiprocessor(&nblocks, cuda_executor, num_threads, 0);
>    if (nblocks > 0) {
>      break
>    }
> }
> if (num_threads == 0) {
>   // GPU devices is not supported
>   return UCC_ERR_NOT_SUPPORTED;
> }
> ```

How does this take into account the required registers per thread of the executor kernel? The issue is (almost) every SM version has 64k registers, and nvcc compiles `executor_kernel` to have a register per thread count of 156. Setting it to 256 makes it `156 * 256 = 39,936`, less than 64k.

nsarka:
@Sergei-Lebedev I updated the code. Please take a look when you get the chance

src/components/ec/cuda/kernel/ec_cuda_executor.cu
```diff
@@ -269,6 +269,52 @@ __global__ void executor_kernel(volatile ucc_ec_cuda_executor_t *eee,
...
+    ucc_assert(*max % WARP_SIZE == 0);
+
+    low = 0;
+    high = *max / WARP_SIZE;
+
+    while (low < high - 1) {
+        int mid = low + (high - low) / 2;
+        useCoopLaunch ?
+            GET_BLOCKS(mid * WARP_SIZE, true) :
+            GET_BLOCKS(mid * WARP_SIZE, false);
```
Sergei-Lebedev, line 295: why not simply GET_BLOCKS(mid * WARP_SIZE, useCoopLaunch); ?

src/components/ec/cuda/ec_cuda.c
```diff
@@ -88,12 +90,16 @@ static inline void ucc_ec_cuda_set_threads_nbr(int *nt, int maxThreadsPerBlock)
...
         }
-    }
+    } else {
+        *nt = (maxThreadsPerBlock / WARP_SIZE) * WARP_SIZE;
 
-    *nt = (maxThreadsPerBlock / WARP_SIZE) * WARP_SIZE;
+        if (!is_reduce) {
+            // Pass max threads per block, lowering it if necessary
+            // based on kernel occupancy requirements
+            ucc_ec_cuda_executor_kernel_calc_max_threads(nt);
```
Sergei-Lebedev: check return status

src/components/ec/cuda/kernel/ec_cuda_executor.cu
```diff
@@ -269,6 +269,52 @@ __global__ void executor_kernel(volatile ucc_ec_cuda_executor_t *eee,
 
 extern "C" {
 
+#define GET_BLOCKS(num_threads, coop) \
```
Sergei-Lebedev, line 272: why this macro needed?

src/components/ec/cuda/kernel/ec_cuda_executor.cu
nsarka: The template only takes a literal value, passing a variable doesn't work

src/components/ec/cuda/kernel/ec_cuda_executor.cu
nsarka: To reduce duplicate code, everything is the same where it gets used except for coop being passed true or false

Sergei-Lebedev (approved):

PR #1168, opened 2025-07-31
EC/CUDA: Link with stdc++

nsarka:
UCC's ec_cuda component (libucc_ec_cuda.so) is c++, so the compiler should link with libstdc++.so to add the c++ runtime symbols.

With cuda 13 (specifically on x86, ARM does not have this issue--the reason is the ARM libcudart.so depends on libstdc++.so, so it gets loaded, masking the issue), there is an unresolved symbol __cxa_guard_acquire that is left unresolved because libstdc++.so is not marked as a dependency.

Interestingly, with cuda 12, that symbol is not required by libucc_ec_cuda.so. I have not figured out why cuda 12 doesn't add it.

Regardless, I think not having -lstdc++ may have been an issue in UCC for a long time that was only exposed with the new cuda 13.

:x: Cuda 13 before the fix:
```
[nsarkauskas@rock17 ucc_private]$ nm -D /hpc/mtr_scrap/users/nsarkauskas/ucc_private/install/lib/ucc/libucc_ec_cuda.so | grep __cxa_guard_acquire
                 U __cxa_guard_acquire
[nsarkauskas@rock17 ucc_private]$ ldd /hpc/mtr_scrap/users/nsarkauskas/ucc_private/install/lib/ucc/libucc_ec_cuda.so
    linux-vdso.so.1 (0x00007ffd0912b000)
    libcuda.so.1 => /lib64/libcuda.so.1 (0x00007fe4b8544000)
    libcudart.so.13 => /hpc/local/oss/cuda13.0.0/redhat8/lib64/libcudart.so.13 (0x00007fe4b8296000)
    libucc.so.1 => /hpc/mtr_scrap/users/nsarkauskas/ucc_private/install/lib/libucc.so.1 (0x00007fe4b805b000)
    libucs.so.0 => /hpc/mtr_scrap/users/nsarkauskas/ucx/install/lib/libucs.so.0 (0x00007fe4b7c9c000)
    libm.so.6 => /lib64/libm.so.6 (0x00007fe4b791a000)
    libucm.so.0 => /hpc/mtr_scrap/users/nsarkauskas/ucx/install/lib/libucm.so.0 (0x00007fe4b76fb000)
    libz.so.1 => /lib64/libz.so.1 (0x00007fe4b74e3000)
    libdl.so.2 => /lib64/libdl.so.2 (0x00007fe4b72df000)
    libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fe4b70bf000)
    librt.so.1 => /lib64/librt.so.1 (0x00007fe4b6eb7000)
    libc.so.6 => /lib64/libc.so.6 (0x00007fe4b6af2000)
    /lib64/ld-linux-x86-64.so.2 (0x00007fe4beaeb000)
[nsarkauskas@rock17 ucc_private]$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Wed_Jul__2_10:42:27_AM_PDT_2025
Cuda compilation tools, release 13.0, V13.0.39
Build cuda_13.0.r13.0/compiler.36195134_0
```

:white_check_mark: Cuda 13 after the fix:

```
[nsarkauskas@rock17 ucc_private]$ nm -D /hpc/mtr_scrap/users/nsarkauskas/ucc_private/install/lib/ucc/libucc_ec_cuda.so | grep __cxa_guard_acquire
                 U __cxa_guard_acquire
[nsarkauskas@rock17 ucc_private]$ ldd /hpc/mtr_scrap/users/nsarkauskas/ucc_private/install/lib/ucc/libucc_ec_cuda.so | grep c++
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fb177515000)
```


Sergei-Lebedev:
perhaps nvcc in CUDA 12 compiles EC CUDA differently avoiding usage of this symbol. However I believe we don't need to include dependency on std lib in every component that uses CUDA like it's done now in the PR. I think the issue comes from the fact that we use C++ compiler to build kernels, but then we use C compiler to link kernels with ucc_ec_cuda.so. If this is true then we just need to fix link phase in EC CUDA and don't change MC CUDA, TL CUDA and TL NCCL which also depend on CUDA_LIBS

janjust (approved):

PR #1180, opened 2025-08-21
TL/CUDA: NVLS barrier

ikryukov:
## What
Introduce a cross-GPU barrier implementation using NVLS symmetric memory, eliminating the need for host-side synchronization.

## Why ?
The existing barrier relies on a host-managed shared memory synchronization mechanism. While functional, this approach is:
- **Expensive in performance** - frequent host-device roundtrips add latency.
- **Not scalable across nodes** - shared memory is not available in multinode deployments, making the current design unsuitable for larger topologies.
This PR addresses both issues by moving barrier management fully into device-side execution.

## How ?
- **Control space**: Reserve 1024 bytes in the NVLS symmetric buffer for barrier control segments.
- **Mechanism**:
  - The barrier is represented by a single monotonically incremented **uint64_t** counter.
  - Each participating GPU in the NVLS group increments the counter upon reaching the barrier.
  - Progress is determined by comparing the counter against the expected value, which scales with the NVLS group size.
- **Kernel integration**:
  - NVLS kernels launch with 1–32 blocks.
  - Each block designates a leader thread (`threadIdx.x == 0`).
  - The leader thread performs the atomic increment and spins until the counter matches the expected group-wide value.

This design ensures all GPUs in the NVLS group complete a given phase of the algorithm before any proceed, without requiring host intervention.

## Performance results:
bandwidth_comparison
latency_comparison

### DIFF
--- a/src/components/tl/cuda/allreduce/allreduce_nvls.c
+++ b/src/components/tl/cuda/allreduce/allreduce_nvls.c
@@ -15,16 +15,8 @@
 
 
 enum {
-    STAGE_COPY,                /*< Copy src buffer to symmetric memory */
-    STAGE_COPY_WAIT,           /*< Wait for the copy to complete */
-    STAGE_COPY_BAR_START,      /*< Start barrier after copy */
-    STAGE_COPY_BAR_TEST,       /*< Test barrier after copy */
-    STAGE_KERNEL_START,        /*< Start kernel */
-    STAGE_KERNEL,              /*< Kernel is running */
-    STAGE_BARRIER_START,       /*< Start barrier after kernel */
-    STAGE_BARRIER_TEST,        /*< Test barrier after kernel */
-    STAGE_COPY_POST,           /*< Copy result buffer from symmetric memory to dst buffer */
-    STAGE_COPY_POST_WAIT,      /*< Wait for the copy to complete */
+    STAGE_KERNEL,    /*< Post memcpy to symmetric buffer, launch kernel, memcpy to destination */
+    STAGE_WAIT,      /*< Wait for the copies and kernel to complete */
 };
 
 ucc_status_t ucc_tl_cuda_allreduce_nvls_start(ucc_coll_task_t *coll_task)
@@ -48,7 +40,7 @@ ucc_status_t ucc_tl_cuda_allreduce_nvls_start(ucc_coll_task_t *coll_task)
              task, stream, (void *)task->allreduce_nvls.uc_va, (void *)task->allreduce_nvls.mc_va,
              task->allreduce_nvls.buf_size_bytes, UCC_IS_INPLACE(*args));
 
-    task->allreduce_nvls.stage = STAGE_COPY;
+    task->allreduce_nvls.stage = STAGE_KERNEL;
 
     return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
 }
@@ -72,7 +64,7 @@ void ucc_tl_cuda_allreduce_nvls_progress(ucc_coll_task_t *coll_task)
     cudaError_t         cuda_status;
 
     switch (task->allreduce_nvls.stage) {
-    case STAGE_COPY:
+    case STAGE_KERNEL:
         // copy src buffer to symmetric memory first
         cuda_status =
             cudaMemcpyAsync((void *)uc_va, task->allreduce_nvls.sbuf,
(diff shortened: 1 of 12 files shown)


src/components/tl/cuda/tl_cuda_nvls.c
```diff
@@ -385,6 +386,29 @@ ucc_status_t ucc_tl_cuda_nvls_init(struct ucc_tl_cuda_team *self,
...
     nvls->mc_size      = mcSize;
     nvls->mc_offset    = mcOffset;
+    nvls->coll_ids     = ucc_malloc(lib->cfg.max_concurrent * sizeof(size_t), "coll_ids");
+    if (!nvls->coll_ids) {
+        status = UCC_ERR_NO_MEMORY;
+        goto cleanup;
+    }
+    for (i = 0; i < lib->cfg.max_concurrent; ++i) {
+        nvls->coll_ids[i] = 0;
+    }
```
Sergei-Lebedev: just memset

src/components/tl/cuda/tl_cuda_nvls.c
```diff
@@ -385,6 +386,29 @@ ucc_status_t ucc_tl_cuda_nvls_init(struct ucc_tl_cuda_team *self,
...
+
+    if (UCC_TL_TEAM_RANK(self) == 0) {
+        // root rank initializes the arrival counter for each task
+        ucc_tl_cuda_nvls_control_t control;
+        control.arrival_counter = 0;
+
+        for (i = 0; i < lib->cfg.max_concurrent; ++i) {
+            void *control_uc =
```
Sergei-Lebedev: can we use TASK_NVLS_CONTROL_UC here?

src/components/tl/cuda/tl_cuda_nvls.c
```diff
@@ -385,6 +386,29 @@ ucc_status_t ucc_tl_cuda_nvls_init(struct ucc_tl_cuda_team *self,
...
+                PTR_OFFSET((void *)uc_va, i * (lib->cfg.nvls_symmetric_size +
+                                               NVLS_CONTROL_SIZE) +
+                                              lib->cfg.nvls_symmetric_size);
+            CUDA_CHECK(cudaMemcpy(control_uc, &control, sizeof(ucc_tl_cuda_nvls_control_t),
```
Sergei-Lebedev: can we use cudaMemset instead?

src/components/tl/cuda/tl_cuda_nvls.c
```diff
@@ -439,5 +463,10 @@ ucc_status_t ucc_tl_cuda_nvls_destroy(struct ucc_tl_cuda_team *self,
         // release the multicast handle
         CUDADRV_FUNC(cuMemRelease(self->nvls.mc_handle));
     }
+
+    if (self->nvls.coll_ids) {
```
Sergei-Lebedev, line 459: is it assgined to NULL at init?

src/components/tl/cuda/allreduce/allreduce_nvls.c
```diff
@@ -84,98 +76,20 @@ void ucc_tl_cuda_allreduce_nvls_progress(ucc_coll_task_t *coll_task)
...
-        task->allreduce_nvls.stage = STAGE_KERNEL_START;
-        // fallthrough
-    case STAGE_KERNEL_START:
+
         status = post_allreduce_kernel(stream, sm_count, threads, mc_va,
                                        task->allreduce_nvls.buf_size_bytes,
-                                       trank, UCC_TL_TEAM_SIZE(team), dt);
+                                       (CUdeviceptr)TASK_NVLS_CONTROL_MC(task),
+                                       (CUdeviceptr)TASK_NVLS_CONTROL_UC(task),
+                                       sm_count * UCC_TL_TEAM_SIZE(team),
```
Sergei-Lebedev: i think this arg is not needed, we already pass sm_count and team_size as separate args

src/components/tl/cuda/allreduce/allreduce_nvls.c
```diff
@@ -84,98 +76,20 @@ void ucc_tl_cuda_allreduce_nvls_progress(ucc_coll_task_t *coll_task)
...
-        }
-        task->allreduce_nvls.stage = STAGE_KERNEL_START;
-        // fallthrough
-    case STAGE_KERNEL_START:
+
         status = post_allreduce_kernel(stream, sm_count, threads, mc_va,
                                        task->allreduce_nvls.buf_size_bytes,
-                                       trank, UCC_TL_TEAM_SIZE(team), dt);
+                                       (CUdeviceptr)TASK_NVLS_CONTROL_MC(task),
+                                       (CUdeviceptr)TASK_NVLS_CONTROL_UC(task),
```
Sergei-Lebedev: why cast to CUdeviceptr?

src/components/tl/cuda/tl_cuda_nvls.c
ikryukov: For this specific case we could simplify the initialization by using `cudaMemset`, since the control segment is just being zeroed. That said, in the more general case initialization may be more complex (e.g. setting structured fields, seeding counters, or writing precomputed metadata), which will require host -> device transfers.

src/components/tl/cuda/tl_cuda_nvls.c
ikryukov: No, since there is no tasks here. This is team initialization stage.

src/components/tl/cuda/tl_cuda_nvls.c
ikryukov: zeroed structure

src/components/tl/cuda/allreduce/allreduce_nvls.c
ikryukov: Fixed macroses to work with `CUdeviceptr`

src/components/tl/cuda/tl_cuda_nvls.c
Sergei-Lebedev: sure, that's why I asked to change to cudamemset here since we only to zero out struct

Sergei-Lebedev (approved):

PR #1186, opened 2025-09-07
TEST: bus bw

yaeliyac:
## What
Specifying perf test print to be bus BW instead of just BW 

## Why ?
each collective perf test has its own get_bw function which calculates the bus bw depending more then just on S/t.
As seen also in NCCL-tests explanation- 
https://github.com/NVIDIA/nccl-tests/blob/9a5c15461abcef145b907c54d04aea4e8d1cb21f/doc/PERFORMANCE.md?plain=1#L22
Bus BW depends also on number of ranks and other collective specifics.


janjust:
@yaeliyac probably need to fix the commit message something like: TEST: fix bus bw.

Sergei-Lebedev (approved):

Sergei-Lebedev:
ok to test

PR #1188, opened 2025-09-24
TL/UCP: enable knomial allgatherv

Sergei-Lebedev:
## What
Check if dst buffer is contig and use knomial allgatherv in this case

### DIFF
--- a/src/components/tl/ucp/allgatherv/allgatherv_knomial.c
+++ b/src/components/tl/ucp/allgatherv/allgatherv_knomial.c
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


src/components/tl/ucp/allgatherv/allgatherv_knomial.c
```diff
@@ -1,18 +1,23 @@
...
 
 #include "allgatherv/allgatherv.h"
 #include "allgather/allgather.h"
 
+
 ucc_status_t ucc_tl_ucp_allgatherv_knomial_init(ucc_base_coll_args_t *coll_args,
                                                 ucc_base_team_t *team,
                                                 ucc_coll_task_t **task_h)
 {
-    if (!UCC_COLL_IS_DST_CONTIG(&coll_args->args)) {
```
nsarka, line 14: Why not use this macro?

Sergei-Lebedev: it's already part of ucc_coll_args_is_disp_contig

janjust (approved):

PR #1191, opened 2025-10-03
CORE: port ucs logger to ucc

Sergei-Lebedev:
## What
Port UCS logger from libucs to ucc

## Why ?
- logger is no longer depend on libucs
- fix traces are not printing
- enable logger features such as file filter, log to file etc.


src/core/ucc_global_opts.h
```diff
@@ -11,7 +11,8 @@
 #include "utils/ucc_component.h"
 
 #include "utils/ucc_parser.h"
-#include "utils/ucc_log.h"
+//#include "utils/ucc_log.h"
```
wfaderhold21: dead code?

PR #1194, opened 2025-10-08
TOOLS: add support for onesided alltoall perftest

Sergei-Lebedev:
## What
Support for onesided alltoall in ucc perftest. Add buffer registration


wfaderhold21 (approved):
LGTM

PR #1200, opened 2025-10-16
TL/CUDA: enable NVLS build by default 

ikryukov:
## What
Enable NVLS build by default.
default: enable nvls on cuda platforms with sm90+
`--with-nvls` is set during config
 `--without-nvls` if you need to disable NVLS build

## Why ?
validate NVLS build path

### DIFF
--- a/config/m4/nvls.m4
+++ b/config/m4/nvls.m4
@@ -7,11 +7,18 @@ AC_DEFUN([CHECK_NVLS],[
 AS_IF([test "x$nvls_checked" != "xyes"],[
     nvls_happy="no"
 
+    # Use "check" as default to auto-detect NVLS support
+    # "yes" = user explicitly requested --with-nvls (error if unavailable)
+    # "no"  = user explicitly requested --without-nvls (skip entirely)
+    # "check" = auto-detect (enable if available, silently disable if not)
     AC_ARG_WITH([nvls],
-            [AS_HELP_STRING([--with-nvls], [Enable NVLS (NVLINK SHARP) support (default is no).])],
-            [], [with_nvls=no])
+            [AS_HELP_STRING([--with-nvls], [Enable NVLS (NVLINK SHARP) support (default is auto-detect).])],
+            [], [with_nvls=check])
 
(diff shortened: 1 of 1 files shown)


svcnbu-swx-hpcx:

wfaderhold21 (approved):

config/m4/nvls.m4
```diff
@@ -22,40 +29,58 @@ AS_IF([test "x$nvls_checked" != "xyes"],[
...
+                    AC_MSG_ERROR([NVLS support is requested but NVLS cannot be enabled. Requires CUDA 12.0 or later (detected: $CUDA_MAJOR_VERSION.$CUDA_MINOR_VERSION).])
+                ],
+                [
+                    AC_MSG_NOTICE([NVLS requires CUDA 12.0 or later, but CUDA $CUDA_MAJOR_VERSION.$CUDA_MINOR_VERSION was detected])
+                ])
             ])
         ],
         [
             nvls_happy="no"
+            AS_IF([test "x$with_nvls" = "xyes"],
```
Sergei-Lebedev, line 67: i don't think we need to check cuda version here. CUDA 11 doesn't support sm90 anyway

config/m4/nvls.m4
```diff
@@ -22,40 +29,58 @@ AS_IF([test "x$nvls_checked" != "xyes"],[
             # Check for CUDA 12.0+ which supports NVLS
             AS_IF([test $CUDA_MAJOR_VERSION -ge 12],
             [
-                nvls_happy="yes"
-                AC_DEFINE([HAVE_NVLS], [1], [Enable NVLS support])
+                # NVLS kernels use multimem PTX instructions which require sm_90+
+                # Check if NVCC_ARCH includes sm_90 or higher (90, 100, 110, 120)
```
Sergei-Lebedev, line 33: how does it work in runtime if we compile UCC for sm80 and sm90 and then run on sm80?

ikryukov: It relies on runtime checks: CUDA should report that multicast is not supported on tl/cuda team creation, and so on.

PR #1218, opened 2025-11-07
TL/CUDA: NVLS enable non root

ikryukov:
## What
This PR finalizes the implementation of setting ptrace permissions for NVLS POSIX handle sharing. 

## Why ?
In single-node NVLS configurations, rank 0 creates a multicast object and exports it as a POSIX file descriptor. Peer processes need to access this file descriptor using the `pidfd_getfd` system call, which requires ptrace permissions on the target process.
By default, the Linux kernel's Yama security module may block `pidfd_getfd` calls between processes. Setting `PR_SET_PTRACER_ANY` allows any process to use `pidfd_getfd` on the current process, enabling seamless file descriptor sharing without requiring system-wide ptrace policy changes.

### DIFF
--- a/src/components/tl/cuda/tl_cuda_nvls.c
+++ b/src/components/tl/cuda/tl_cuda_nvls.c
@@ -13,6 +13,7 @@
 #include "utils/ucc_math.h"
 
 #include <sys/syscall.h> // for pidfd_open and pidfd_getfd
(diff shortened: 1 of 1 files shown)


svcnbu-swx-hpcx:

Sergei-Lebedev (approved):

PR #1226, opened 2025-11-18
CI: add timeout to MPI tests

dpressle:
## What
Add timeout to CI tests

## Why ?
In some rare cases tests can get stuck and job will get timeout after 5 hours we want to have timeout for each step in the pipeline to avoid unneeded resource holding 

## How ?
add timeout on mpi tests and DLRM tests


dpressle:
> is that 20 minutes and 40 minutes respectively?

Yes, would you like me to add inline comments?

janjust:
> > is that 20 minutes and 40 minutes respectively?
> 
> Yes, would you like me to add inline comments?

yes please for future ref

dpressle:
bot:retest

PR #1228, opened 2025-11-25
Fix various typos

janjust:
Fixing typos throughout comments and some outputs.

### DIFF
--- a/config/m4/ax_prog_doxygen.m4
+++ b/config/m4/ax_prog_doxygen.m4
@@ -21,7 +21,7 @@
 #   The DX_*_FEATURE macros control the default setting for the given
 #   Doxygen feature. Supported features are 'DOXYGEN' itself, 'DOT' for
 #   generating graphics, 'HTML' for plain HTML, 'CHM' for compressed HTML
-#   help (for MS users), 'CHI' for generating a seperate .chi file by the
+#   help (for MS users), 'CHI' for generating a separate .chi file by the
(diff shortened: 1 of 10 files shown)


src/core/ucc_context.c
```diff
@@ -508,6 +514,12 @@ ucc_status_t ucc_core_addr_exchange(ucc_context_t *context, ucc_oob_coll_t *oob,
                 ucc_error("failed to start oob allgather");
                 return status;
             }
+            /* Validate that oob_req was properly initialized */
+            if (ucc_unlikely(addr_storage->oob_req == NULL)) {
```
Sergei-Lebedev: why req can't be null? i think it's implementation defined

src/core/ucc_context.c
```diff
@@ -542,6 +554,12 @@ ucc_status_t ucc_core_addr_exchange(ucc_context_t *context, ucc_oob_coll_t *oob,
             ucc_error("failed to start oob allgather");
             return status;
         }
+        /* Validate that oob_req was properly initialized */
+        if (ucc_unlikely(addr_storage->oob_req == NULL)) {
```
Sergei-Lebedev: why not null?

janjust: wops - not supposed to be there - I have a stray commit here.

janjust:
@Sergei-Lebedev I have a stray commit here - need to remove it. This Pr was only supposed to touch comments and output.

janjust:
@Sergei-Lebedev fixed - this was supposed to be just a small typos fix PR (I have an old "issue" from QA want to finally close it)

PR #1260, opened 2026-01-30
TL/NCCL: add user buffer registration via memmap

wfaderhold21:
## What
This PR implements NCCL User Buffer Registration (UBR) in TL/NCCL. It replaces the three stub functions (mem_map, mem_unmap, memh_pack) that previously returned UCC_ERR_NOT_SUPPORTED with an implementation that registers GPU memory buffers with NCCL communicators.

## Why
NCCL UBR (available since NCCL 2.19.0) lets applications pre-register GPU buffers so NCCL can bypass its internal memory registration on each collective call. Without this, NCCL must re-pin and register the buffer every time it's used in a collective, which adds latency. This is particularly impactful for AI/ML workloads where the same buffers (e.g., model gradients) are used in AllReduce repeatedly across training steps. This PR wires UCC's existing mem_map API into NCCL's ncclCommRegister/ncclCommDeregister, enabling that optimization.

## How
 - Lazy registration: mem_map only allocates a ucc_tl_nccl_memh_data_t tracking struct — no NCCL call
  yet. The actual ncclCommRegister call is deferred to the first collective operation that uses the
  buffer (ucc_tl_nccl_lazy_register_memh in tl_nccl_coll.c). This avoids registering with comms that
  never use the buffer.
  - Per-comm tracking: Each memh_data maintains dynamic arrays of (ncclComm_t, handle) pairs — one entry
   per NCCL communicator that has registered the buffer — growing as needed.
  - Thread safety: Spinlocks guard both the per-handle arrays and the context-level handle list. The
  NCCL calls (ncclCommRegister/ncclCommDeregister) are made outside spinlocks to avoid holding them
  during potentially slow CUDA operations.
  - Clean teardown: team_destroy calls ucc_tl_nccl_team_deregister_memhs to null out all slots for the
  dying comm before calling ncclCommDestroy, preventing use-after-free in a later mem_unmap. The reverse
   order (mem_unmap before team_destroy) is also handled cleanly.
  - Export/import support: memh_pack packs the buffer address and length for cross-process sharing.
  mem_map in import mode unpacks the local address from the blob.
  - Config toggle: A new UCC_TL_NCCL_ENABLE_UBR config option (try/yes/no) controls the feature, with
  version gating against NCCL 2.19.0 at compile time.

### DIFF
--- a/src/components/tl/nccl/tl_nccl.c
+++ b/src/components/tl/nccl/tl_nccl.c
@@ -45,8 +45,7 @@ static ucs_config_field_t ucc_tl_nccl_context_config_table[] = {
     {"SYNC", "auto",
      "Determines how UCC tests completion of NCCL collective",
      ucs_offsetof(ucc_tl_nccl_context_config_t, sync_type),
-     UCS_CONFIG_TYPE_ENUM(ucc_tl_nccl_completion_sync_names)
-    },
+     UCS_CONFIG_TYPE_ENUM(ucc_tl_nccl_completion_sync_names)},
 
     {"BLOCKING", "yes",
      "If set to no will use non-blocking mode communicator behavior, "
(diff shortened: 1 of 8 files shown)


wfaderhold21:
Tested allgather performance on GAIA with 8 nodes, 8 PPN:
allgather_ubr_bandwidth_clean

PR #1305, opened 2026-04-30
TOPO: resolve NVML fabric symbols lazily

janjust:
## What
This PR avoids a hard runtime dependency on newer NVML fabric-info symbols in the CUDA sysinfo component.

Specifically, `nvmlDeviceGetGpuFabricInfoV` and the legacy `nvmlDeviceGetGpuFabricInfo` path are now resolved lazily with `dlsym()` before use.

## Why ?
HPC SDK users can build UCC/HPC-X with newer CUDA/NVML headers but run on systems with older NVIDIA drivers. In that case, `libucc_sysinfo_cuda.so` could fail to load with:

`undefined symbol: nvmlDeviceGetGpuFabricInfoV`

That symbol is only available in newer `libnvidia-ml.so.1` versions, so directly referencing it raises the effective runtime driver requirement even when fabric metadata is optional.

## How ?
The CUDA sysinfo fabric query now uses a small runtime resolver for optional NVML fabric APIs. If the versioned fabric-info symbol is available, UCC uses it and preserves partition ID support. If it is unavailable, UCC falls back to the legacy fabric-info API when present. If neither symbol exists, fabric metadata remains unset and UCC continues without failing library load.

## Testing
Tested before/after with CUDA 12.9.1 and NVML 555 at build time, then switched runtime NVML to R525.60.13.

Before this PR:
- `config.h` defined `HAVE_NVML_GPU_FABRIC_INFO_V`.
- `readelf -Ws libucc_sysinfo_cuda.so | grep nvmlDeviceGetGpuFabricInfo` showed an undefined relocation for `nvmlDeviceGetGpuFabricInfoV`.
- With R525 NVML loaded, `ldd -r libucc_sysinfo_cuda.so` failed with:
  `undefined symbol: nvmlDeviceGetGpuFabricInfoV`

After this PR:
- `readelf` and `nm -D` show no dynamic dependency on `nvmlDeviceGetGpuFabricInfo*`.
- With R525 NVML loaded, `ldd -r libucc_sysinfo_cuda.so` reports no undefined fabric-info symbols.

### DIFF
--- a/src/components/topo/cuda/ucc_sysinfo_cuda.c
+++ b/src/components/topo/cuda/ucc_sysinfo_cuda.c
@@ -62,7 +62,13 @@ static void *ucc_sysinfo_cuda_nvml_get_handle(void)
     }
 #endif
 
-    return handle ? handle : RTLD_DEFAULT;
+    if (!handle) {
+        ucc_debug("could not obtain dedicated NVML handle; "
+                  "falling back to RTLD_DEFAULT");
+        return RTLD_DEFAULT;
+    }
+
+    return handle;
 }
 
 static void *ucc_sysinfo_cuda_nvml_get_symbol(const char *symbol)
--- a/src/components/topo/cuda/ucc_sysinfo_cuda.c
+++ b/src/components/topo/cuda/ucc_sysinfo_cuda.c
@@ -34,50 +34,18 @@ static pthread_mutex_t ucc_sysinfo_cuda_nvml_lock = PTHREAD_MUTEX_INITIALIZER;
  * headers while users run against an older libnvidia-ml.so.1, so resolve these
  * symbols lazily and leave fabric metadata unset when they are unavailable.
  */
-static void *ucc_sysinfo_cuda_nvml_get_handle(void)
-{
-    static int    resolved;
-    static void  *handle;
-    Dl_info       dl_info;
-
-    if (resolved) {
-        return handle ? handle : RTLD_DEFAULT;
-    }
-
-    resolved = 1;
-    if (dladdr((void *)nvmlInit_v2, &dl_info) && dl_info.dli_fname) {
-#ifdef RTLD_NOLOAD
-        handle = dlopen(dl_info.dli_fname,
-                        RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
-#endif
-        if (!handle) {
-            handle = dlopen(dl_info.dli_fname, RTLD_LAZY | RTLD_LOCAL);
-        }
-    }
-
-#ifdef RTLD_NOLOAD
-    if (!handle) {
-        handle = dlopen("libnvidia-ml.so.1",
-                        RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
-    }
-#endif
-
-    if (!handle) {
-        ucc_debug("could not obtain dedicated NVML handle; "
-                  "falling back to RTLD_DEFAULT");
-        return RTLD_DEFAULT;
-    }
-
-    return handle;
-}
-
 static void *ucc_sysinfo_cuda_nvml_get_symbol(const char *symbol)
 {
     void *sym;
     char *error;
 
+    /*
+     * libnvidia-ml.so.1 is already mapped (the component links it and calls
+     * nvmlInit_v2), so RTLD_DEFAULT resolves whatever fabric symbols the
+     * running driver exports without an explicit dlopen.
+     */
     dlerror();
-    sym = dlsym(ucc_sysinfo_cuda_nvml_get_handle(), symbol);
+    sym = dlsym(RTLD_DEFAULT, symbol);
     error = dlerror();
     if (error) {
         ucc_debug("NVML symbol %s is unavailable: %s", symbol, error);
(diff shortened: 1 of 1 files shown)


janjust:
@Sergei-Lebedev @ikryukov codex solution - I check it with the above tests, before/after PR, if you get a chance and have pointers how to guide it for a better fix if needed - let me know

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,155 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
+    }
+
+#ifdef RTLD_NOLOAD
+    if (!handle) {
+        handle = dlopen("libnvidia-ml.so.1",
+                        RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
+    }
+#endif
+
+    return handle ? handle : RTLD_DEFAULT;
```
Juee14Desai: If we fall through to RTLD_DEFAULT, it might be worth logging a ucc_debug() message so it's visible in debug output that we couldn't obtain a dedicated NVML handle.

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,161 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
+
+#ifdef HAVE_NVML_GPU_FABRIC_INFO_V
+typedef nvmlReturn_t (*ucc_nvmlDeviceGetGpuFabricInfoV_t)(
+    nvmlDevice_t device, nvmlGpuFabricInfoV_t *gpuFabricInfo);
+
+static ucc_nvmlDeviceGetGpuFabricInfoV_t
+ucc_sysinfo_cuda_nvml_get_gpu_fabric_info_v_fn(void)
+{
+    static int                                resolved;
+    static ucc_nvmlDeviceGetGpuFabricInfoV_t fn;
```
ikryukov: format?

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,161 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
+
+#ifdef HAVE_NVML_GPU_FABRIC_INFO
+typedef nvmlReturn_t (*ucc_nvmlDeviceGetGpuFabricInfo_t)(
+    nvmlDevice_t device, nvmlGpuFabricInfo_t *gpuFabricInfo);
+
+static ucc_nvmlDeviceGetGpuFabricInfo_t
+ucc_sysinfo_cuda_nvml_get_gpu_fabric_info_fn(void)
+{
+    static int                               resolved;
+    static ucc_nvmlDeviceGetGpuFabricInfo_t fn;
```
ikryukov: format? (use clang format)

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,161 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
+                                                ucc_gpu_info_t *gpu)
+{
+    nvmlReturn_t nvml_st;
+
+#ifdef HAVE_NVML_GPU_FABRIC_INFO_V
+    ucc_nvmlDeviceGetGpuFabricInfoV_t get_fabric_info_v;
+
+    get_fabric_info_v = ucc_sysinfo_cuda_nvml_get_gpu_fabric_info_v_fn();
+    if (get_fabric_info_v) {
+        nvmlGpuFabricInfoV_t fabric_info;
```
ikryukov: minor: declare all variables at the top of the function, but it is ok for me

ikryukov (approved):

nsarka (approved):

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,169 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
 
+#if defined(HAVE_NVML_H) && \
+    (defined(HAVE_NVML_GPU_FABRIC_INFO_V) || \
+     defined(HAVE_NVML_GPU_FABRIC_INFO))
+/*
+ * NVML fabric-info APIs are optional at runtime. Builds may use newer NVML
+ * headers while users run against an older libnvidia-ml.so.1, so resolve these
+ * symbols lazily and leave fabric metadata unset when they are unavailable.
+ */
+static void *ucc_sysinfo_cuda_nvml_get_handle(void)
```
Sergei-Lebedev: this function is not needed, nvmlInit_v2 is already loaded so we can just use RTLD_DEFAULT

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,169 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
+
+    if (!resolved) {
+        fn = (ucc_nvmlDeviceGetGpuFabricInfo_t)
+            ucc_sysinfo_cuda_nvml_get_symbol("nvmlDeviceGetGpuFabricInfo");
+        resolved = 1;
+    }
+
+    return fn;
+}
+#endif
```
Sergei-Lebedev: these two functions are identical, pls define macro to load symbol

src/components/topo/cuda/ucc_sysinfo_cuda.c
```diff
@@ -22,6 +26,169 @@ static ucc_config_field_t ucc_sysinfo_cuda_config_table[] = {
...
+#ifdef HAVE_NVML_FABRIC_PARTITION_ID
+                gpu->fabric_partition_id = fabric_info_v.partitionId;
+#endif
+                /* Globally-unique fabric id; needed for cross-node match. */
+                memcpy(gpu->fabric_cluster_uuid, fabric_info_v.clusterUuid,
+                       UCC_GPU_FABRIC_CLUSTER_UUID_LEN);
+            }
+            return;
+        }
+        ucc_debug("nvmlDeviceGetGpuFabricInfoV failed: %s",
```
Sergei-Lebedev: if fabric_info_v fails we probably don't need to try fabric_info after it, we need fabric_info only if fabric_info_v is missing

janjust:
@Sergei-Lebedev  done - and I verified that this fix actually works.

addressed all three in the latest commit:

RTLD_DEFAULT: dropped the dlopen/handle helper; libnvidia-ml is already loaded (we link it + call nvmlInit_v2), so we just dlsym(RTLD_DEFAULT, ...).
dedup: the two identical loaders are now generated by a single UCC_NVML_SYMBOL_LOADER macro.
fallback: if the versioned symbol resolves we treat it as authoritative and no longer fall through to the legacy API on a runtime error — legacy is only used when the versioned symbol is absent.
Also rebased on latest master and folded in the clusterUuid handling from #1308.

How I tested
Built against CUDA 13 (so the versioned symbol is compiled in), then two checks:

No hard dependency — confirms the symbol is resolved by name, not linked:


$ nm -D --undefined-only libucc_sysinfo_cuda.so | grep -i GpuFabricInfo
(nothing)      # pre-fix this showed:  U nvmlDeviceGetGpuFabricInfoV
Graceful fallback at runtime — ran a small osu_allreduce -d cuda (UCC) with an LD_PRELOAD shim that hides nvmlDeviceGetGpuFabricInfoV from dlsym, simulating an older driver:


NVML symbol nvmlDeviceGetGpuFabricInfoV is unavailable ...
GPU 0: caps=0x1 ... cluster_uuid=0000...
1024   197.23        # collective still completes, no load failure
Interestingly our R580 driver no longer exports the legacy symbol either, so this actually hit the "neither API available" path — loads and runs fine, fabric info just left unset.

Sergei-Lebedev (approved):

PR #1322, opened 2026-07-08
stop perftest after argument parse errors

jeffnvidia:
## What
Stop `ucc_perftest` immediately when `ucc_pt_config::process_args()` returns an error.

Add a regression check that invokes `ucc_perftest --gen exp:` and verifies a nonzero exit, the parser diagnostic, and no benchmark header.

## Why
Malformed arguments were diagnosed by the parser, but `main()` discarded the returned status and continued into CUDA/ROCm probing, communicator initialization, and potentially benchmark execution.

## Validation
- Build & Test passed: https://github.com/jeffnvidia/ucc/actions/runs/28795294451
- The `UCC perftest` job passed, including the new malformed-argument regression check.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -177,7 +177,7 @@ jobs:
         fi
         echo "${output}"
         case "${output}" in
-          *"Invalid format for --gen exp:min=N[[]@max=M]"*) ;;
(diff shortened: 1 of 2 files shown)


Sergei-Lebedev (approved):
don't think we need negative tests in CI

janjust (approved):

Issue #176, opened 2021-04-30
Component architecture  diagram


Sergei-Lebedev:
Readme file points to old component architecture diagram (docs/images/ucc_components.png) and there is no png file in docs directory for the new diagram.

manjugv:
Thanks @Sergei-Lebedev Will fix it. I need to update the component diagram, anyways.

PR #1326, opened 2026-07-12
TOOLS: add selectable iteration cadence

jeffnvidia:
## Summary

Add `--iteration-mode isolated|streaming` to `ucc_perftest`.

- `isolated` preserves the existing per-iteration barrier behavior and remains the default.
- `streaming` runs measured collectives back-to-back, with ranks aligned once before the measured batch.

The streaming mode matches the default measurement cadence in `nccl-tests`, enabling like-for-like comparisons without changing existing `ucc_perftest` results by default. This distinction can be significant: in one representative 32-rank, 2 GiB Alltoall run, isolated and streaming UCC measurements reached 94.33 and 54.16 GB/s bus bandwidth respectively; the `nccl-tests` default cadence measured 53.67 GB/s through the UCC NCCL net plugin.

## Validation

- Built with warnings treated as errors.
- Runtime-tested both modes on multi-node Alltoall.
- Verified the default remains equivalent to `--iteration-mode isolated`.

### DIFF
--- a/tools/perf/ucc_pt_benchmark.cc
+++ b/tools/perf/ucc_pt_benchmark.cc
@@ -231,6 +231,11 @@ ucc_status_t ucc_pt_benchmark::run_single_coll_test(ucc_coll_args_t args,
 
     args.root = config.root % comm->get_size();
     for (int i = 0; i < nwarmup + niter; i++) {
+        /* Start the measured streaming batch with ranks aligned. */
+        if (i == nwarmup && nwarmup > 0 &&
+            config.iteration_mode == UCC_PT_ITER_MODE_STREAMING) {
+            UCCCHECK_GOTO(comm->barrier(), exit_err, st);
+        }
         double s = get_time_us();
 
         if (!persistent) {
@@ -265,7 +270,9 @@ ucc_status_t ucc_pt_benchmark::run_single_coll_test(ucc_coll_args_t args,
             time += f - s;
         }
         args.root = (args.root + config.root_shift) % comm->get_size();
-        UCCCHECK_GOTO(comm->barrier(), exit_err, st);
+        if (config.iteration_mode == UCC_PT_ITER_MODE_ISOLATED) {
+            UCCCHECK_GOTO(comm->barrier(), exit_err, st);
+        }
     }
 
     if (persistent) {
@@ -376,6 +383,9 @@ void ucc_pt_benchmark::print_header()
                   << "  small" << config.n_iter_small << std::endl
                   << std::left << std::setw(24)
                   << "  large" << config.n_iter_large << std::endl;
+        std::cout << std::left << std::setw(24)
+                  << "Iteration mode:"
+                  << ucc_pt_iter_mode_str(config.iteration_mode) << std::endl;
         std::cout.copyfmt(iostate);
         std::cout << std::endl;
         std::cout << std::setw(12) << "Count"
--- a/tools/perf/ucc_pt_benchmark.cc
+++ b/tools/perf/ucc_pt_benchmark.cc
@@ -202,6 +202,8 @@ ucc_status_t ucc_pt_benchmark::run_single_coll_test(ucc_coll_args_t args,
 {
     const bool    triggered  = config.triggered;
     const bool    persistent = config.persistent;
+    const bool    streaming  =
+        config.iteration_mode == UCC_PT_ITER_MODE_STREAMING;
     ucc_team_h    team       = comm->get_team();
     ucc_context_h ctx        = comm->get_context();
     ucc_status_t  st         = UCC_OK;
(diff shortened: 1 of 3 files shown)


tools/perf/ucc_pt_benchmark.cc
```diff
@@ -265,7 +270,9 @@ ucc_status_t ucc_pt_benchmark::run_single_coll_test(ucc_coll_args_t args,
             time += f - s;
         }
         args.root = (args.root + config.root_shift) % comm->get_size();
-        UCCCHECK_GOTO(comm->barrier(), exit_err, st);
+        if (config.iteration_mode == UCC_PT_ITER_MODE_ISOLATED) {
```
Sergei-Lebedev: merge two barriers under one if

tools/perf/ucc_pt_benchmark.cc
```diff
@@ -231,6 +231,11 @@ ucc_status_t ucc_pt_benchmark::run_single_coll_test(ucc_coll_args_t args,
 
     args.root = config.root % comm->get_size();
     for (int i = 0; i < nwarmup + niter; i++) {
+        /* Start the measured streaming batch with ranks aligned. */
+        if (i == nwarmup && nwarmup > 0 &&
+            config.iteration_mode == UCC_PT_ITER_MODE_STREAMING) {
```
Sergei-Lebedev: pls add streaming bool variable similar to persistent and triggered

tools/perf/ucc_pt_benchmark.cc
```diff
@@ -265,7 +267,9 @@ ucc_status_t ucc_pt_benchmark::run_single_coll_test(ucc_coll_args_t args,
             time += f - s;
         }
         args.root = (args.root + config.root_shift) % comm->get_size();
-        UCCCHECK_GOTO(comm->barrier(), exit_err, st);
+        if (!streaming || (niter > 0 && i == nwarmup - 1)) {
```
Sergei-Lebedev, line 270: Why need to check niter > 0? Seems like unnecessary.

Sergei-Lebedev (approved):

tools/perf/ucc_pt_benchmark.cc
jeffnvidia: if niter=0, it would do an unnecessary barrier since there is no iterations. I thought it didnt make sense to do this barrier but it's not harmful to do it so up to you.

wfaderhold21 (approved):
may want to add a note somewhere that using the streaming mode and onesided alltoallv in tl/ucp may result in a hang

svcnbu-swx-hpcx:
🤖 **CI Triage Agent** — `UCC/ucc-run-coverity` · commit `f5fd5d13`

**TL;DR:** The Coverity build and analysis succeeded; the job failed only when `cov-commit-defects` couldn't reach the upload server `coverity.mellanox.com:8443` ("Connection refused"). This is an infrastructure/network outage, not a code defect in PR #1326.

**Summary:** `UCC/ucc-run-coverity` #176 failed in the "Run Coverity" stage at the defect-upload step; `cov-commit-defects` could not connect to the Coverity server.

**Root cause:** Network connectivity failure to the Coverity commit server. The log shows analysis completing normally (2263 defects found, exit 0), then `cov-commit-defects` failing with:
- `libcurl CONNECT to https://coverity.mellanox.com:8443/ccd ... failed with code Couldn't connect to server`
- `[ERROR] Attempting to connect to coverity.mellanox.com:8443: connect: 10.213.105.1:8443: Connection refused`

The `handle_error` trap then propagated exit code 1, aborting the stage. This is a server/network availability problem (server down, firewall, or VPN/routing issue to 10.213.105.1), completely independent of the commit under test. Note the analysis step took only ~26s and the whole failure happened in seconds — there is no hang; timing is not a factor.

**Implicated commit:** none — not caused by commit f5fd5d13 (PR #1326). The failure is in CI infrastructure.

**File:** `.ci/scripts/coverity.sh` (the `cov-commit-defects ... --host coverity.mellanox.com --port 8443` invocation and its `handle_error` trap)

**Suggested fix:**
- Primary: this is an infra issue — verify the Coverity server at `coverity.mellanox.com:8443` (10.213.105.1) is up and reachable from the Jenkins agent (firewall/VPN/routing). Once reachable, simply re-run the build.
- Hardening (optional): make the upload step resilient so a transient server outage doesn't fail the whole pipeline — e.g., retry `cov-commit-defects` with backoff, and/or treat an upload-only connection failure as `UNSTABLE` rather than `FAILURE`, since the build+analysis (the parts that validate the PR) already succeeded. Note there is an exposed password on the `cov-commit-defects` command line in the script/log context; ensure it stays passed via env/credentials and is masked (it appears masked as `****` in the log).

**Related:** none found.

</details>

PR #1072, opened 2025-02-03
TL/UCP: add radix selection to kn radix

Sergei-Lebedev:
## What
Select optimal knomial radix for non host memory

PR #1327, opened 2026-07-12
TL/UCP: add exact mixed-radix allgather schedules

jeffnvidia:
## What

Add exact per-phase radix schedules to TL/UCP K-nomial AllGather. A schedule
such as `8x6` uses radix 8 in the first phase and radix 6 in the second; its
product must equal the team size.

Set an explicit schedule with, for example:

```
UCC_TL_UCP_ALLGATHER_KN_RADIX=8x6
```

The same option accepts a fixed radix or `auto`. The existing fixed-radix path
and AllGatherV behavior remain unchanged. Invalid schedules are rejected, and
existing algorithm tuning still decides whether K-nomial AllGather is selected.

## Why

For non-power team sizes, a fixed radix can produce a truncated final phase or
proxy/extra ranks. An exact mixed schedule covers the complete team with an
explicit radix for every phase.

Across three tested environments at 96 ranks, a representative exact schedule
improved the large-message endpoint by **1.81x--1.99x** over the best measured
fixed radix by avoiding the fixed-radix remainder path.

## Performance evidence

Results are intentionally anonymized and normalized to the best measured
fixed-radix execution for each environment. Each row uses one exact schedule
at the endpoint; it is not a pointwise best-schedule envelope.

| Environment | Team | Exact schedule | Fixed baseline | Normalized speedup |
|---|---:|---|---|---:|
| System A | 96 ranks | `4x4x6` | R4 | 1.98x |
| System B | 96 ranks | `2x2x2x2x2x3` | R4 | 1.99x |
| System C | 96 ranks | `8x3x4` | R3 | 1.81x |

These relative results demonstrate the remainder-path mechanism; they are not
intended as absolute system-performance comparisons.

## Validation

- Capability debug build and focused schedule tests passed.
- 16-rank CUDA AllGather covered fixed R2/R4/R8 and explicit `8x2`, `2x8`,
  and `4x4`; all cases passed with zero failures.
- Tests cover parsing and bounded formatting, legacy fixed-radix behavior, and
  exact mixed peer sets, segment sizes, and offsets for every rank.

Automatic schedule selection is proposed separately in #1328.

### DIFF
--- a/src/coll_patterns/recursive_knomial.h
+++ b/src/coll_patterns/recursive_knomial.h
@@ -7,7 +7,11 @@
 #ifndef RECURSIVE_KNOMIAL_H_
 #define RECURSIVE_KNOMIAL_H_
 
+#include <limits.h>
+
 #define UCC_KN_PEER_NULL ((ucc_rank_t)-1)
+/* An exact schedule of radix >= 2 cannot have more phases than this. */
+#define UCC_KN_MAX_RADIX_PHASES (sizeof(ucc_rank_t) * CHAR_BIT - 1)
 typedef uint16_t ucc_kn_radix_t;
 
 enum {
@@ -29,7 +33,7 @@ enum {
 
 typedef struct ucc_knomial_pattern {
 
-    ucc_kn_radix_t radix;         /* knomial tree radix */
+    ucc_kn_radix_t radix;         /* radix for current iteration */
     uint8_t        type;          /* pattern type */
     uint8_t        iteration;     /* current iteration */
     uint8_t        n_iters;       /* number of iterations in knomial algorithm */
@@ -54,6 +58,8 @@ typedef struct ucc_knomial_pattern {
     ucc_rank_t     block_size;
     ptrdiff_t      block_offset;
     int            is64;
+    uint8_t        is_mixed;
+    ucc_kn_radix_t radices[UCC_KN_MAX_RADIX_PHASES];
 } ucc_knomial_pattern_t;
 
 /**
@@ -101,6 +107,7 @@ ucc_knomial_pattern_init_impl(ucc_rank_t size, ucc_rank_t rank,
     p->rank          = rank;
     p->backward      = backward;
     p->iteration     = 0;
+    p->is_mixed      = 0;
     n_full_subtrees  = ucc_kn_pattern_n_full(p);
     p->n_extra       = has_extra ? size - n_full_subtrees * p->full_pow_size : 0;
     p->n_iters       = (p->n_extra && n_full_subtrees == 1) ?
@@ -229,6 +236,9 @@ ucc_knomial_pattern_next_iteration(ucc_knomial_pattern_t *p)
 {
     p->iteration++;
     p->radix_pow *= p->radix;
+    if (p->is_mixed && !ucc_knomial_pattern_loop_done(p)) {
+        p->radix = p->radices[p->iteration];
+    }
 }
 
 static inline void ucc_knomial_pattern_prev_iteration(ucc_knomial_pattern_t *p)
--- a/src/coll_patterns/sra_knomial.h
+++ b/src/coll_patterns/sra_knomial.h
@@ -17,6 +17,9 @@
 static inline
 ucc_rank_t ucc_kn_compute_step_radix(ucc_knomial_pattern_t *p)
 {
+    if (p->is_mixed) {
+        return p->radix;
+    }
     int n_full = ucc_kn_pattern_n_full(p);
 
     return p->radix_pow * p->radix >= p->size ? (n_full > 1 ? n_full : p->radix)
@@ -270,6 +273,38 @@ ucc_kn_ag_pattern_init(ucc_rank_t size, ucc_rank_t rank, ucc_kn_radix_t radix,
     p->block_size   = p->radix_pow * radix;
     p->block_offset = ucc_knomial_pattern_loop_rank(p, rank) / p->block_size *
                       p->block_size;
+}
+
+static inline void ucc_kn_ag_pattern_reset(ucc_knomial_pattern_t *p)
+{
+    p->iteration    = 0;
+    p->radix        = p->radices[0];
+    p->radix_pow    = 1;
+    p->block_size   = p->radix;
+    p->block_offset = ucc_knomial_pattern_loop_rank(p, p->rank) /
+                      p->block_size * p->block_size;
+}
+
+static inline void ucc_kn_ag_pattern_init_mixed(
+    ucc_rank_t size, ucc_rank_t rank, size_t count,
+    const ucc_kn_radix_t *radices, uint8_t nradices, ucc_knomial_pattern_t *p)
+{
+    uint8_t i;
+
+    ucc_assert(nradices > 0 && nradices <= UCC_KN_MAX_RADIX_PHASES);
+    p->type      = KN_PATTERN_ALLGATHER;
+    p->n_iters   = nradices;
+    p->node_type = KN_NODE_BASE;
+    p->backward  = 0;
+    p->size      = size;
+    p->rank      = rank;
+    p->n_extra   = 0;
+    p->count     = count;
+    p->is_mixed  = 1;
+    for (i = 0; i < nradices; i++) {
+        p->radices[i] = radices[i];
+    }
+    ucc_kn_ag_pattern_reset(p);
 }
 
 static inline void
--- a/src/components/tl/ucp/allgather/allgather_knomial.c
+++ b/src/components/tl/ucp/allgather/allgather_knomial.c
@@ -10,6 +10,7 @@
 #include "tl_ucp_sendrecv.h"
 #include "tl_ucp_copy.h"
 #include "core/ucc_progress_queue.h"
+#include "allgather.h"
 #include "coll_patterns/sra_knomial.h"
 #include "tl_ucp_task.h"
 #include "ucc/api/ucc.h"
@@ -198,6 +199,7 @@ UCC_KN_PHASE_EXTRA:
             return;
         }
         ucc_kn_ag_pattern_next_iter(p);
+        radix = p->radix;
     }
 
     if (KN_NODE_PROXY == node_type) {
@@ -243,8 +245,11 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
     task->allgather_kn.copy_task = NULL;
     task->allgather_kn.phase     = UCC_KN_PHASE_INIT;
     if (ct == UCC_COLL_TYPE_ALLGATHER) {
-        ucc_kn_ag_pattern_init(size, rank, radix, args->dst.info.count,
-                               &task->allgather_kn.p);
+        if (p->is_mixed) {
+            ucc_kn_ag_pattern_reset(p);
+        } else {
+            ucc_kn_ag_pattern_init(size, rank, radix, args->dst.info.count, p);
+        }
         offset = ucc_buffer_block_offset(args->dst.info.count, size, rank) *
                  ucc_dt_size(args->dst.info.datatype);
         rbuf   = args->dst.info.buffer;
@@ -306,9 +311,10 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
     return ucc_progress_queue_enqueue(UCC_TL_CORE_CTX(team)->pq, &task->super);
 }
 
-ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
+static ucc_status_t ucc_tl_ucp_allgather_knomial_init_common(
     ucc_base_coll_args_t *coll_args, ucc_base_team_t *team,
-    ucc_coll_task_t **task_h, ucc_kn_radix_t radix)
+    ucc_coll_task_t **task_h, ucc_kn_radix_t radix,
+    const ucc_kn_radix_t *radices, uint8_t nradices)
 {
     ucc_tl_ucp_team_t    *tl_team = ucc_derived_of(team, ucc_tl_ucp_team_t);
     ucc_tl_ucp_context_t *ctx     = UCC_TL_UCP_TEAM_CTX(tl_team);
@@ -323,7 +329,20 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
         task->subset.myrank = sbgp->group_rank;
         task->subset.map    = sbgp->map;
     }
-    task->allgather_kn.p.radix = radix;
+    if (nradices > 0) {
+        ucc_rank_t tsize = UCC_TL_TEAM_SIZE(tl_team);
+
+        ucc_kn_ag_pattern_init_mixed(
+            tsize,
+            task->subset.myrank,
+            GET_TOTAL_COUNT(&coll_args->args, tsize),
+            radices,
+            nradices,
+            &task->allgather_kn.p);
+    } else {
+        task->allgather_kn.p.radix    = radix;
+        task->allgather_kn.p.is_mixed = 0;
+    }
     if (!UCC_IS_INPLACE(coll_args->args)) {
         if (ctx->cfg.local_copy_type == UCC_TL_UCP_LOCAL_COPY_TYPE_EC) {
             task->super.flags         |= UCC_COLL_TASK_FLAG_EXECUTOR;
@@ -357,6 +376,14 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
     return UCC_OK;
 }
 
+ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
+    ucc_base_coll_args_t *coll_args, ucc_base_team_t *team,
+    ucc_coll_task_t **task_h, ucc_kn_radix_t radix)
+{
+    return ucc_tl_ucp_allgather_knomial_init_common(
+        coll_args, team, task_h, radix, NULL, 0);
+}
+
 ucc_status_t ucc_tl_ucp_allgather_knomial_init(ucc_base_coll_args_t *coll_args,
                                                ucc_base_team_t      *team,
                                                ucc_coll_task_t     **task_h)
@@ -368,8 +395,30 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init(ucc_base_coll_args_t *coll_args,
     size_t             count   = GET_TOTAL_COUNT(&coll_args->args, tsize);
     ucc_datatype_t     dtype   = GET_DT(&coll_args->args);
     ucc_kn_radix_t     radix;
+    ucc_kn_radix_t     radices[UCC_KN_MAX_RADIX_PHASES];
+    uint8_t            nradices = 0;
+    ucc_status_t       status;
 
     radix = ucc_tl_ucp_get_knomial_radix(tl_team, count, dtype, mtype, p, 0);
+    if (coll_args->args.coll_type != UCC_COLL_TYPE_ALLGATHER) {
+        return ucc_tl_ucp_allgather_knomial_init_r(
+            coll_args, team, task_h, radix);
+    }
 
-    return ucc_tl_ucp_allgather_knomial_init_r(coll_args, team, task_h, radix);
+    if (tl_team->cfg.allgather_kn_mixed_radices != NULL &&
+        tl_team->cfg.allgather_kn_mixed_radices[0] != '\0') {
+        status = ucc_tl_ucp_allgather_knomial_parse_radices(
+            tl_team->cfg.allgather_kn_mixed_radices, tsize, radices, &nradices);
+        if (status != UCC_OK) {
+            tl_error(
+                UCC_TL_TEAM_LIB(tl_team),
+                "invalid ALLGATHER_KN_MIXED_RADICES '%s' for team size %u",
+                tl_team->cfg.allgather_kn_mixed_radices,
+                tsize);
+            return status;
+        }
+    }
+
+    return ucc_tl_ucp_allgather_knomial_init_common(
+        coll_args, team, task_h, radix, radices, nradices);
 }
--- a/src/components/tl/ucp/allgather/allgather_knomial_schedule.c
+++ b/src/components/tl/ucp/allgather/allgather_knomial_schedule.c
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * See file LICENSE for terms.
+ */
+
+#include "allgather.h"
+
+#include <stdlib.h>
+
+ucc_status_t ucc_tl_ucp_allgather_knomial_parse_radices(
+    const char *value, ucc_rank_t team_size, ucc_kn_radix_t *radices,
+    uint8_t *nradices)
+{
+    const char   *p = value;
+    char         *end;
+    unsigned long parsed;
+    ucc_rank_t    product = 1;
+    uint8_t       n       = 0;
+
+    *nradices             = 0;
+    if (value == NULL || value[0] == '\0') {
+        return UCC_ERR_NOT_FOUND;
+    }
+
+    while (*p != '\0') {
+        if (n == UCC_KN_MAX_RADIX_PHASES) {
+            return UCC_ERR_INVALID_PARAM;
+        }
+        parsed = strtoul(p, &end, 10);
+        if (end == p || parsed < 2 || parsed > UINT16_MAX ||
+            product > UCC_RANK_MAX / parsed) {
+            return UCC_ERR_INVALID_PARAM;
+        }
+        radices[n++] = (ucc_kn_radix_t)parsed;
+        product *= (ucc_rank_t)parsed;
+        if (*end == '\0') {
+            break;
+        }
+        if (*end != ',' || end[1] == '\0') {
+            return UCC_ERR_INVALID_PARAM;
+        }
+        p = end + 1;
+    }
+
+    if (product != team_size) {
+        return UCC_ERR_INVALID_PARAM;
+    }
+    *nradices = n;
+    return UCC_OK;
+}
--- a/src/components/tl/ucp/tl_ucp.c
+++ b/src/components/tl/ucp/tl_ucp.c
@@ -172,6 +172,13 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT_RANGED},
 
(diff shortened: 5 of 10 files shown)


jeffnvidia:
Maintainer note on the `tl_ucp.c` diff and codestyle:

I intentionally reduced this file to the functional seven-line
`ALLGATHER_KN_MIXED_RADICES` entry and removed the unrelated formatting churn.
With clang-format 21.1.8, touching that entry expands formatting to the entire
legacy configuration-array initializer, changing the file from 7 additions / 0
deletions to 105 additions / 49 deletions. I did not add formatter-suppression
directives.

The minimal diff may therefore make the codestyle check fail. Could maintainers
please accept/waive that result, or advise if a separate formatting-only change
is preferred?

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -243,8 +245,11 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
     task->allgather_kn.copy_task = NULL;
     task->allgather_kn.phase     = UCC_KN_PHASE_INIT;
     if (ct == UCC_COLL_TYPE_ALLGATHER) {
-        ucc_kn_ag_pattern_init(size, rank, radix, args->dst.info.count,
-                               &task->allgather_kn.p);
+        if (p->is_mixed) {
+            ucc_kn_ag_pattern_reset(p);
+        } else {
+            ucc_kn_ag_pattern_init(size, rank, radix, args->dst.info.count, p);
+        }
```
Sergei-Lebedev: need to unify these two paths and rewrite without mixed

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -368,8 +395,30 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init(ucc_base_coll_args_t *coll_args,
...
+    uint8_t            nradices = 0;
+    ucc_status_t       status;
 
     radix = ucc_tl_ucp_get_knomial_radix(tl_team, count, dtype, mtype, p, 0);
+    if (coll_args->args.coll_type != UCC_COLL_TYPE_ALLGATHER) {
+        return ucc_tl_ucp_allgather_knomial_init_r(
+            coll_args, team, task_h, radix);
+    }
+
+    if (tl_team->cfg.allgather_kn_mixed_radices != NULL &&
```
Sergei-Lebedev: don't parse radices every time in collective init. Instead do it once during team init

src/components/tl/ucp/tl_ucp.c
```diff
@@ -172,6 +172,13 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT_RANGED},
 
+    {"ALLGATHER_KN_MIXED_RADICES", "",
+     "Optional exact mixed-radix schedule for knomial allgather, for example "
+     "4,4,6. Every radix must be at least 2 and their product must equal the "
+     "team size. Invalid schedules are rejected",
+     ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_mixed_radices),
+     UCC_CONFIG_TYPE_STRING},
+
```
Sergei-Lebedev: let's keep only one config value ALLGATHER_KN_RADIX but also allow to set multiple radixes using for instance "x". E.g 2x3x4:cuda

src/components/tl/ucp/allgather/allgather_knomial_schedule.c
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * See file LICENSE for terms.
+ */
+
+#include "allgather.h"
+
+#include <stdlib.h>
+
+ucc_status_t ucc_tl_ucp_allgather_knomial_parse_radices(
```
Sergei-Lebedev: move this function to parser, probably need to implement scanf, printf etc for the new field type

src/coll_patterns/recursive_knomial.h
```diff
@@ -54,6 +58,8 @@ typedef struct ucc_knomial_pattern {
     ucc_rank_t     block_size;
     ptrdiff_t      block_offset;
     int            is64;
+    uint8_t        is_mixed;
+    ucc_kn_radix_t radices[UCC_KN_MAX_RADIX_PHASES];
```
Sergei-Lebedev: instead of keeping radices here make it a point to external array. Currently it increases take size a lot

src/components/tl/ucp/tl_ucp.c
```diff
@@ -172,6 +172,13 @@ ucc_config_field_t ucc_tl_ucp_lib_config_table[] = {
      ucc_offsetof(ucc_tl_ucp_lib_config_t, allgather_kn_radix),
      UCC_CONFIG_TYPE_UINT_RANGED},
 
+    {"ALLGATHER_KN_MIXED_RADICES", "",
+     "Optional exact mixed-radix schedule for knomial allgather, for example "
```
Sergei-Lebedev: i would maybe change semantic here, instead of exact decompostion consider it as allowed factors in decomposition so it generalises to any team size in a run.

jeffnvidia: don't we want the option for the user to provide the exact order ? (between 8x6 and 6x8 for example) I'm thinking that maybe the user would for example want to start with the first radix being the number of ranks in one node.

In the second PR, I provide an auto tuner that selects the radices and the order anyways but if the user chooses to select, I was thinking we ggive him "full control"

svcnbu-swx-hpcx:
🤖 **CI Triage Agent** — `Linter` · commit `d17161bf`

**TL;DR:** The `Linter` (clang-tidy) job failed because `allgather_knomial.c:250` contains a dead store — `radix = p->radix;` is assigned but never read in the `ALLGATHER` branch — which clang-tidy's `clang-analyzer-deadcode.DeadStores` check flags as an error (exit 125). Remove the dead assignment.

**Summary:** clang-tidy failed with 1 warning-treated-as-error: "Value stored to 'radix' is never read" in `allgather_knomial.c`.

**Root cause:** In `ucc_tl_ucp_allgather_knomial_start`, the ALLGATHER branch reassigns `radix = p->radix;` (line 250) after `ucc_kn_ag_pattern_init`, but nothing in that branch subsequently reads `radix`. `radix` is only consumed in the ALLGATHERV/else branches, which use the value set at initialization (line 234) and don't depend on line 250. This makes line 250 a dead store, which clang-analyzer-deadcode.DeadStores treats as an error under `-warnings-as-errors`, so the `run-clang-tidy` step counts an `error:` and exits 125.

**Implicated commit:** d17161bf — "TL/UCP: address mixed-radix review" by jeffnvidia (the exact commit under test); the mixed-radix pattern work in ec516448 is the origin of the pattern-init reshuffle.

**File:** src/components/tl/ucp/allgather/allgather_knomial.c:250

**Suggested fix:** Delete the dead assignment `radix = p->radix;` on line 250. If the intent was for later code in the ALLGATHER branch to use the pattern-selected radix, then actually use it (e.g. pass it where needed); otherwise removing the line clears the clang-tidy error. If retaining it for symmetry/readability is desired, annotate with `// NOLINT(clang-analyzer-deadcode.DeadStores)`, but simple removal is cleanest.

**Related:** PR #1327 (this change); prior mixed-radix PRs touching this file (#1176, #1072). No existing issue found for this specific warning.

</details>

svcnbu-swx-hpcx:
🤖 **CI Triage Agent** — `Lint (ROCm)` · commit `d17161bf`

**TL;DR:** The `Lint (ROCm)` job failed because clang-tidy (run with `-warnings-as-errors`) flagged a dead store at `allgather_knomial.c:250` — `radix = p->radix;` is assigned but never read afterward. Remove that unused assignment (or use the variable) to fix the lint error.

**Summary:** clang-tidy `clang-analyzer-deadcode.DeadStores` error caused the Lint job to exit with code 125 (the build itself succeeded).

**Root cause:** In `ucc_tl_ucp_allgather_knomial_start`, the `UCC_COLL_TYPE_ALLGATHER` branch assigns `radix = p->radix;` at line 250, but `radix` is not read again in that branch (only `offset`/`rbuf` are used, then the branch ends). clang-analyzer reports "Value stored to 'radix' is never read," and because the CI step counts any `error:` line and exits 125, the lint gate fails. This dead store was introduced by the new mixed-radix allgather logic, where `ucc_kn_ag_pattern_init` computes the effective radix into `p->radix`.

**Implicated commit:** `ec516448` (jeffnvidia, "TL/UCP: add exact mixed-radix allgather"), on the current head `d17161bf` (jeffnvidia, "TL/UCP: address mixed-radix review").

**File:** `src/components/tl/ucp/allgather/allgather_knomial.c:250`

**Suggested fix:** Remove the now-unused `radix = p->radix;` line in the ALLGATHER branch. If the intent was to make the effective (possibly clamped) radix visible for later use, then it must actually be used — but since the subsequent code in that branch doesn't reference `radix`, the clean fix is to delete line 250. If `radix` is unused after `ucc_kn_ag_pattern_init` in all branches, also confirm the local isn't left otherwise unused.

**Related:** PR #1327 (this PR, `jmahou/ucc-mixed-radix-clean`). No prior matching issue found.

</details>

svcnbu-swx-hpcx:
🤖 **CI Triage Agent** — `Linter-NVIDIA` · commit `d17161bf`

**TL;DR:** The `Linter-NVIDIA` clang-tidy job failed because `allgather_knomial.c:250` stores `p->radix` into the local `radix` variable that is never read again, tripping `clang-analyzer-deadcode.DeadStores` (treated as error). Remove the dead `radix = p->radix;` assignment.

**Summary:** clang-tidy-17 static analysis failed with 1 warning treated as error: dead store to `radix` in `allgather_knomial.c`, causing the job to exit 125.

**Root cause:** In `ucc_tl_ucp_allgather_knomial_start()`, the `UCC_COLL_TYPE_ALLGATHER` branch calls `ucc_kn_ag_pattern_init(...)` and then executes `radix = p->radix;` (line 250). The local `radix` is not used anywhere after this assignment in that branch (the code uses `p` members, `offset`, and `rbuf`), so the store is dead. clang-analyzer flags it as `clang-analyzer-deadcode.DeadStores`, and the linter workflow treats warnings as errors (`1 warning treated as error`), producing exit code 125. This was introduced by the mixed-radix refactor on this branch, where the pattern-init API changed and left the reassignment vestigial.

**Implicated commit:** d17161bf — jeffnvidia, "TL/UCP: address mixed-radix review" (HEAD of PR #1327; the dead-store pattern originates in the mixed-radix series starting ec516448).

**File:** src/components/tl/ucp/allgather/allgather_knomial.c:250

**Suggested fix:** Delete the dead assignment on line 250 (`radix = p->radix;`). Since `radix` is already initialized at line 234 from `task->allgather_kn.p.radix` and is only read in the ALLGATHERV/AGX branches (lines 271, 296, 300), removing line 250 clears the warning without changing behavior. If the intent was to refresh `radix` after `ucc_kn_ag_pattern_init` mutates `p->radix`, verify whether any later ALLGATHER-branch code actually needs it — currently none does — otherwise the removal is correct.

**Related:** PR #1327 (jmahou/ucc-mixed-radix-clean); prior mixed-radix commit ec516448.

</details>

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Addressed. Fixed and scheduled radices now share the same pattern initialization and progress path; there is no separate mixed-radix execution path.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Addressed. The schedule is parsed by the configuration parser and stored in the team configuration. Collective initialization only performs a range lookup and validates the schedule product.

src/components/tl/ucp/tl_ucp.c
jeffnvidia: Addressed. ALLGATHER_KN_RADIX is now the only option and accepts fixed values, ordered schedules such as 2x3x4, and auto.

src/components/tl/ucp/allgather/allgather_knomial_schedule.c
jeffnvidia: Addressed. Parsing, bounded formatting, cloning, and release are implemented through the common K-nomial radix configuration type, with parser regression coverage.

src/coll_patterns/recursive_knomial.h
jeffnvidia: Addressed. The pattern references the team-owned schedule instead of embedding a radix array in every task.

src/coll_patterns/sra_knomial.h
```diff
@@ -261,15 +261,37 @@ static inline void ucc_kn_g_pattern_next_iter(ucc_knomial_pattern_t *p)
...
+    p->rank         = rank;
+    p->n_extra      = 0;
     p->count        = count;
-    p->block_size   = p->radix_pow * radix;
-    p->block_offset = ucc_knomial_pattern_loop_rank(p, rank) / p->block_size *
-                      p->block_size;
+    p->block_size   = p->radix;
+    p->block_offset = ucc_knomial_pattern_loop_rank(p, rank) /
+                      p->block_size * p->block_size;
+    p->radices      = radices;
```
Sergei-Lebedev: this can be simplified, most of the field here relevant for nradices=1 and nradices>1

src/coll_patterns/recursive_knomial.h
```diff
@@ -54,6 +55,7 @@ typedef struct ucc_knomial_pattern {
     ucc_rank_t     block_size;
     ptrdiff_t      block_offset;
     int            is64;
+    const ucc_kn_radix_t *radices;
```
Sergei-Lebedev: alignment

src/utils/ucc_datastruct.h
```diff
@@ -7,13 +7,22 @@
...
 
 #define UCC_LIST_HEAD UCS_LIST_HEAD
 typedef uint32_t ucc_rank_t;
 #define UCC_RANK_INVALID UINT32_MAX
 #define UCC_RANK_MAX UCC_RANK_INVALID - 1
 
+typedef uint16_t ucc_kn_radix_t;
+#define UCC_KN_MAX_RADIX_PHASES (sizeof(ucc_rank_t) * CHAR_BIT - 1)
+
+typedef struct ucc_kn_radix_schedule {
```
Sergei-Lebedev: this strucutre belongs to knomial pattern. Also schedule is not a good name, it mixes with ucc_schedule_t . Maybe ucc_kn_radix_seq?

src/coll_patterns/recursive_knomial.h
Sergei-Lebedev: why not to use ucc_kn_radix_schedule?

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -324,6 +329,13 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
         task->subset.map    = sbgp->map;
     }
     task->allgather_kn.p.radix = radix;
+    if (schedule && schedule->n_radices > 1) {
+        task->allgather_kn.p.radices = schedule->radices;
+        task->allgather_kn.p.n_iters = schedule->n_radices;
+    } else {
+        task->allgather_kn.p.radices = NULL;
+        task->allgather_kn.p.n_iters = 0;
```
Sergei-Lebedev: why need to set n_iters here? It should be part of init or reset

src/components/tl/ucp/tl_ucp.c
Sergei-Lebedev: Agree, but at the same time if we consider this config as exact decomposition then it is less useful. If context creates multiple teams of different size then most of them will skip this value. Maybe extend with extra "=", for instance
RADIX="2x3x4" tells UCC to decompose using provided values
RADIX="=2x3x4" use exact decomposition and works for team size 24 only

src/coll_patterns/sra_knomial.h
jeffnvidia: Done. Fixed and exact sequences now both use `ucc_knomial_pattern_init()`. The exact case only overrides its specific invariants, and the common AllGather fields are initialized once.

src/coll_patterns/recursive_knomial.h
jeffnvidia: Fixed.

src/utils/ucc_datastruct.h
jeffnvidia: Done. The radix type and sequence now live in `coll_patterns/knomial.h`, and I renamed the type to `ucc_kn_radix_seq_t` to avoid confusion with `ucc_schedule_t`.

src/coll_patterns/recursive_knomial.h
jeffnvidia: Done. The pattern now keeps a pointer to the complete `ucc_kn_radix_seq_t`, rather than only the raw radix array.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Done. Collective initialization now only stores the sequence pointer. `n_iters` is set by `ucc_kn_ag_pattern_init()` whenever the pattern is initialized or restarted.

src/components/tl/ucp/tl_ucp.c
jeffnvidia: I kept the `x`-separated value as an exact ordered sequence. We already have `auto` for team-size-independent selection and the explicit sequence for users who need exact ordering. To handle contexts containing teams of different sizes, a sequence-product mismatch now falls back to `auto` instead of failing. This addresses the multi-team concern without introducing a third allowed-factors mode or additional `=` syntax.

src/coll_patterns/sra_knomial.h
```diff
@@ -261,15 +261,30 @@ static inline void ucc_kn_g_pattern_next_iter(ucc_knomial_pattern_t *p)
 }
 
 static inline void
-ucc_kn_ag_pattern_init(ucc_rank_t size, ucc_rank_t rank, ucc_kn_radix_t radix,
-                       size_t count, ucc_knomial_pattern_t *p)
+ucc_kn_ag_pattern_init(ucc_rank_t size, ucc_rank_t rank,
+                       ucc_kn_radix_t radix,
+                       const ucc_kn_radix_seq_t *radix_seq, size_t count,
```
Sergei-Lebedev: let's pass only radix_seq here as argument. No reason to keep both while single value radix can be represented as radix_seq of len 1

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -357,19 +364,68 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
     return UCC_OK;
 }
 
+ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
+    ucc_base_coll_args_t *coll_args, ucc_base_team_t *team,
+    ucc_coll_task_t **task_h, ucc_kn_radix_t radix)
+{
+    return ucc_tl_ucp_allgather_knomial_init_common(
+        coll_args, team, task_h, radix, NULL);
+}
```
Sergei-Lebedev: why do we need init_common if it just same as init_r?

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -229,8 +231,10 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_start(ucc_coll_task_t *coll_task)
     ucc_tl_ucp_context_t       *ctx   = UCC_TL_UCP_TEAM_CTX(team);
     ucc_coll_type_t             ct    = args->coll_type;
     ucc_rank_t                  size  = task->subset.map.ep_num;
-    ucc_kn_radix_t              radix = task->allgather_kn.p.radix;
     ucc_knomial_pattern_t      *p     = &task->allgather_kn.p;
+    const ucc_kn_radix_seq_t   *radix_seq = p->radix_seq;
+    ucc_kn_radix_t              radix = radix_seq ? radix_seq->radices[0] :
+                                                   p->radix;
```
Sergei-Lebedev: alignment

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -324,6 +329,8 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
         task->subset.map    = sbgp->map;
     }
     task->allgather_kn.p.radix = radix;
+    task->allgather_kn.p.radix_seq =
```
Sergei-Lebedev: keep only radix_seq, single radix is radix_seq of size 1

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -357,19 +364,68 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
...
+    } else {
+        radix = radix_seq->radices[0];
+    }
+    radix = ucc_min(radix, tsize);
+    if (coll_args->args.coll_type != UCC_COLL_TYPE_ALLGATHER) {
+        return ucc_tl_ucp_allgather_knomial_init_r(
+            coll_args, team, task_h, radix);
+    }
+    if (!ucc_tl_ucp_allgather_knomial_seq_matches(radix_seq, tsize)) {
+        tl_debug(UCC_TL_TEAM_LIB(tl_team),
```
Sergei-Lebedev, line 399: rework code duplication

src/components/tl/ucp/allgather/allgather_knomial.c
```diff
@@ -357,19 +364,68 @@ ucc_status_t ucc_tl_ucp_allgather_knomial_init_r(
...
+                                                tl_team->opt_radix;
+    } else {
+        radix = radix_seq->radices[0];
+    }
+    radix = ucc_min(radix, tsize);
+    if (coll_args->args.coll_type != UCC_COLL_TYPE_ALLGATHER) {
+        return ucc_tl_ucp_allgather_knomial_init_r(
+            coll_args, team, task_h, radix);
+    }
+    if (!ucc_tl_ucp_allgather_knomial_seq_matches(radix_seq, tsize)) {
```
Sergei-Lebedev: is it enough? will it work if we have a sequence like this [0]?

src/coll_patterns/recursive_knomial.h
```diff
@@ -54,6 +55,7 @@ typedef struct ucc_knomial_pattern {
     ucc_rank_t     block_size;
     ptrdiff_t      block_offset;
     int            is64;
+    const ucc_kn_radix_seq_t *radix_seq;
```
Sergei-Lebedev: alignment

test/gtest/utils/test_parser.cc
```diff
@@ -69,3 +71,105 @@ UCC_TEST_F(test_parse_mrange, check_range_multiple) {
...
+    }
+};
+
+UCC_TEST_F(test_parse_kn_radix, valid_ranged_schedules)
+{
+    ucc_mrange_kn_radix_t p;
+
+    ASSERT_TRUE(ucc_config_sscanf_kn_radix(
+        "0-4K:host:8x6,4K-inf:cuda:4,auto", &p, NULL));
+    expect_schedule(
```
Sergei-Lebedev: need to rename after changing kn_schedule > kn_seq. Here and other places

test/gtest/Makefile.am
```diff
@@ -144,6 +144,10 @@ gtest_LDFLAGS  += $(UCX_LDFLAGS)
 gtest_LDADD    += $(UCX_LIBS) $(UCX_LIBADD)
 endif
 
+if TL_UCP_ENABLED
```
Sergei-Lebedev: why test knomial_schedule depends on TL UCP?

src/coll_patterns/sra_knomial.h
jeffnvidia: Done. `ucc_kn_ag_pattern_init()` now takes only `radix_seq` and derives the initial radix from it. A fixed radix is represented as a sequence of length one.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Removed `init_common`. The sequence initializer is now the private implementation. The existing scalar `init_r` entry point is retained as a thin compatibility wrapper and delegates through a one-element sequence.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Fixed.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Done. The task now stores only the persistent `radix_seq`. Fixed radix is stored as a one-element sequence. `p->radix` is initialized by the pattern and is used only as the active radix for the current phase.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Done. Initialization now constructs the auto sequence once and only overrides it for a valid configured fixed or exact sequence. An invalid exact sequence simply leaves the auto value in place. Sequence validation is centralized.

src/components/tl/ucp/allgather/allgather_knomial.c
jeffnvidia: Handled. The parser rejects radix zero, and validation independently rejects a constructed one-element zero sequence, zero inside a mixed sequence, missing storage, overflow, and product mismatch. These cases are covered by the new validation test.

src/coll_patterns/recursive_knomial.h
jeffnvidia: Fixed.

test/gtest/utils/test_parser.cc
jeffnvidia: Done. The helper, test names, test file, and configuration descriptions now consistently use “sequence” rather than “schedule”.

test/gtest/Makefile.am
jeffnvidia: It no longer depends on TL/UCP. The test was renamed to `test_knomial_seq.cc`, uses only the generic K-nomial pattern code, and is now included unconditionally in the gtest sources.

nsarka (approved):

jeffnvidia:
@Sergei-Lebedev, I addressed the latest review comments and replied to each thread. The current head adds only a final formatting cleanup, and all CI checks pass. Could you please take another look? I left the threads unresolved for your confirmation.

Sergei-Lebedev (approved):

PR #1335, opened 2026-07-30
CORE: add team (communicator) caching

bwestheimer:
## What

Adds transparent team caching to UCC core. When enabled
(`UCC_TEAM_CACHE_ENABLE=y`, default off), teams built for identical-membership
communicator groups are cached and reused instead of being rebuilt on every
create. No public API changes; non-cache create and destroy paths are untouched.

The main pieces:
- **Dormant reuse.** A destroyed team is retained and re-adopted when an
  identical-membership team is created again.
- **Derived teams.** A second identical-membership create while the first team
  is still live (e.g. `MPI_Comm_dup`) builds a lightweight derived team that
  borrows the parent's immutable membership and topology artifacts while getting
  its own team id and tag domain — skipping address exchange and topology build.
- **Cross-rank agreement** (`UCC_TEAM_CACHE_AGREEMENT`, default on). A small
  member-scoped allreduce reconciles each member's reuse-versus-fresh decision
  so reuse stays safe even for overlapping subcommunicators.
- **Eviction policies:** `none`, `fifo`, and `lfu` (with `lru` accepted as an
  alias), clamped against the team-id pool so cached teams cannot exhaust it.
- **Experimental cid-drift reseat** (`UCC_TEAM_CACHE_RESEAT`, default off).

New configuration knobs are documented in `docs/user_guide.md`.

## Why

Building a `ucc_team` is expensive: it involves an out-of-band address exchange,
unique team-id allocation, and per-component initialization. Workloads that
repeatedly create and destroy communicators, or that duplicate identical-membership
communicators, pay that full cost every time. Caching lets those patterns reuse
existing teams, which cuts team-management overhead substantially while keeping
collective results identical.

## How

**Identity.** A team is keyed by an FNV-1a hash over the materialized rank
membership plus the caller's own endpoint. The external id is compared but not
hashed, and lookup verifies the full membership array to guard against hash
collisions.

**Lifecycle.** A cached team moves through NONE, DORMANT, RESERVED, and LIVE
states, all under the cache lock. A destroyed team drops to DORMANT at refcount 0
and is re-adopted on a later matching create.

```mermaid
stateDiagram-v2
    direction TB

    [*] --> NONE : allocate

    NONE --> LIVE : build complete (addr-exchange + topo + CL/TL)

    LIVE --> DORMANT : destroy — refcount → 0, team retained

    DORMANT --> LIVE : direct re-adopt (size == 1 or agreement off)
    DORMANT --> RESERVED : cache hit, size > 1 — cross-rank vote posted

    RESERVED --> LIVE : vote unanimous → promote
    RESERVED --> NONE : vote lost → rebuild in place

    DORMANT --> [*] : eviction

    note right of LIVE
        DERIVED_FROM_LIVE: when a LIVE team of identical membership
        exists, a new derived team borrows the parent's ctx_map,
        ctx_ranks, and topo — skipping address exchange and topo
        build entirely. The derived team follows its own NONE→LIVE
        path; the parent stays LIVE unchanged.
    end note
```

**Artifacts refactor.** `ctx_map`, `ctx_ranks`, and `topo` are moved from direct
fields on `ucc_team_t` into a refcounted `ucc_team_artifacts_t` holder so derived
teams can safely borrow a parent's immutable data. New accessor macros
`UCC_TEAM_CTX_MAP()`, `UCC_TEAM_CTX_RANKS()`, and `UCC_TEAM_TOPO()` replace all
direct field references across CL/TL components.

**Agreement.** On a cacheable multi-rank create, every member posts a nonblocking
vote (an allreduce over the context service team) so all ranks commit the same
reuse-or-fresh action. If any rank lacks a service team, every rank disables the
cache for that create.

**Derived teams** borrow the parent's refcounted artifacts holder. The topology is
fully materialized before sharing (via `ucc_topo_prepare_shared`) so concurrent
access across the parent and its derived teams is race-free.

**Eviction** is coupled to the team-id pool. `max_size` is clamped so dormant
teams reserve id-pool headroom, and eviction runs proactively so a free id is
always available before the next pre-active allocation.

Also fixes a pre-existing bug in the team-id pool bit-manipulation: the word
index used `id/64` instead of `(id-1)/64`, causing misindexing for every
multiple-of-64 team id.

## Testing

- gtest unit coverage in `test/gtest/core/test_team_cache.cc`: identity and
  hashing, eviction policies, chained buckets, lifecycle and refcount, the
  RESERVED state, the agreement vote, and concurrency.
- MPI multi-rank correctness in `test/mpi/test_team_cache.cc`: dup coexistence,
  dormant reuse, derived rebuild, freed-callback safety, divergent-eviction
  agreement, and non-blocking create. Set `UCC_TEAM_CACHE_CORRECTNESS_TESTS=y`
  to activate. Passing at 8 ranks.
- A cache on-versus-off equivalence script (`test/mpi/run_cache_equivalence.sh`)
  that also asserts the cache was actually exercised.

## Notes for reviewers

- Disabled by default — no behavior change unless `UCC_TEAM_CACHE_ENABLE=y`.
- The RESEAT feature (`UCC_TEAM_CACHE_RESEAT`) is experimental and off by default;
  it is safe to ignore for an initial review.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -195,6 +195,100 @@ which of those TLs with the highest score is selected.
 
 Tuning UCC heuristics is also possible with the UCC configuration file (`ucc.conf`). This file provides a unified way of tailoring the behavior of UCC components - CLs, TLs, and ECs. It can contain any UCC variables of the format `VAR = VALUE`, e.g. `UCC_TL_NCCL_TUNE=allreduce:cuda:inf#alltoall:0` to force NCCL allreduce for "cuda" buffers and disable NCCL for alltoall. See [`contrib/ucc.conf`](../contrib/ucc.conf) for an example and the [FAQ](https://github.com/openucx/ucc/wiki/FAQ#13-ucc-configuration-file-and-priority) for further details.
 
+## Team (communicator) caching
+
+Building a UCC team (collective communicator structure) is an expensive operation involving out-of-band communication, unique ID allocation, and per-component initialization. UCC can transparently cache and reuse teams built for identical-membership communicator groups, avoiding repeated reconstruction for the same process sets.
+
+### When caching helps
+
+Caching accelerates **create → use → free → recreate-identical** workloads via two
+complementary mechanisms:
+
+1. **Dormant reuse** — a freed team is retained and re-adopted when an
+   identical-membership communicator is recreated:
+   - An MPI program repeatedly creates and destroys communicators with the same rank membership (e.g., iterative solvers splitting a rank subset each iteration).
+   - Repeated `MPI_Comm_split` or `MPI_Comm_create` with stable rank sets or colors (common in ScaLAPACK-style process-grid splitting and domain decomposition).
+
+2. **Derived teams (simultaneous coexistence)** — when a second identical-membership
+   communicator is created while the first is *still live* (e.g., `MPI_Comm_dup` of a
+   live communicator), UCC builds a lightweight **derived team** that shares the
+   parent's immutable membership and topology while getting its **own** team id (its
+   own tag/sequence domain, so the two live teams never alias). This skips the
+   address exchange and topology build — the bulk of team-creation cost.
+
+Both mechanisms are transparent and work for externally-provided team ids (e.g. an
+MPI communicator's context id): dormant reuse re-adopts only when membership **and**
+id match, while derived-create keys coexistence on membership alone (a dup legitimately
+carries a different id than its live parent).
+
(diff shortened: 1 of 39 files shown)


bwestheimer:
Superseded by the decomposed series #1346-#1353, which carries the same change split into eight reviewable PRs.

Each PR in the series is self-contained and stacked in this order:

- #1346 CORE: Fix team-id pool off-by-one at 64 boundary
- #1347 CORE: Add ctx-scoped service allreduce
- #1348 CORE: Add team cache with exact reuse and agreement
- #1349 CORE: Extract shared team state into artifacts holder
- #1350 CORE: Add derived-team caching and experimental reseat
- #1351 CORE: Add team cache LFU eviction policy
- #1352 CORE: Guard team cache against persistent handles
- #1353 TL/NCCL: Guard ncclCommDestroy and document update_id

Closing this one so review effort lands on the series rather than here.

PR #1343, opened 2026-08-25
UCP: Tune Alltoall pairwise posting depth

jeffnvidia:
## Summary

The existing TL/UCP pairwise Alltoall automatic selector has a binary cliff:
after the 66,000-byte compatibility boundary, teams through 32 ranks post the
full team while larger teams post one peer at a time. This PR replaces that
choice with coarse per-peer bands. Explicit
`ALLTOALL_PAIRWISE_NUM_POSTS` values remain authoritative.

The evidence is organized by team size because the selector changes a
different payload region at each scale:

- **N16:** focused repeated validation over 16/32/64/128 MiB per peer gives
  **+18.19%/+53.08%/+38.48%** geometric-mean bandwidth on Systems A/B/C.
- **N32:** the four common power-of-two band representatives are nonnegative
  on all three systems. A newer independent System C control confirms
  **+45.47%/+50.10%/+52.99%/+59.25%** at 16/32/64/128 MiB per peer.
- **N64:** the four common power-of-two band representatives are nonnegative
  on all three systems; the selector returns to legacy P1 above 8 MiB.
- **N <= 8:** full posting is retained, so these ranks are present in the
  policy diagram but omitted from performance panels.
- **N128:** evidence remains partial rather than a three-system comparison;
  the focused 512-KiB controls improve by **+28.35%** and **+49.75%** under
  the two measured protocol configurations.

Alltoall policy and rank-specific power-of-two validation

Systems A/B/C denote independent validation environments. Gray is legacy
posting normalized to 1.00x; colored bars are the PR selector. Every payload
label also shows `legacy P -> PR P`.

Panel B is focused repeated N16 validation. Panels C/D reconstruct the four
power-of-two band representatives shared by the available N32/N64 sweeps;
they are measured points from different dates, not interpolation or a uniform
new campaign. Their logarithmic axes keep the large lower-band improvements
from hiding smaller changes. The System C lower-band N32/N64 values are
historical single-sweep scouts and are presented as such; the System C N32
16-MiB point is the newer independent 48/48 control.

## Automatic policy

```text
if aggregate_bytes <= 66,000:      P = N
else if N <= 8:                    P = N
else if peer_bytes <= 64 KiB:      P = min(N, 32)
else if peer_bytes <= 1 MiB:       P = min(N, 16)
else if peer_bytes <= 4 MiB:       P = N for N <= 16, otherwise 8
else if peer_bytes <= 8 MiB:       P = min(N, 4)
else if N <= 32:                   P = 4
else:                              P = 1
```

Per-peer bytes use the same calculation as the progress path:
`(count / team_size) * datatype_size`. The aggregate guard preserves existing
behavior in an extremely small-message region that was not independently
retuned.

## Power-of-two evidence

At N16 the policy changes only the large-message region, so the common focused
grid is 16--128 MiB per peer:

| Per peer | System A | System B | System C |
| ---: | ---: | ---: | ---: |
| 16 MiB | +40.98% | +47.90% | +36.66% |
| 32 MiB | +15.07% | +51.31% | +37.64% |
| 64 MiB | +8.49% | +55.29% | +39.23% |
| 128 MiB | +10.86% | +58.00% | +40.41% |
| **Geometric mean** | **+18.19%** | **+53.08%** | **+38.48%** |

System A uses P15 as the nearest sampled proxy for legacy P16. Systems B/C
compare P4 directly with P16.

<details>
<summary><strong>Reconstructed N32 power-of-two band representatives</strong></summary>

| Per peer | Legacy -> PR | System A | System B | System C |
| ---: | ---: | ---: | ---: | ---: |
| 64 KiB | P32 -> P32 | +0.00% | +0.00% | +0.00% |
| 1 MiB | P32 -> P16 | +12.83% | +20.49% | +282.18%* |
| 4 MiB | P32 -> P8 | +39.43% | +13.52% | +1222.10%* |
| 16 MiB | P32 -> P4 | +50.56% | +24.89% | +45.47% |

`*` Historical single-sweep scout. The very large ratio reflects an unusually
poor legacy P32 measurement in that same sweep and is not used as a portable
speedup claim.

</details>

<details>
<summary><strong>Reconstructed N64 power-of-two band representatives</strong></summary>

| Per peer | Legacy -> PR | System A | System B | System C |
| ---: | ---: | ---: | ---: | ---: |
| 64 KiB | P1 -> P32 | +915.51% | +500.80% | +808.04%* |
| 1 MiB | P1 -> P16 | +87.35% | +81.86% | +233.66%* |
| 4 MiB | P1 -> P8 | +17.96% | +11.17% | +104.43%* |
| 16 MiB | P1 -> P1 | +0.00% | +0.00% | +0.00% |

`*` Historical single-sweep scout. These points establish direction, not an
exact portable gain.

</details>

The power-of-two reconstruction deliberately uses the four exact sizes shared
by all three systems. It does not fill missing intermediate points or average
across incompatible campaigns.

## Supplementary focused and held-out validation

The position-balanced/held-out System A grid remains useful as an independent
whole-selector check because its sizes fall inside the policy bands rather
than directly on their power-of-two boundaries:

| Ranks | Per-peer range | Selected P sequence | Aggregate change | Worst measured change |
| ---: | ---: | --- | ---: | ---: |
| 32 | 48 KiB--12 MiB | 32/16/16/16/16/8/8/4/4 | **+22.27% GM** | **0.00%** |
| 64 | 48 KiB--12 MiB | 32/16/16/16/16/8/8/4/1 | **+104.13% GM** | **0.00%** |
| 128 | 512 KiB | 16 | **+28.35% / +49.75%** | **+28.35%** |
| 32 (System C control) | 16--128 MiB | 4/4/4/4 | **+51.87% GM** | **+45.47%** |

The N32 held-out series uses position-balanced replacements for four
launch-order-sensitive cells; N64 uses all nine held-out cells. The System C
N32 large-message control passed both correctness canaries and all 48
streaming measurements.

## Large-message portability boundary

P1 is excellent on Systems A and B at N16 but regresses against legacy on
System C. P4 improves over legacy on all three measured systems, although it
is not the sampled oracle on System A. This is why the PR uses a conservative
static compromise and treats a dynamic tuner as the longer-term solution.

## Scope and relationship to topology ordering

This PR changes only automatic posting depth for TL/UCP pairwise Alltoall. It
does not change Alltoallv, peer order, or explicit posting-depth settings.

The node-interleaved order in #1344 remains opt-in. This PR's validation uses
the ordinary ring and does not claim that the two PRs form one automatic
policy.

## Checks

- Focused tests cover the aggregate guard, every inclusive per-peer threshold,
  the 8/9, 16/17, and 32/33-rank transitions, valid range, and monotonicity.
- A standalone selector boundary harness passes with warning-as-error flags.
- The selector header is included in TL/UCP's Automake source manifest.
- Exact head `2ec433a4` passes all 16 GitHub checks.


wfaderhold21 (approved):
lgtm

PR #1346, opened 2026-08-31
CORE: Fix team-id pool off-by-one at 64 boundary

bwestheimer:
## What
Fix an off-by-one in the team-id pool release path. `ucc_team_release_id` computed `map_pos = id / 64`, which misplaces the bit for any id that is a multiple of 64: id 64 writes to word 1 bit 63 (the slot for id 128) instead of word 0 bit 63. The fix is `(id - 1) / 64`. The assert guarding `id >= 1` is also moved before the arithmetic so it fires before any signed subtraction on an invalid id.

A new `test_team_id_pool.cc` gtest covers boundary ids 1, 63, 64, 65, 127, 128, 129, 191, 192.

## Why
The allocation path produces ids via `i * 64 + pos` where `pos ∈ [1, 64]`, so multiples of 64 are valid ids. The release path missed them, leaking the id and double-freeing an unrelated slot. The bug is latent in typical workloads where fewer than 64 teams coexist, but fires reliably once more teams are live simultaneously.

PR #1347, opened 2026-08-31
CORE: Add ctx-scoped service allreduce

bwestheimer:
## What
Add `embedded` flag to `ucc_service_coll_req_t` and a new `ucc_service_allreduce_ctx()` function.

When `embedded` is set, `ucc_service_coll_finalize` skips `ucc_free(req)` so callers can embed the request inside a larger struct rather than heap-allocating it. `ucc_service_allreduce_ctx()` runs a BAND allreduce over a caller-supplied ctx-rank subset via the context-level service team, without requiring a built team `ctx_map`.

## Why
Part of the communicator caching series (see #1346 for context). The cache agreement vote runs during `ucc_team_create_post` before the new team's `ctx_map` is available. The existing `ucc_service_allreduce` requires a built map; this variant takes a direct ctx-rank list. The `embedded` flag lets the vote request live inside `ucc_team_t` with no separate allocation.

Depends on #1346.

### DIFF
--- a/src/core/ucc_service_coll.c
+++ b/src/core/ucc_service_coll.c
@@ -21,6 +21,14 @@ uint64_t ucc_service_coll_map_cb(uint64_t ep, void *cb_ctx)
     return ucc_ep_map_eval(team->ctx_map, team_rank);
 }
 
+/* Maps a subset index straight to a context rank, bypassing team->ctx_map */
+static uint64_t ucc_service_coll_map_cb_direct(uint64_t ep, void *cb_ctx)
+{
+    ucc_service_coll_req_t *req = cb_ctx;
+
+    return ucc_ep_map_eval(req->subset.map, (ucc_rank_t)ep);
+}
+
 static inline ucc_status_t
 ucc_service_coll_req_init(ucc_team_t *team, ucc_subset_t *subset,
                           ucc_tl_team_t          **service_team,
@@ -38,6 +46,7 @@ ucc_service_coll_req_init(ucc_team_t *team, ucc_subset_t *subset,
     }
     req->team   = team;
     req->subset = *subset;
+    req->embedded = 0;
 
     if (ctx->service_team) {
         *service_team         = ctx->service_team;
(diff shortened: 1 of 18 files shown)


src/core/ucc_service_coll.c
```diff
@@ -38,6 +46,7 @@ ucc_service_coll_req_init(ucc_team_t *team, ucc_subset_t *subset,
     }
     req->team   = team;
     req->subset = *subset;
+    req->embedded = 0;
```
wfaderhold21, line 49: alignment

src/core/ucc_service_coll.c
```diff
@@ -80,6 +89,41 @@ ucc_status_t ucc_service_allreduce(ucc_team_t *team, void *sbuf, void *rbuf,
...
+                                       size_t count, ucc_reduction_op_t op,
+                                       ucc_subset_t subset,
+                                       ucc_service_coll_req_t *req)
+{
+    ucc_context_t  *ctx = team->contexts[0];
+    ucc_tl_team_t  *steam;
+    ucc_tl_iface_t *tl_iface;
+    ucc_status_t    status;
+
+    ucc_assert(ctx->service_team != NULL);
```
wfaderhold21, line 103: question: should we assert or return UCC_ERR_NOT_SUPPORTED here?

src/core/ucc_service_coll.c
```diff
@@ -146,9 +190,12 @@ ucc_status_t ucc_service_coll_test(ucc_service_coll_req_t *req)
 ucc_status_t ucc_service_coll_finalize(ucc_service_coll_req_t *req)
 {
     ucc_status_t status;
+    uint8_t      embedded = req->embedded;
```
wfaderhold21, line 193: initialized vars first. then declarations

wfaderhold21 (approved):
lgtm

src/core/ucc_service_coll.c
bwestheimer: Done.

src/core/ucc_service_coll.c
bwestheimer: Done.

src/core/ucc_service_coll.c
bwestheimer: Thanks, good catch. This is a real NULL dereference rather than just a style question.

`ctx->service_team` is NULL when the context has no OOB or `UCC_INTERNAL_OOB` disallows it, and its creation is non-fatal, so NULL is a supported state. `ucc_service_coll_req_init` already handles it with a fallback. Release builds define NDEBUG, so the assert compiled out and `UCC_TL_TEAM_IFACE` dereferenced NULL. I reproduced it as a SIGSEGV with `UCC_INTERNAL_OOB=0 UCC_TEAM_CACHE_ENABLE=y`.

Changed to return `UCC_ERR_NOT_SUPPORTED`. There is no fallback to `team->service_team` because this collective addresses context ranks directly.

#1348 also drops the team cache when agreement is requested but no context service team exists, so this error path should not be reachable in practice.

janjust (approved):

PR #1348, opened 2026-08-31
CORE: Add team cache with exact reuse and agreement vote

bwestheimer:
## What
Add a per-context communicator team cache. Teams that are destroyed are retained as DORMANT rather than freed, and re-adopted on the next `ucc_team_create_post` with identical membership, skipping the ADDR_EXCHANGE and topology build that a fresh create would require.

A cross-rank BAND allreduce agreement vote runs on every cacheable create so all members reach the same reuse-or-build decision before any rank skips the exchange. This makes reuse safe when subcommunicators overlap. The cache is off by default (`UCC_TEAM_CACHE_ENABLE=n`).

Seven commits, intended to be read in order:
1. Team cache identity fingerprint (FNV-1a hash + exact-compare, no container yet)
2. Cache container, refcount API, and FIFO eviction
3. Wire cache into context config and team lifecycle (first end-to-end working cache; `AGREEMENT=n` is a supported config, user guide documents the overlapping-scope restriction)
4. Tests: dormant reuse lifecycle and cache knobs
5. Cross-rank agreement vote
6. Tests: agreement vote and overlap safety (includes a test that deadlocks with the vote disabled and passes with it enabled)
7. CI: cache-on vs cache-off equivalence pass

## Why
Team creation is the dominant latency in workloads that repeatedly create subcommunicators, such as split-collectives and windowed pipelines. Caching avoids the per-create allgather and topology build without changing the UCC API.

## How
New files: `src/core/ucc_team_cache.{c,h}`. Modifications to `ucc_team.{c,h}`, `ucc_context.{c,h}`, and `ucc_service_coll.{c,h}` for lifecycle hooks and config. No TL or CL changes in this PR.

Depends on #1346, #1347.

PR #1349, opened 2026-08-31
CORE: Extract shared team state into artifacts holder

bwestheimer:
## What
Extract `ctx_map`, `ctx_ranks`, and `topo` from flat fields in `ucc_team_t` into a new `ucc_team_artifacts_t` refcounted holder. All teams initially use an embedded inline holder, so there is no behavioral change in this PR. Every call site across TL and CL components is updated to use `UCC_TEAM_CTX_MAP()`, `UCC_TEAM_CTX_RANKS()`, and `UCC_TEAM_TOPO()` accessor macros.

Three commits:
1. Add holder struct, inline init/put functions, and accessor macros (core files only)
2. Accessor macro sweep across TL/CL components (mechanical, one-liner per site)
3. `ucc_topo_prepare_shared()`: eagerly materialize all lazily built topo fields before a topo is shared across derived teams

## Why
Prerequisite for derived-team caching (#1349), which shares one holder between a live parent team and a derived team that borrows its already-built membership map and topology.

Depends on #1348.


svcnbu-swx-hpcx:
🤖 **CI Triage Agent** — `Linter` · commit `f918f6fe`

**TL;DR:** The Linter (clang-tidy-17) job failed because the static analyzer flagged use-after-free bugs in the new team-cache code — `ucc_debug` calls that reference pointers (`cache_ptr` / `team_ptr`) which alias memory already released by `ucc_free`. Move the debug logging *before* the `ucc_free`.

**Summary:** GitHub Actions "Linter" job exited 125 — clang-tidy-17 reported 3 `-warnings-as-errors` in `core/ucc_team.c` and 1 in `core/ucc_team_cache.c`.

**Root cause:** In `ucc_team_cache_destroy()` the sequence `ucc_free(cache)` (line 95) then `ucc_debug("ucc_team_cache destroyed: %p", cache_ptr)` (line 96) triggers `clang-analyzer-unix.Malloc` — `cache_ptr` aliases the freed `cache`. The same pattern in `ucc_team_cache_progress_pending()` (`ucc_team.c`): after `ucc_team_destroy_single(team)` frees `team` (via `ucc_free(team)` at line 955), the `ucc_debug(...)` at line 1114 references `team_ptr`, which the analyzer treats as use-after-free; it also emits an `UndefinedBinaryOperatorResult` (garbage value) at line 1101 as a knock-on. All are treated as errors, so the job aborts.

**Implicated commit:** 830d5ad5 / 8e8001c0 "CORE: Add cross-rank team cache vote" / "Add team cache container and FIFO eviction" — Bryce Westheimer (same author/branch as this PR).

**File:** `src/core/ucc_team_cache.c:95-96` and `src/core/ucc_team.c:1114` (with knock-on at `:1101`).

**Suggested fix:** Emit the debug log *before* freeing, so no freed pointer is referenced. In `ucc_team_cache_destroy()` move the `ucc_debug("ucc_team_cache destroyed: %p", cache_ptr)` above `ucc_free(cache)` (and you can then drop the `cache_ptr` alias). In `ucc_team.c`, move the success `ucc_debug(...)` (lines 1114-1118) so it runs before the team's memory is released inside `ucc_team_destroy_single` — e.g. log immediately after checking `status == UCC_OK` but before the team can be freed, using the already-captured `team_ptr`/`team_id` only for values that don't require dereferencing freed memory. Alternatively, if these are deemed false positives, add a targeted `// NOLINT(clang-analyzer-unix.Malloc)` — but reordering the log is the correct fix.

**Related:** PR #1349 (bwestheimer/pub-team-cache-artifacts) — none other found.

</details>

